PR #56, opened 2019-08-21
Waralee

woutervdub:


joostdecock:
Thanks Wouter :)  I don't have the bandwidth to merge this in prior to the v2 release, so I'm going to kick this one down the road until after. It gives us a good reason to bump to 2.1 :)

PR #407, opened 2020-05-10
Addition of mirror plugin

### LINKED ISSUE #405: Feature request: A plugin to mirror points around a line
This feature would flip points/paths across a line defined by two points similar to how the current flip plugin works but on an arbitrary axis.
Current flip plugin: 
https://github.com/freesewing/freesewing/blob/develop/packages/plugin-flip/src/index.js

Unlike the current flip plugin, it would need to take in two points to build the flip axis, and possibly be configured to flip selected points and paths. 

My use case is a section of a part in a pattern that needs mirrored/reflected across 45°. The entire part does not need to be reflected.
 
suggested macro API:
```javascript
macro("reflect", {
  start: Point,
  end: Point,
  points: [Point],
  paths: [Path]
}
```

Similarly to the current flip plugin, it would loop through all supplied paths ( or optionally, points )
and reflect them across the line defined by start/end. the new points/paths are added to the part.

Should we add an option to copy/move the existing points? My use case is to create a copy.

As references to points/paths are referenced as properties on the part, we need a way to provide new names to the points/paths if they are being added to the part's existing points/paths.

destos:
closes #405 

The initial draft of mirror plugin, I have some tests I wrote to check results of `mirrorGen` and `lineValues` but am unsure how tests of plugin packages are run with lerna. Any pointers there would help.

### DIFF
--- a/packages/plugin-mirror/.babelrc
+++ b/packages/plugin-mirror/.babelrc
@@ -1,0 +1,10 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "useBuiltIns": "entry"
+      }
+    ]
+  ]
+}
--- a/packages/plugin-mirror/CHANGELOG.md
+++ b/packages/plugin-mirror/CHANGELOG.md
@@ -1,0 +1,3 @@
+# Change log for: @freesewing/plugin-mirror
+
+## Unreleased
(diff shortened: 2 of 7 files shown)


joostdecock:
Great stuff. Plugin tests seem to have been overlooked a bit when me migrated to Lerna :see_no_evil: 

I'll need to fix that. I created https://github.com/freesewing/freesewing/issues/408 for that. In the meanwhile, don't worry about it.

destos:
@joostdecock, the only reason I had exported`lineValues` and `mirrorGen` was for testing purposes.

PR #493, opened 2020-08-14
Master

woutervdub:
Fixed error on Benjamin ('band')


joostdecock:
Since there's a good bit of *noise* in this PR, [I've *cherry-picked* the commit that fixes the issue](https://github.com/freesewing/freesewing/commit/7848df795e03689403120a182a761bf3e1fa9dac) and committed that, rather than merge this PR.

While this means it's now committed by me, the code change is still attributed to you (git is smart that way).

 As such, I'm closing this as the relevant changes have been merged in a cleaner way. Thanks for the fix! I'll bundle this up and put out 2.8.1 so I can deploy this fix to freesewing.org.

PR #605, opened 2020-10-13
Make Albert author consistent with other patterns by @woutervdub

joostdecock:


joostdecock:
Sorry, added an irrelevant commit to this PR, will create a new one

PR #694, opened 2020-11-23
Remove extra space in svg group translate

### LINKED ISSUE #692: SVG opens with all piece overlapping in Adobe Illustrator and Photoshop
SVGs open with all pattern pieces in the upper right of the image on top of each other in Adobe Illustrator and Photoshop. 

PS files open correctly

sannek:
Closes #692 

There was an extra space in the second `translate(x, y)` for SVG group tags, causing Illustrator to ignore that translation and render all pattern pieces in the same spot.


joostdecock:
Flawless!

Issue #754, opened 2020-12-26
Create issue template for documentation/content updates


ephphatha:
As suggested in discord, we should have a template for content updates that don't fit the existing bug report/showcase templates.

(mainly creating this to have an issue to attach to the pull request)

PR #755, opened 2020-12-26
Add template for documentation/content updates

ephphatha:
Might be too dry? Happy to have the wording/format changed. I based this off the showcase template as that appeared most suitable for content changes.

Addresses https://github.com/freesewing/freesewing/issues/754

I haven't updated the changelog as I'm not sure what the convention is with keeping that up to date (do it before or after the merge?) or what you'd classify this sort of change as...

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation-update.md
+++ b/.github/ISSUE_TEMPLATE/documentation-update.md
@@ -1,0 +1,16 @@
+---
+name: Documentation/content update
+about: Suggest updates to FreeSewing documentation or site content
+title: Documentation/content update
+labels: "\U0001F44D good first issue, \U0001F4D6 documentation"
+assignees: ''
+
+---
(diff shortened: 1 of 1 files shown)


joostdecock:
I would rather drop the licensing part. We don't have a Contributor License Agreement for any code contributrions, so it seems weird to have it for documentation things.

I guess this is a side-effect of you basing this of the showcase template. But that's somewhat different since we are adding pictures to the repo that are (almost always) owned by our users, not our contributors. So there we want to ensure that we have the ok from the user (since that's the right thing to do) and adding it to the issue template creates a paper trail.

But in this case, I would just drop that. Apart from that, it's good to go for me.

joostdecock:
Thanks for this @ephphatha I've sent you an invite to join the freesewing organisation as a contributor.

Issue #781, opened 2021-01-11
NPM install fails on windows as the template files are not copied


ephphatha:
**What seems to be the problem?**

create-freesewing-pattern/lib/create-library.js calls globby() to get a list of files to copy when creating the folder/file structure for a new pattern. This uses a path to the create-freesewing-pattern/template/ directory built with path.join(). On windows this returns a path like C:\<node_cache>\create-freesewing-pattern\template\ (using standard windows path separators of backslashes).

This is not handled by globby so the call results in an empty list and none of the template files are copied.
> Note that glob patterns can only contain forward-slashes, not backward-slashes, so if you want to construct a glob pattern from path components, you need to use path.posix.join() instead of path.join(). ([From the globby package documentation](https://www.npmjs.com/package/globby).)

This causes npm install and npm link to fail.

**Are you a FreeSewing patron?**

 - [ ] Yes, I am :hugs:
 - [x] No, I am not :thinking:

joostdecock:
Hey, thanks for this quality bug report. Windows support is a bit sketchy, because... well I don't have a Windows machine.

Still, this seems like low-hanging fruit and we should try to correct it.

PR #784, opened 2021-01-12
Normalise source directory when passing it to globby

ephphatha:
Fixes https://github.com/freesewing/freesewing/issues/781

globby() expects forward slash characters as path separators, so calling normalize-path transforms backslashes to forward slashes on windows while preserving the drive letter (handles the case where node packages are installed on a different drive to where the command is being run).

The unixify package could be an alternative but in testing it was overly destructive on the path (stripping the drive letter) which caused issues with multi-drive environments.

When run on unix this effectively does nothing, so I didn't bother trying to test for the platform to see if calling normalize was necessary, it works fine in both environments.

### DIFF
--- a/packages/create-freesewing-pattern/lib/create-library.js
+++ b/packages/create-freesewing-pattern/lib/create-library.js
@@ -4,6 +4,7 @@ const handlebars = require('handlebars')
 const execa = require('execa')
 const fs = require('fs')
 const globby = require('globby')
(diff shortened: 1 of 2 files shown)


joostdecock (approved):
LGTM 👍

joostdecock:
Great work. I'll roll this into v2.11.3

PR #820, opened 2021-01-27
Add normalize-path dependency to support #781

ephphatha:
The previous pull request manually edited the package.json file, I didn't realise it was autogenerated at the time.


joostdecock:
Thanks for this @ephphatha !

PR #967, opened 2021-04-22
Add curved waistband option to Charlie

sannek:
Like discussed on Discord, this adds a curved waistband option to Charlie to provide better fit for folks like myself with a large waist/hip ratio.

The curve amount is configurable from 0 (default - just drafts the straight waistband) to 35%. This upper limit is mostly given by the accuracy of the outer curve, until 35-40% the curve that attaches to the waistband is no more than 0.25% off (a millimetre or two on human scale), but above that the error increases rapidly. At 35% curve the waistband is already significantly more curved than any sewing pattern I own, so I think this will be enough for by far most people who want a curved waistband.

Updated instructions are still coming - I hope to get to those tomorrow or this weekend. (The main thing that needs updating is that you need to cut 2 of the curved waistband and can't just fold it over like the straight one).

I'd love to know if the paperless measurements seem reasonable because I had a hard time figuring out what was the best way to do that.

Lastly, I think I messed up trying to get my fork up to date and I don't know how to get rid of this Merge PR commit 🙃 
Ok, lastly -lastly, Idk if you merge or squash, so let me know if you'd like me to clean up the commit history a bit better.


sannek:
Ok I have no idea why these two tests are failing 😭 both say they can't find a module, so perhaps it has something to do with my weird merge commit?

joostdecock:
It's a missing dev dependency. Don't worry, it's not you, it's me. I'm working on some tests right now, will fix this after that.

joostdecock:
Great work @sannek I should be able to bundle this up in v2.15.1 this weekend :)

PR #1077, opened 2021-06-04
Adds eslint to dependencies and resolves code to match rules

jejacks0n:
Add eslint to the dev dependencies for create-freesewing-pattern templates. When firing up a new pattern I was getting some eslint errors/warnings -- because the dependencies weren't clear, and so I've added the relevant plugins and standard configurations.

After installing the eslint dependencies, I resolved any violations that I found.

### DIFF
--- a/packages/create-freesewing-pattern/template/freesewing/example/src/App.js
+++ b/packages/create-freesewing-pattern/template/freesewing/example/src/App.js
@@ -10,12 +10,11 @@ import Pattern from './pattern/src/index.js'
  * Without it, we can't import the pattern as a local file
  * since create-react-app does not allow imports outside ./src
  * If it's imported as a dependency, webpack will cache the
- * build and there will be no hot-relaoding of changes
+ * build and there will be no hot-reloading of changes
  */
 
-
 const App = (props) => {
-  // You can use this to add transations
+  // You can use this to add translations
   /*
   let translations = {
     JSON: 'JSON',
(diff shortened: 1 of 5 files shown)


packages/create-freesewing-pattern/template/freesewing/example/src/App.js
```diff
@@ -26,17 +25,19 @@ const App = (props) => {
...
+  } else {
+    recreate = false
+  }
 
   return (
     <Workbench
       freesewing={freesewing}
       Pattern={Pattern}
-      userLanguage="{{language}}"
+      userLanguage="en"
```
jejacks0n, line 38: Just noticing this might be inadvertently changed. Needs to be resolved before merging.

joostdecock: Put this back to be the placeholder

joostdecock:
Hey, this is great stuff. Just one thought on the matter: We use prettier across (certain parts of) the monorepo to keep consistent code formatting.

Obviously, a linter is more/better, we just need to keep an eye on things so that we don't end up in this perpetual back and forth between eslint and prettier changing things like trailing commas and whatnot.

Not saying this will happen. Just writing this down so when it does, I'll remember why ;)

jejacks0n:
> Hey, this is great stuff. Just one thought on the matter: We use prettier across (certain parts of) the monorepo to keep consistent code formatting.
> 
> Obviously, a linter is more/better, we just need to keep an eye on things so that we don't end up in this perpetual back and forth between eslint and prettier changing things like trailing commas and whatnot.
> 
> Not saying this will happen. Just writing this down so when it does, I'll remember why ;)

Haha, yeah, I saw that but didn’t do a deep dive to find/resolve differences. I think we should change the linting rules to match prettier if that’s the case. I’m not opinionated on the rules, just that the base code doesn’t violate any of them.

Want me to follow this up with a more compatible eslint config?

joostdecock:
I've re-added eslint to the patterns too. I took it out a while back because it wasn't properly setup (and I seemed to have a hard time getting around to it). So it would be a good occasion to do it across the board (or at least for all patterns/plugins).

Perhaps something that is not obvious: Pretty much all `package.json` files are auto-generated. If you want to update/add a dependency you can change `config/dependencies.yaml` and run `yarn reconfigure`.

This is perhaps a bit hackish, but since so many of our packages are typically used together in 1 project (think freesewing.org) I really wanted to ensure that dependency versions are consistent across the board so we don't end up with multiple copies in the bundle.

Long story short: Yes, I would love your help on this eslint stuff.
Also, we use the rules in `packages/prettier-config` (published as `@freesewing/prettier-config`

PR #1086, opened 2021-06-06
remove departments in pattern descriptions

starfetch:
remove references to "womenswear"/"menswear"/"men"/"women" in descriptions of patterns, and replace with "with/without breasts"


joostdecock:
Good catch!

Issue #1111, opened 2021-06-13
Dev docs: dx(), dy() and dist() do not use methods in example


starfetch:
The examples given in the developer documentation for the methods `dx()`, `dy()` and `dist()` do not show how to use them in the given examples, `vd`/`hd`/`ld` macros are used instead.

Low priority, since `dx()` and `dy()` are covered extensively in the tutorial.

Relevant pages:
https://freesewing.dev/reference/api/point/dist/
https://freesewing.dev/reference/api/point/dx/
https://freesewing.dev/reference/api/point/dy/

PR #1126, opened 2021-06-15
dev docs: change dx(), dy(), dist() examples -- code

starfetch:
As mentioned in issue #1111, I have rewritten the examples given in the `dist()`, `dx()` and `dy()` dev docs to show the usage of these methods. It doesn't look as nice as before, but is hopefully useful as an example.

This is the corresponding PR for `freesewing`. (PR for `markdown` [here](https://github.com/freesewing/markdown/pull/246))

Feedback and improvements welcome :)

### DIFF
--- a/packages/examples/src/point_dist.js
+++ b/packages/examples/src/point_dist.js
@@ -1,13 +1,18 @@
 export default (part) => {
-  let { Point, points, macro } = part.shorthand()
+    let { Point, points, Path, paths} = part.shorthand()
 
   points.from = new Point(10, 10)
-  points.to = new Point(90, 40)
+  points.to = new Point(80, 70)
 
-  macro('ld', {
-    from: points.from,
-    to: points.to,
-  })
-
-  return part
+points.text = points.from
+	.shiftFractionTowards(points.to, 0.6)
+	.attr("data-text", points.from.dist(points.to)+"mm")
+	.attr("data-text-class", "text-sm fill-note center");
+    
+    paths.line = new Path()
+	.move(points.from)
+	.line(points.to)
+	.attr("class", "dashed");
+    
+    return part
 }
--- a/packages/examples/src/point_dist.js
+++ b/packages/examples/src/point_dist.js
@@ -1,18 +1,19 @@
 export default (part) => {
-    let { Point, points, Path, paths} = part.shorthand()
 
-  points.from = new Point(10, 10)
-  points.to = new Point(80, 70)
+let { Point, points, Path, paths } = part.shorthand()
+
+points.from = new Point(10, 10)
+points.to = new Point(80, 70)
 
 points.text = points.from
-	.shiftFractionTowards(points.to, 0.6)
-	.attr("data-text", points.from.dist(points.to)+"mm")
-	.attr("data-text-class", "text-sm fill-note center");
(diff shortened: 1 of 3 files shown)


joostdecock:
Hmm, I see your point. It would make more sense to have examples that use the methods, rather than illustrate what they do via a macro. But I'm not convinced this PR is the way to do it.

See, the code example is part of the docs. For example the `dist` example lives here: https://github.com/freesewing/markdown/blob/develop/dev/reference/api/point/dist/en.md

The code in the `example` pattern is what is shown above the code example. The graphical representation of the result, so to speak. 

Perhaps what we need to do is come up with a better example to illustrate what these methods do.
My concern is that these are really easy methods, kinda trivial really. But this code example now is more complicated and seems to suggest that this is how to do it (when the macro is way simpler).

Am I making any sense?

starfetch:
Hi @joostdecock, I can totally see your point regarding the complexity of the example. (Sorry for the late reply, by the way. Was a bit swamped at work.)

Maybe we can leave `dx()` and `dy()` as they are for the time being (since, as you say, they really are trivial), but change the `dist()` example; at least to me it does not seem more complex than e.g. [the one for `angle()`](https://freesewing.dev/reference/api/point/angle/). (Although maybe this is more of a point *against* the `angle()` example instead of one *for* the `dist()` one.)

It just bugs me that an example claiming to be about a certain method does not even use the method. All the macro does is show what `x` and `y` distances are. If you didn't go through the bib tutorial (which is hopefully not many people, since it is excellent), you wouldn't know that you can easily create points with `dx()` and `dy()`.
But maybe I'm trying to catch too many contingencies/solve problems that are not really there. If you think that having a more involved example for such trivial methods would confuse people rather than help, I'm happy to concede.

I'm not sure I understand what you mean with the code example being part of the docs. Please correct me if I'm wrong, but I thought I changed the relevant files in `markdown`, too ([PR #246 in `markdown`](https://github.com/freesewing/markdown/pull/246)). Am I missing a level of complexity here?

joostdecock:
> But maybe I'm trying to catch too many contingencies/solve problems that are not really there. If you think that having a more involved example for such trivial methods would confuse people rather than help, I'm happy to concede.

The issue you raise is valid. I just would like to find an example that is a more isolated/simple use of the methods.
You are :100: correct that an example illustrating a method should use that method.

How about something like (this is from the top of my head):

```js
points.origin = new Point(0,0)
for (const i = 1; i++; i<6) { 
  const key = `p${i}`
  points[key] = new Point(i*20,0)
  snippets[key] = new Snippet(
    (points.origin.dx(points.key) > 50) ? 'notch' : 'bnotch',
    points[key]
  )
}
```

> I'm not sure I understand what you mean with the code example being part of the docs.

Sorry, I saw this PR in isolation and did not realize it came with a flanking PR in markdown

starfetch:
I wonder if that's really easier?  Having a `for` loop creating new points seems like a whole other level of complexity coding-wise, and the conditional `snippet` even more so. (It's a really neat example though.)

joostdecock:
Well if you have to wonder whether it's easier, it's clearly not, is it :joy: 

You've convinced me. Merging. Thank you for taking the time to let me get smarter.

starfetch:
Sure! And maybe we (or someone else) come up with an easier example one day :)

PR #1207, opened 2021-07-12
fix(create-freesewing-pattern): remove old departments

starfetch:
removed old departments from initial create-freesewing-pattern prompt and replaced with `tops` and `bottoms`

Not sure if anything else depends on those values, so kindly requesting review from @joostdecock once he is back from his code vacation :)

(Also: Not sure what the changelog policy is, so I didn't touch it for now.)


joostdecock (approved):

joostdecock:
Ah crap. I forgot to pull this in before making the 2.17.1 release :see_no_evil:

PR #1273, opened 2021-08-28
Method re-evaluation. Holmes Construction en.md

bobgeorgethe3rd:
I was re-reading the method as I am planning to make up the drawings for this and realised Oh yea you can't understitch after closing the lining opening gg me,
Changes
-Steps 11 and 12 of Assembly and Lining Method 2 have been swapped
-Steps 11 and 13 now both state that they are optional or Alternates for each other.
-Tutorial changed to video as Bernadette Banner generally does not refer to these videos as tutorials and it is more of a visual aid rather than a tutorial.


darigovresearch:
Reordering looks fine, not sure what is causing the checks to not pass. @joostdecock any advice on why the automatic checks are not working?

joostdecock:
I need to look into it, but I'm working for $work now. However, I doubt it's related to this PR and more likely I broke something when adding the strapi instance to the monorepo.

joostdecock:
Thanks @bobgeorgethe3rd :+1: 

@darigovresearch I've fixed the tests, there were some small issues left and right with dependency handling in the auto-generated tests. Everything should be ok again now.

Issue #1267, opened 2021-08-24
Fix Wahid frontLining Seam Allowance


bobgeorgethe3rd:
**What seems to be the problem?**

The seam allowance path of the frontLining part of Wahid is incomplete. It appears that the paths.seam is simply not .close() so when offset, the seam allowance path does not close.

**Are you a FreeSewing patron?**

 - No, I am not :thinking:

**Additional context**

https://cdn.discordapp.com/attachments/757632180980547686/879677504091201556/unknown.png

bobgeorgethe3rd:
Pull request opened #1276

PR #1276, opened 2021-09-01
 Fix for Issue #1267 Wahid Seam Allowance

bobgeorgethe3rd:
#1267 
Was going to push last night then thouht I'd better test it so put it away. Glad I did, my original fix of putting .close on paths.seam would have altered the seam line. It has been changed to being on paths.sa instead which is probably where it should of been from the get go.

Changes
- .close() on paths.sa not on paths.seam


joostdecock:
A closed path that you `.offset()` will generate a closed path. So the correct approach here was to close `paths.seam`.

The problem in your original commit was that you closed the path before it was complete.
Should have moved the `.close()` call a couple of lines lower and it would have been :ok_hand:

Issue #462, opened 2020-07-19
Make simon & simone support button placement for left-handed people 


joostdecock:
Update from Discord by @joostdecock :

> I personally find it difficult to motivate myself to work on this because of the history.
> Women's shirts are inverted because (wealthy) women would not dress themselves but had a chamber maid for that.
> So the buttons are inverted to make it easier for a person standing in front of you to button them (so in reality, both women's and men's shirts are optimized for right-handed people, the majority)
> So while I understand that women are used to this, I find it personally difficult to justify spending my weekends trying to implement the continuation of a traditional that throws all women under the bus because that makes it a bit easier for the servants of the wealthy.
> 
> But, as I'm writing this, I realize there's a loophole to bypass my own moral block: we can make it about right-handed vs left-handed people. And that makes it a matter of going the extra mile to support a minority.
> And I for one find that far easier to get excited about 🤗

Original description:

> Both simon and simone have their buttons and buttonholes as men are used to them.
>
> The idea here is to have the option to reverse that. And perhaps have that option enabled by default for simone.

bobgeorgethe3rd:
[Example of Classic Style Button and French Style (Seamless) Buttonholes ](https://cdn.discordapp.com/attachments/550472363922423808/885687322820112404/unknown.png)
https://github.com/bobgeorgethe3rd/freesewing/tree/develop/packages/simon

So I thought I would take a stab at this and the above is what I have come up with ^. Essentially it flips and renames the left and right Fronts so the options for buttons and the options for buttonholes do not get twisted and tangled.

The option is buttonholePlacement: {dflt:'Left over Right', list:['Left over Right','Right over Left']} and is in the closure sub-catergory. <- Names can be changed just best mine could come up with.

The reason this is not a pull request right now is the paperless is not done and I just want to check I am on the right track before going down the rabbit hole of paperless. Also just in case I have missed anything.

bobgeorgethe3rd:
I have yet to figure out how this will factor onto Simone and/or how to make a Simone to use a development Simon when drafting. But we'll cross that bridge when Simon is ready.

bobgeorgethe3rd:
Added the collar stand but I feel like buttons and butonholes are Right over Left to begin with? I haven't changed it as the default but I thought I'd get a second pair of eyes on it because to me they look the Right over Left on the current Pattern.

joostdecock:
Please submit a PR, you can label it a draft, but it makes it easy to compare changes and test them out in a feature branch.

[...]

PR #1311, opened 2021-09-10
Simon ButtonHole and Button Placement Support

bobgeorgethe3rd:
Additions:
- Added options.buttonholePlacement

Changes:
- Front Left, Front Right, Buttonhole Placket, Button Placket now flip when options.buttoholePlacement is rightOverLeft
- Collarstands snippets are changed to button to buttonhole, buttonhole to button when options.buttoholePlacement is rightOverLeft

What still needs doing:
- Paperless changing on options.buttonholePlacement
- Seeing how this will affect Simone
-Double checking if the current Collarstands buttons and buttonholes are the right way.


bobgeorgethe3rd:
Draft for #462

PR #1355, opened 2021-09-30
Update Ursula

nataliasayang:
- Added translations
- Updated option defaults


joostdecock:
Can't believe you figured out where to add these translations :joy:

PR #1369, opened 2021-10-08
Bug fix for crash report.

woutervdub:
This fixes the following crash report:
TypeError: proxy set handler returned false for property '"seatWaistDiff"'


joostdecock:
Tx Wouter, we'll put this in 2.19 as that's being wrapped up

Issue #444, opened 2020-07-08
Feature request: Add a mandarin collar to Simon(e)


joostdecock:
Jesse writes:

> it would be great if there were a "no collar" or "mandarin" collar option

stoffsuchti:
Great idea with the mandarin collar

DocSpencer77:
Just yesterday was thinking to suggest multiple collar options.  Thought drafting a collar is about a match for my skill level. Maybe.

joostdecock:
This is very stale, so I'm closing it. There's the *packs* idea on [our roadmap](https://github.com/freesewing/freesewing/discussions/1278) so perhaps it can be part of that effort.

PR #1380, opened 2021-10-13
Simon Buttonhole RightOverLeft Support

bobgeorgethe3rd:
Ok so this is the updated version off #1311 adding support for #462 as it had gotten a bit messy due to me mucking up my repository and also needed updating to remove the mm options. Added support for #444

Additions
- Added option "buttonholePlacement" which allows to switch between "leftOverRight" and "rightOverLeft" buttonhole placement.
- Added option "collarStyle" which allows switch between "straight" and "noCollar"
- Added missing Button Paperless for frontright-seamless

Changes
- "buttonholeplacket" now supports options.buttonholePlacement
- "buttonplacket" now supports options.buttonholePlacement
- "collarstand" now supports options.buttonholePlacement
- "frontright" now supports options.buttonholePlacement
- "frontleft-classic-cuton" now supports options.buttonholePlacement
- "frontleft-classic-seperate" now supports options.buttonholePlacement
- "frontleft-seamless" now supports options.buttonholePlacement
- "frontright-classic-cuton" now supports options.buttonholePlacement
- "frontright-classic-seperate" now supports options.buttonholePlacement
- "frontright-seamless" now supports options.buttonholePlacement
- added store to shorthand of "buttonholeplacket" and "buttonplacket" (not sure if they were meant to be missing but needed it to run)
- Paperless support added

Still needed
- Checking Simone.

This works by using a combination of macro('flip') and manually changing negative and positive signs where the flip does not work/apply to "flip" the frontright and frontleft parts to appear as eachother whilst retaining their button and buttonhole configurations respectfully. It also changes the title of both parts when flipped.

I said screw it and did the paperless.

I also added collarStyle support as I saw that it was an old issue. I added it as a list rather than a bool to leave it open for future if we wanted to add any more styles.


bobgeorgethe3rd:
Also. I am still unsure on how this will affect/ what changes will need to be applied to Simone for this to take effect. I know that Simone imports Simon so I am wondering if there is a way to use the local version of Simon to test Simone.

joostdecock:
So to test the impact of this on Simone, you could just use the local build of simon in your monorepo fork:

- enter `packages/simon`
- run `yarn build` (or `npm run build`)
- copy build into simone example: `cp dist* ../simone/example/node_modules/@freesewing/simon/dist/`
- Enter the `example` folder in `simone` and run `yarn start`
- Simone will now use the local simon

bobgeorgethe3rd:
Conflict with the buttonholePlacket and buttonPlacket. Simply was that I had added the store, Store, shorthand when working on it as it had came up as an error and I just autopilot fixed it not thinking it was an issue.

bobgeorgethe3rd:
Removed the delete snippets.armholePitch which was removing the back notch from the back. #1509

Issue #1342, opened 2021-09-21
Directory of missing Markdown and Illustrative Work.


bobgeorgethe3rd:
## Please read before helping.
The following is a list of all the missing and needed to be worked on Markdown and Illustrative Work. Ultimately the goal would be to get these all done before the v3 update so it is complete in conjuction with the update but it is not a deal breaker if it does not. Please note that cover arts are done by one specific artist and are on the list mostly for show, if you have an idea on how to represent the pattern though we would like to hear it.

To help, First check if there is already an issue assigned to the task you wish to help with.
- If there is go there and see if you can help.
- If it does not create an issue for the task and comment with the task and #issue number below so the list can be updated. Note: Tag me if the list hasn't been updated after 24 hours. You may also need to assign the Designers/Coders to the issue to get insight on the pattern.
- If it is strikethroughed/ticked it has been completed and does not need further assistance.
- If it is a simple fix just create a pull request with the solution rather than an issue. i.e. adding missing markdown.

If you think anything needs to be added/removed from the list please leave a comment below and it will be changed.

*Note*: A custom cutting layout system is currently being worked on and will more than likely supersede the need for cutting diagrams so these are not a priority.

Definitoin of Tags
**[illustrative]**
Stuff like the cover arts that are more interpretive.
**[diagram]**
Diagrams and illustrations for instructions/cutting diagrams/technical drawings etc.
**[markdown]**
Text amd formatting on the website such as instructions and other documentation.
**[open graph]**
Optimsing for the information taken when posting a link in places such as Discord
**[website]**
Incorrect displays that cannot be fixed by markdown.
   
#### All or a lot of
 - [x] Add yardage to all need pages and correct plural and singulars.
 - [x] Now that the split has finalised many now have markdown pattern options updates required **[markdown]** 
 
#### Aaron
 - [ ] Aaron cutting diagrams **[diagram]**

#### Albert
- [x] Albert Fabric Options proof read **[markdown]**
- [x] Albert Instructions **[markdown]** #1082
- [ ] Albert instruction diagrams **[diagram]**
- [ ] Albert cutting diagrams **[diagram]**

#### Bee
- [x] Bee Cover Art **[illustrative]**
- [ ] Bee Cross Back Tie Variant Line Drawing? **[diagram]**
- [ ] Bee cutting diagrams **[diagram]**

#### Bella
- [ ] Bella Cover Art **[illustrative]** #1626
[...]

PR #1389, opened 2021-10-15
Declartive Titles for Aaron en.md for Open Graph

bobgeorgethe3rd:
One of the goals for #1342 is to make declaritive titles for all pattern docs for open graph.
This is Aaron's but only the English. I did try to do the non-english ones but realised that Crowdin would do it similar to how it has done for Holmes plus I am not a speaker of any of them so best not.

### DIFF
--- a/markdown/org/docs/patterns/aaron/cutting/en.md
+++ b/markdown/org/docs/patterns/aaron/cutting/en.md
@@ -1,3 +1,6 @@
+---
+title: Aaron Cutting
+---
 
  - Cut **1 back** on the fold
  - Cut **1 front** on the fold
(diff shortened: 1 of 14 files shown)


bobgeorgethe3rd:
Also wanted to make sure this is what we discussed in the contributer call before doing the rest of the patterns.

joostdecock:
Thank you for submitting this, and I'm glad you did so before the contributor call so we can discuss this.

It's actually great from an *this is what we want* point of view, but it's not going to be taken into account because these titles are not pulled from the markdown.

The reason is that there is no guarantee that whoever translates the title of a markdown documentation page will also be the person translating the options themselves.

This leaves the door open for a given option having two different translations, and now users will be confused because the option listed on the pattern page is not listed in the documentation.
For that reason, we pull the title out of the options, and use that instead.

I'll leave this open for now, and we can discuss this further tonight.

Issue #1421, opened 2021-10-27
Convert Waralee from mm to snapped percentage options


joostdecock:
Unit tests now check for `mm` options, and having makes tests fail.
Which is the case in Waralee: https://github.com/freesewing/freesewing/runs/4024610907?check_suite_focus=true

Creating this to see if @woutervdub would be willing to convert this. You can use Paco as an example.

joostdecock:
Done by @woutervdub :+1:

PR #1423, opened 2021-10-27
Convert Waralee from mm to snapped percentage options (Issue #1421)

woutervdub:


darigovresearch:
@woutervdub Thanks for making a pull request & wanting to contribute to the project. Could you take a look at the automated checks and see if they can be resolved?

This would help make your pull request more likely to be merged

For convenience a link to issue #1421 which this is attempting to resolve can be found [here](https://github.com/freesewing/freesewing/issues/1421)

joostdecock:
I noticed you converted the mm options to standard pct options, no snapping.
I don't think it matters all that much for the hem, but for the waistband width, I'd say it does.

Did you chose to do it this way for a specific reason? Or is it more due to unfamiliarity with the new snap settings?

woutervdub:
That was both unfamiliarity, and just looking at how you solved it in Charlie. I’m not quite sure why it would matter in the waistband for Waralee, since that doesn’t use anything that should be sized. It’s just a slightly bigger hem. I don’t even line mine with anything. 

Wouter

joostdecock:
Ok, sounds good. Merging this as the failing test is unrelated to this (and on my todo list)

PR #1440, opened 2021-11-03
Bent Documentation Update

bobgeorgethe3rd:
Documentation Update for Bent Like Bella.
All written for default block and has some info on pattern altering.

Added
- Bent Cutting
- Bent Fabric Options
- Bent What you need
- Bent Instructions


joostdecock:
You've been busy! Thanks for these :+1:

PR #1475, opened 2021-11-07
Charlie Missing Options

bobgeorgethe3rd:
Added
- Cross Seam Angle
- Crotch Seam Angle

I took the info from the Pattern like I did from Breanna but they may be a little vague.


darigovresearch (approved):
Looks good to us, this solves the immediate problem of missing documentation and anyone can make a further pull request to improve it if required

Issue #1467, opened 2021-11-05
Waist Height Option not working on Titan and by extenstion Charlie


bobgeorgethe3rd:
As mentioned by user @dreamingfifi in discord
https://discord.com/channels/698854858052075530/757632953097388043/906288586507321407

I looked into this and yea the option is not working on either Titan or Charlie.

dreamingfifi:
Or Paco, which is how I stumbled onto it.

PR #1483, opened 2021-11-08
Fix for #1467 Titan Waist Height not Working

bobgeorgethe3rd:
Fix for Issue #1467 
On inspection found the option had been removed from front.js and back.js in commit 8252b12 
Made a comment on the commit but decided it was just best to make a pull request with the fix and have it be closed if the removal of the option was intentional.
This just undoes the removal of the options on front.js and back.js

Hope this helps.


joostdecock:
Good Fix. Bad Joost.

PR #1502, opened 2021-11-20
Updating and adding unit tests

chri5b:
I've updated the unit tests for the first handful of plugins starting in alphabetical order. I've also added some tests for banner and bartack. 
As discussed with @joostdecock on developer-help, I've added calls to Utils.Round where it made sense but have left failing tests as is. 
Please let me know if I've messed anything up or if there are conventions you'd like me to respect.


chri5b:
Will resubmit once I've fixed the failing variable error

PR #1503, opened 2021-11-20
Updating and adding unit tests

chri5b:
I've updated the unit tests for the first handful of plugins starting in alphabetical order. I've also added some tests for banner and bartack. This time with a fix to the issue with my previous commit.
As discussed with @joostdecock on developer-help, I've added calls to Utils.Round where it made sense but have left failing tests as is.
Please let me know if I've messed anything up or if there are conventions you'd like me to respect.


darigovresearch:
@chri5b thanks for making the time to do this pull request! Could you take a look at the automated checks to see if you can do anything about the one that is failing?

chri5b:
Thanks @darigovresearch. I'm struggling to work out from the details of the failed check what I need to change in order for it to pass. If you have any guidance, I'd appreciate it.

joostdecock:
We've discussed this on Discord; the plugin unit tests have been broken for a while so this is great help to out them back on the rails.

I'm happy to take this PR as-is since it's a step towards where we want to be: which is in a place where all tests run successfully.

I just need to find time to work on this, which is proving difficult these days 🙈

joostdecock:
@chri5b This was a great contribution. I've sent you an invite to the FreeSewing organisation, which is much deserved.

Here's hoping you'll stick around :slightly_smiling_face:

Issue #408, opened 2020-05-10
Plugin tests have not been adapted for monorepo setup


joostdecock:
For example, plugin-title has tests, but `package.json` is not setup to run them.

joostdecock:
Closing this, because while there are still a few plugins without (plugin-specific) unit tests, they are now all properly integrated in the monorepo.

Special thanks to @chri5b for their work on this which really motivated me to finally tackle this :pray:

PR #1521, opened 2021-11-27
Fixing/Adding more plugin unit tests

chri5b:
Fixing or adding unit tests for remaining plugins. There is a failing test on plugin-validate because the tests expect an error but the code doesn't raise one. I've not changed the code because I'm not certain about the downstream impacts. Thanks for your welcome and feedback on my previous PR.


joostdecock:
I'm going to have to go through this to pull out the actual tests, because I've made changes to the way we handle the unit tests in the monorepo ( to address #408 ).

Specifically:
 - Plugin unit tests are ESM now (import instead of require)
 - The unit tests are auto-configured for plugins, just as they are for patterns, resulting in a `shared.test.mjs` file that runs the shared tests
 - Plugin-specific tests go in `plugin.test.mjs`
 - I've also setup workflows for github actions to run the plugin tests

This is all *around* the actual tests, so I'm sure there's plenty of useful stuff in here.

There's a couple of plugins that are also somewhat in a problematic state:
 - plugin-export-dxf: Is de-facto unmaintained because we do not use it
 - plugin-validate: Is not compatible with the current version of core. It needs to be re-written. I've made it do nothing for now since at least using it won't break stuff

I will resolve the conflicts here to make sure we can get your contributions in the repo. I wont' be able to do it today, but I will take care of it.

Thank you once again for your help, you deserve all the praise :heart:

chri5b:
Hi Joost, thanks for the detailed feedback. If you're OK with me going about it at my own speed, I don't mind updating my PR to work with the new approach. I'll leave plugin-export-dxf and plugin-validate alone for now. Of course, if you get to it before me, no problem.

joostdecock:
Sure thing, take your time :)

nicholasdower:
This looks great. As it hasn't been touched in a while I thought I might take a stab at rebasing and getting the tests working. Please let me know if that will conflict with your work.

chri5b:
Hi - please feel free. Thank you.

joostdecock:
Thanks guys, I've been meaning to revisit this. Great that it's getting some TLC 👍

nicholasdower:
I had trouble rebasing and decided it might be easier to just copy the tests to a new branch. I opened a new PR (https://github.com/freesewing/freesewing/pull/1748). It contains one commit with @chri5b as the author. ptal

I'm going to close this PR in favor of that PR. Please feel free to reopen.

PR #1552, opened 2021-12-01
Huey Documentation Update

bobgeorgethe3rd:
The instructions will probably need a proof read as it is late for me when I did this.
Also the zipper instructions may need double checking.

Added

- Huey Cutting
- Huey What you need
- Huey Instructions


joostdecock:
Done is better than perfect, so I'm happy to merge this as-is. (fixed just a small thingie)

PR #1569, opened 2021-12-08
Technical Change to Bee cup.js

bobgeorgethe3rd:
As I am working on Peach again I have realised that the points.waistDartRightNew could of easily been a .rotate() rather than the nonsence that it was. I made the change as I assume it will be less of a stress to run but it certainly tidies up the code a little.
To reflect this waistDartRightNew has been changed to waistDartRightRotated.

### DIFF
--- a/packages/bee/src/cup.js
+++ b/packages/bee/src/cup.js
@@ -46,15 +46,7 @@ export default function (part) {
     points.bustDartTop,
     points.bustDartBottom.dist(points.sideHemInitial)
   )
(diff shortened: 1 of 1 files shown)


bobgeorgethe3rd:
This does not effect the way that Bee drafts. I also realise that sideHemNew is technically not in the correct posistion due to it not being rotated but it is really there for a beam and the bellaGuide option so can doesn't warrant correcting.

joostdecock:
Very happy when people maintain their own code :+1:

PR #1629, opened 2022-01-09
Address misaligned match lines on Simon front and left pieces

nicholasdower:
# Background

I recently attempted the Simon pattern and ran into a small issue with pattern matching. As I am very new to sewing and this was my first time ever sewing from a pattern, I think it is likely that I have misunderstood something or made a mistake.  Would you mind reviewing my issue and proposed change to see whether it makes sense?

# Problem Description

The issue is with the pattern matching between the two front pieces. When my shirt is fully buttoned, the pattern alignment is off by around six millimeters:
image

You can see how the pattern should match if I unbutton the shirt and shift the sides:
image

Looking at the generated pattern pieces and aligning the button and buttonhole lines, I see that the match lines are indeed misaligned:
image

As far as I can tell, this is caused by the fact that the position of the match line on the front right piece is based solely on the width of the button placket. I believe it should be based on the size of the buttonhole placket.

# Approach 

In this change I have attempted to add a separate match line to the front right piece. It is based on the buttonhole placket width and is usually in a different position that the line marking the edge of the button placket. I have also updated both front pieces to only display match lines if the placket on the other side is not separate. As far as I can tell, match lines are never displayed on separate plackets.

# Testing

I was a little unsure which combinations of plackets I needed to support. I believe there are 16 theoretically possible placket combinations: separate/attached buttonhole * classic/seamless buttonhole * separate/attached button * classic/seamless button. In practice it seems all combinations involving separate plackets can be ignored since they do not include match lines. That leaves the following combinations:

## attached-classic buttonhole, attached-classic-button

image

## attached-classic buttonhole, attached-seamless-button

image

## attached-seamless buttonhole, attached-classic-button

image

## attached-seamless buttonhole, attached-seamless-button
image

## Close lines

I also tested the behavior when the match line and placket edge are very close:
image
image

# Open Questions

I wasn't sure what to do about notches. I see that notches are normally displayed on the match lines and on the edges of the plackets, but as far as I can tell these notches do not match with any other notches. I have tentatively removed the notches I could not understand from the button placket. I did not remove the corresponding notches from the buttonhole placket.. I could use some guidance here.

~~I didn't find any way to have the match line meet the curve of the neck. Please let me know if there is a way to improve this.~~ (Found `utils.lineIntersectsCurve`. Added a separate commit for this.)

### DIFF
--- a/packages/simon/src/frontleft-classic-cuton.js
+++ b/packages/simon/src/frontleft-classic-cuton.js
@@ -52,8 +52,12 @@ export default (part) => {
       .move(points.placketBottomOuterEdgeOver)
       .line(points.placketTopOuterEdgeOver)
       .attr('class', 'dotted')
-      .attr('data-text', 'matchHere')
-      .attr('data-text-class', 'text-xs center')
+    if (!options.seperateButtonPlacket) {
+      // Match lines are only displayed on attached plackets
+      paths.placketOuterEdgeOver
+        .attr('data-text', 'matchHere')
+        .attr('data-text-class', 'text-xs center')
+    }
     paths.placketOuterEdgeUnder = new Path()
       .move(points.placketTopOuterEdgeUnder)
       .line(points.placketBottomOuterEdgeUnder)
--- a/packages/simon/src/frontleft-classic-cuton.js
+++ b/packages/simon/src/frontleft-classic-cuton.js
@@ -32,6 +32,12 @@ export default (part) => {
       .move(points.placketCfNeck)
       .line(points.placketCfHem)
       .attr('class', 'help')
+    if (!options.seperateButtonPlacket) {
+      // Match lines are only displayed on attached plackets
+      paths.frontCenter
+        .attr('data-text', 'matchHere')
+        .attr('data-text-class', 'text-xs center')
+    }
     paths.placketInnerEdgeFold = new Path()
       .move(points.placketTopInnerEdgeFold)
       .line(points.placketBottomInnerEdgeFold)
@@ -52,12 +58,6 @@ export default (part) => {
       .move(points.placketBottomOuterEdgeOver)
       .line(points.placketTopOuterEdgeOver)
       .attr('class', 'dotted')
-    if (!options.seperateButtonPlacket) {
-      // Match lines are only displayed on attached plackets
-      paths.placketOuterEdgeOver
-        .attr('data-text', 'matchHere')
-        .attr('data-text-class', 'text-xs center')
-    }
     paths.placketOuterEdgeUnder = new Path()
       .move(points.placketTopOuterEdgeUnder)
       .line(points.placketBottomOuterEdgeUnder)
--- a/packages/simon/src/frontright-classic-cuton.js
+++ b/packages/simon/src/frontright-classic-cuton.js
@@ -42,7 +42,11 @@ export default (part) => {
     macro('sprinkle', {
       snippet: 'notch',
       on: [
+        'placketTopIn',
+        'placketTopOut',
         'cfNeck',
+        'placketBottomIn',
+        'placketBottomOut',
         'cfHem',
       ],
     })
--- a/packages/simon/src/frontright-classic-cuton.js
+++ b/packages/simon/src/frontright-classic-cuton.js
@@ -20,14 +20,16 @@ export default (part) => {
   points.placketBottomEdge = points.cfHem.shift(0, width * 1.5)
 
   const buttonholePlacketWidth = store.get('buttonholePlacketWidth')
-  if (options.buttonholePlacketStyle === 'seamless') {
-    points.placketTopMatch = points.cfNeck.shift(180, buttonholePlacketWidth / 2)
-    points.placketBottomMatch = points.cfHem.shift(180, buttonholePlacketWidth / 2)
-  } else {
-    const fold = store.get('buttonholePlacketFoldWidth')
-    points.placketTopMatch = points.cfNeck.shift(180, buttonholePlacketWidth / 2 - fold)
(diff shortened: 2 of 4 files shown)


joostdecock:
I'll need to have a look at this in detail (on my phone right now) but I can already say I'm very impressed with the quality of this PR 👌

joostdecock:
I am not entirely certain why I ended up putting that match line there, but I was wondering whether we'd not make life a lot easier on ourselves if we simply moved the *match fabric here* text to center front (in other words, the line on which the buttons and buttonholes lie.

Because if you match fabric on the CF line, it will always match regardless of how wide the plackets are.

I feel that might save us some headaches in the future as there's already a lot of different pocket combinations and this would remove one thing that needs to be checked for each combo.

I'd like to hear your thoughts on this @nicholasdower as I don't want to be dismissive of your PR. I'm just wondering if we should use this opportunity to simplify things.

nicholasdower:
> I am not entirely certain why I ended up putting that match line there, but I was wondering whether we'd not make life a lot easier on ourselves if we simply moved the _match fabric here_ text to center front (in other words, the line on which the buttons and buttonholes lie.
> 
> Because if you match fabric on the CF line, it will always match regardless of how wide the plackets are.
> 
> I feel that might save us some headaches in the future as there's already a lot of different pocket combinations and this would remove one thing that needs to be checked for each combo.
> 
> I'd like to hear your thoughts on this @nicholasdower as I don't want to be dismissive of your PR. I'm just wondering if we should use this opportunity to simplify things.

This kind of blew my mind. It makes perfect sense!

Let me make the change real quick. Also, no hard feelings at all. I learned a ton poking around and I think it was worth the effort.

nicholasdower:
Thanks @joostdecock for the quick and thoughtful review. I have made the suggested change. Please take another look.

packages/simon/src/frontleft-classic-cuton.js
```diff
@@ -32,6 +32,12 @@ export default (part) => {
       .move(points.placketCfNeck)
       .line(points.placketCfHem)
       .attr('class', 'help')
+    if (!options.seperateButtonPlacket) {
```
nicholasdower, line 35: I'm not sure it makes sense to have this check. I am trying to avoid displaying a match line when there is a separate placket but it looks like this setting cannot be relied on for that. If I choose separate placket for either side of the shirt, but also set the corresponding placket style to seamless, the pattern does not actually contain a separate placket. Assuming this is a bug, maybe it makes more sense for me to address that issue first, then add this check. wdyt?

packages/simon/src/frontright-classic-cuton.js
```diff
@@ -25,24 +25,24 @@ export default (part) => {
...
       .line(points.placketBottomOut)
       .attr('class', 'dotted')
     macro('sprinkle', {
       snippet: 'notch',
       on: [
-        'placketTopIn',
-        'placketTopOut',
         'cfNeck',
-        'placketBottomIn',
-        'placketBottomOut',
```
nicholasdower, line 45: Does it make sense to remove these? Maybe these notches were being used for something other than matching the sides? If it does make sense, should I also remove the notches on the buttonhole placket?

packages/simon/src/frontleft-classic-cuton.js
nicholasdower: The idea would be to prevent the user from choosing seamless when a separate placket has been selected.

joostdecock: Yeah so the issue here is that *seamless* is a style without any seam line at the placket, and thus it is not possible to have a separate placket and seamless style. And in practice, the seamless setting takes precedent.

What's needed here is some way to designate options dependent on others and hide them in the frontend when they become irrelevant. So that when the user picks a 'seamless' placket style, the option to make it separate is no longer available.

That is not something we currently support, but it's on our roadmap now, and given that we're also building a new frontend, this is :100: something I want to add.

As to the matter at hand, we can leave the check in.

joostdecock: Forgot the link to the roadmap: #1278

packages/simon/src/frontright-classic-cuton.js
joostdecock: These notches are here to help people fold the placket during construction.

nicholasdower: Ok, I will revert these changes. I had the impression that notches were only used to match two pieces of fabric. That said, I have literally never seen a notch before this pattern.

nicholasdower: Added a commit reverting the notch removal.

joostdecock:
LGTM!

joostdecock:
@all-contributors please add @nicholasdower for code

joostdecock:
Thank you very much for this PR @nicholasdower it was an absolute pleasure to receive.

Also, this qualifies for our bug bounty project, so if you get me your postal address, I'd be happy to send you some stickers. You can either DM me on https://discord.freesewing.org/ or email me: `joost@joost.at`

Thanks again for your contribution, really great work :pray:

nicholasdower:
Thanks @joostdecock for the thoughtful review!

joostdecock:
Fix published in 2.19.9: https://www.npmjs.com/package/@freesewing/simon/v/2.19.9
Being deployed on freesewing.org as we speak: https://app.netlify.com/sites/fs-org-en/deploys/61db28e0d8f7420008225c57

nicholasdower:
> Fix published in 2.19.9: https://www.npmjs.com/package/@freesewing/simon/v/2.19.9 Being deployed on freesewing.org as we speak: https://app.netlify.com/sites/fs-org-en/deploys/61db28e0d8f7420008225c57

So rewarding to have a fix deployed same day. Thank you for taking the time.

PR #1652, opened 2022-01-15
Use 2:1 overlap to underlap ratio for Simon sleeve placket

nicholasdower:
Opening this as a draft since I have low confidence I am doing the right thing.

## Problem

I recently created a Simon shirt for myself. It had the following sleeve placket underlap and overlap:
image

Next I started working on a Simone for my wife. It had the following sleeve placket underlap and overlap:
image

The difference in the ratio between the underlap and overlap was very surprising to me and I found the underlap very hard to work with given its size. Looking at [the code](https://github.com/freesewing/freesewing/blob/develop/packages/simon/src/sleeveplacket-underlap.js#L18), I wondered whether maybe this was a bug:

```javascript
const width = store.get('sleevePlacketWidth') > 20 ? 10 : store.get('sleevePlacketWidth') / 4
```

This results in the following relationships:
image

My best guess was that the intention was to cap the max at 10 but maintain a ratio of 2 to 1 below that:
image

### DIFF
--- a/packages/simon/src/sleeveplacket-underlap.js
+++ b/packages/simon/src/sleeveplacket-underlap.js
@@ -15,7 +15,7 @@ export default (part) => {
     store,
   } = part.shorthand()
 
-  const width = store.get('sleevePlacketWidth') > 20 ? 10 : store.get('sleevePlacketWidth') / 4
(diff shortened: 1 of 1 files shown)


joostdecock (approved):

joostdecock:
@all-contributors please add @nicholasdower for bug

PR #1664, opened 2022-01-16
Sandy Documentation Update

bobgeorgethe3rd:
This was a tricky one but this is most certainly a start.
I have had to split the instructions in two, One for seamless, One for closures.
I'm hoping this is less confusing than having them combined but time will tell.
Due to circle skirts having many options on both construction and styles a lot of this is a guide line rather than a full set of methods though you could say that about any pattern.
I have proof read these but knowing me I may have missed one or two. 
This at least gives a ground-work for anyone who comes along and wants to be the ambassodor of the pattern.

### DIFF
--- a/markdown/org/docs/patterns/sandy/cutting/en.md
+++ b/markdown/org/docs/patterns/sandy/cutting/en.md
@@ -1,0 +1,22 @@
+<Note>
+
+If not using seamless follow the default list, If using seamless follow the seamless list.
+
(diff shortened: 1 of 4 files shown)


darigovresearch (approved):
Changes look good to us, thanks for taking the time to document this @bobgeorgethe3rd !

joostdecock:
Good work, thanks for adding it.

PR #1666, opened 2022-01-16
Removing Repeat Step Numbers - Holmes Instructions

bobgeorgethe3rd:
After re-formatting some of the numbers in number remained and needed removing. I was waiting to get some diagrams/pictures done before updating but that may be a minute so have opted just to sort it now.


darigovresearch (approved):
Looks good, thanks for making the updates!

PR #1703, opened 2022-01-23
fix(docs) Holmes typo in Step 3 Note

bobgeorgethe3rd:
Typo in the Ear Flap with Ties note, changed it slightly so the wording flows better and removes the typo.


nicholasdower (approved):
:shipit:

Issue #1708, opened 2022-01-23
yarn kickstart fails for NodeJs > v14


stsydow:
### Tasks
 
- [x] We should have something in the README / *Getting started* that says.
  ```
  You need NodeJS  lts/fermium (v14.8.*), consider too use _nvm_.  
  ```
  because v14.7 is also to low.
- [ ] The designer tutorial at 
https://freesewing.dev/tutorials/pattern-design
also needs an update, as it suggests that any version is fine.
- [ ] Same goes for the developer docs:
https://freesewing.dev/tutorials/getting-started-linux/installing-node

### Original Thread

running `yarn kickstart` fails with:
```
yarn run v1.22.17
$ npx lerna bootstrap && npm run buildall
npm ERR! code EDUPLICATEWORKSPACE
npm ERR! must not have multiple workspaces with the same name
npm ERR! package 'freesewing.dev' has conflicts in the following paths:
npm ERR!     /home/st/Dokumente/privat/freesewing/packages/freesewing.dev
npm ERR!     /home/st/Dokumente/privat/freesewing/packages/freesewing.lab

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/st/.npm/_logs/2022-01-23T13_22_59_006Z-debug-0.log
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I'm clone the repo just now and did no changes.
Yarn and node are preinstalled and I run Arch Linux.

Full log
```log
0 verbose cli [
0 verbose cli   '/usr/bin/node',
0 verbose cli   '/usr/lib/node_modules/npm/bin/npm-cli.js',
0 verbose cli   'exec',
0 verbose cli   '--',
0 verbose cli   'lerna',
0 verbose cli   'bootstrap'
0 verbose cli ]
1 info using npm@8.3.2
2 info using node@v17.3.0
3 timing npm:load:whichnode Completed in 0ms
4 timing config:load:defaults Completed in 2ms
5 timing config:load:file:/usr/lib/node_modules/npm/npmrc Completed in 1ms
6 timing config:load:builtin Completed in 1ms
7 timing config:load:cli Completed in 2ms
8 timing config:load:env Completed in 2ms
9 timing config:load:file:/home/st/Dokumente/privat/freesewing/.npmrc Completed in 2ms
10 timing config:load:project Completed in 2ms
11 timing config:load:file:/home/st/.npmrc Completed in 0ms
[...]

PR #1710, opened 2022-01-23
add NoteJS version to the quickstart instructions

stsydow:
See #1708


joostdecock:
@all-contributors please add @stsydow for docs

PR #1715, opened 2022-01-24
update(yml) Holmes. Crown not Semi-sphere

bobgeorgethe3rd:
Semi-sphere does not indicate properly what part of the pattern this options affects.

### DIFF
--- a/packages/i18n/src/locales/en/options/holmes.yml
+++ b/packages/i18n/src/locales/en/options/holmes.yml
@@ -7,7 +7,7 @@ lengthRatio:
   description: Controls the length of the crown and ear flaps
 gores:
   title: Number of gores
-  description: The number of gores used to construct the semi-sphere
+  description: The number of gores used to construct the crown
 visorAngle:
   title: Visor angle
   description: The arc angle used to draft the inner curve of the visor


packages/i18n/src/locales/en/options/holmes.yml
```diff
@@ -7,7 +7,7 @@ lengthRatio:
   description: Controls the length of the crown and ear flaps
 gores:
   title: Number of gores
-  description: The number of gores used to construct the semi-sphere
+  description: The number of gores used to construct the crown
```
nicholasdower, line 10: Never heard of a "gore" before!

This change makes perfect sense to me and also seems in line with [the documentation page](https://freesewing.org/docs/patterns/holmes/options/gores/) for this option:
> Controls the number of panels there are in the crown

nicholasdower (approved):

packages/i18n/src/locales/en/options/holmes.yml
bobgeorgethe3rd: Gore in this case is just triangle-shaped. It's not really the right term in the context of a hat and sewing but it makes sense for the code etc,

PR #1716, opened 2022-01-24
update(yml, docs, config) Bee

bobgeorgethe3rd:
Was originally suppsoe to be just a little tidy up and adding yml info has divulged a little

Additions
- Add missing options to bee.yml
- Added more info to Show bella Doc
- Added more info to Cross Back Ties Doc

Changes
- Changed Cross Back Ties opening scentence doc
- Changed Cross Back Ties in yml
- Changed fullChestEaseReduction to a slider
- Fixed some typos

I am debating whether it is prudent to add links from the Bee options to their corresponding Bella options though? If it is I'll get right on it.

### DIFF
--- a/markdown/org/docs/patterns/bee/options/bellaguide/en.md
+++ b/markdown/org/docs/patterns/bee/options/bellaguide/en.md
@@ -4,6 +4,14 @@ title: Show Bella
 
 Shows the outline of the Bella block Bee is based on
 
+This is to help give context to the other options of Bee especially when using the fit options
+
+<Note>
+
+It is recommended to have this on when adjusting the options for this pattern but also to turn it off before printing 
+
+</Note>
+
 
 ## Effect of this option on the pattern
 ![This image shows the effect of this option by superimposing several variants that have a different value for this option](bee_bellaguide_sample.svg "Effect of this option on the pattern")
--- a/markdown/org/docs/patterns/bee/options/crossbackties/en.md
+++ b/markdown/org/docs/patterns/bee/options/crossbackties/en.md
@@ -2,8 +2,23 @@
 title: Cross back ties
 ---
 
-A variation of Bee, where the neck ties cross and tie at the back with a band that replaces the band tie
+A variation of Bee, where the neck ties cross and tie in the back by looping into a band that replaces the band tie
 
+#### Default
+
+- 2 neck ties
+- 1 band tie
+
+The neck ties are sewn into the cups and tie at the neck like a halter.  
+The band tie is thread through casings in the cups and ties with itself at the back
+
+#### Cross Back Ties
+
+- 2 neck ties
+- No band tie
+
+The band tie and casing on the cups are replaced by a band which are sewn with loops in the back.  
+The neck ties are longer and cross over each other in the back, then go through the loops in the band and then tie with one another.
 
 ## Effect of this option on the pattern
 ![This image shows the effect of this option by superimposing several variants that have a different value for this option](bee_crossbackties_sample.svg "Effect of this option on the pattern")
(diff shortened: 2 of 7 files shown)


markdown/org/docs/patterns/bee/options/bellaguide/en.md
```diff
@@ -4,6 +4,14 @@ title: Show Bella
 
 Shows the outline of the Bella block Bee is based on
 
+This is to help give context to the other options of Bee especially when using the fit options
```
nicholasdower, line 7: This is super cool. I hadn't seen options like this before.

bobgeorgethe3rd: Thank you :) 
Yea it evolved from the guide that I was using originally to code Bee and I kinda felt like it should remain as Bee is so different from Bella that the options might not make much sense without it.

markdown/org/docs/patterns/bee/options/crossbackties/en.md
```diff
@@ -2,8 +2,23 @@
...
+
+The neck ties are sewn into the cups and tie at the neck like a halter.  
+The band tie is thread through casings in the cups and ties with itself at the back
+
+#### Cross Back Ties
+
+- 2 neck ties
+- No band tie
+
+The band tie and casing on the cups are replaced by a band which are sewn with loops in the back.
```
nicholasdower, line 20:
```suggestion
The band tie and casing on the cups are replaced by a band which is sewn with loops in the back.  
```

packages/i18n/src/locales/en/options/bee.yml
```diff
@@ -72,4 +72,28 @@ neckTieColours:
 
 crossBackTies:
   title: Cross back ties
-  description: Whether you'd like to cross the ties at the back or not
+  description: Whether you'd like to use the cross back tie variation of Bee
+
+bandLength:
```
nicholasdower, line 77: This part surprised me since I couldn't find these options when configuring Bee. Sorry if I am missing something.

packages/i18n/src/locales/en/options/bee.yml
```diff
@@ -72,4 +72,28 @@ neckTieColours:
...
+  title: Band Length (Cross back ties)
+  description: Controls the length of the band around your chest for the cross back ties variation of Bee
+
+backDartHeight:
```
nicholasdower, line 81: This part surprised me since I couldn't find these options when configuring Bee. Sorry if I am missing something.

nicholasdower: Sorry, I added this comment on the wrong line originally and had to delete and re-add it.

nicholasdower (approved):
Looks great!

packages/i18n/src/locales/en/options/bee.yml
bobgeorgethe3rd: They are in the advanced section. Due to Bee being made out of strectch fabrics and a lot of it's fit is to do with gathering the default Bella for the most part doesn't affect Bee but these are the few that do so they are added in case someone has settings they want to bring across from Bella.

nicholasdower: Oh I see it now. And this change makes those options prettier with real names and descriptions. Makes sense. Thanks for explaining.

bobgeorgethe3rd: That's alright. Since I had done Holmes yesterday I thought I would update Bee as well. It's sort of a distraction from some of the tougher docs work I am working on. Plus now Holmes is one step closer to full completion which makes me happy :)

joostdecock:
This PR exposes Bella's `fullchesteasereduction` option in Bee. But it neglects to add that to the documentation.
I've addressed this in 115d4bc but thought it would be good to indicate that this causes an issue.

If you're interested, you can run `node scripts/optiondocsimages.js` in the monorepo root and it will complain about this.
We should perhaps automate that check with a Github action so it gets picked up on PRs.

@nicholasdower since yo have a lot of experience weth actions expert, do you think that can be done?

PR #1717, opened 2022-01-24
Fix broken gore image link.

nicholasdower:
https://freesewing.org/docs/patterns/holmes/options/gores
image


bobgeorgethe3rd (approved):
Good catch! Yea that's a missed update from when the option changed for the recent changes to the gore-plugin

Issue #1718, opened 2022-01-24
Missing Walburga and Tiberius Github Labels


bobgeorgethe3rd:
Noticed that these two were missing their github labels. I would add them but I have no idea how.

starfetch:
Done :)

starfetch:
For reference: Via the github UI you can click on `labels`, and at the very bottom is `Edit labels`, which takes you to a big list with all labels. Some rights may be needed though to access it.

nicholasdower:
I also wondered how the auto-label feature worked. Added some missing ones to https://github.com/freesewing/freesewing/pull/1719

nicholasdower:
> I also wondered how the auto-label feature worked. Added some missing ones in #1719

I'm not sure that config is used after all. I think the real config is in `.github/boring-cyborg.yml`. And that contains the ones mentioned here.

joostdecock:
The old labeler was superseded by the boring-cyborg one which has more features (like welcoming new contributors anf so on). It will add labels, but won't create them. So when a new package/pattern is added, a label for it should be created.

So thanks for doing that @starfetch

bobgeorgethe3rd:
Awesome and yea by looks of it I don't have the permission

PR #1719, opened 2022-01-24
Add missing labeler config.

nicholasdower:
See #1718


nicholasdower:
I'm now thinking this isn't even used!
I think the config is in `.github/boring-cyborg.yml`

PR #1722, opened 2022-01-25
Remove obsolete Labeler config.

nicholasdower:
Assuming this is no longer used now that `.github/boring-cyborg.yml` exists. See [this comment](https://github.com/freesewing/freesewing/issues/1718#issuecomment-1020437476).


nicholasdower:
Interesting. There is also a file `label.yml` that controls the workflow. I've never used workflows before so I am learning as I go. Maybe I need to delete that file or maybe this is used after all. Will dig a little.

joostdecock:
See: https://github.com/marketplace/boring-cyborg-add-labels-to-prs-based-on-filepaths

nicholasdower:
I removed the workflow to see what would happen. Going to try modifying the cyborg config to see whether it is still working.  Also will learn whether it uses the config from my branch or from the base branch.

nicholasdower:
nah, it added the aaron label even though I changed the config to use bee. Rolled back.

nicholasdower:
Good news is it did still add the label. Bad news is I don't know for sure whether it will still work after this PR is merged. I assume it will though since I don't see anything in the docs about using a workflow with the boring cyborg.

nicholasdower:
Going to try merging this then opening in an example PR to test that labels are still being applied.

nicholasdower:
Seems to still be working: https://github.com/freesewing/freesewing/pull/1728. Going to assume this was safe.

PR #1748, opened 2022-01-27
Add some plugin unit tests.

nicholasdower:
From https://github.com/freesewing/freesewing/pull/1521

FYI: I left out several tests like "Should set the plugin name:version attribute" because I found that there is similar test included by default in the "Shared Plugin Tests". See [tests/plugins/shared.mjs](https://github.com/freesewing/freesewing/blob/develop/tests/plugins/shared.mjs).

### DIFF
--- a/packages/plugin-bartack/tests/plugin.test.mjs
+++ b/packages/plugin-bartack/tests/plugin.test.mjs
@@ -1,6 +1,6 @@
 import chai from 'chai'
 import freesewing from '@freesewing/core'
-import plugin from '../dist/index.js'
+import plugin from '../dist/index.mjs'
 
 const expect = chai.expect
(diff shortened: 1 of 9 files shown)


nicholasdower:
Several builds were failing on this PR with:
```
Error: Cannot find module '@babel/preset-env'
```

This error also occurred on the `develop` branch without the changes in this PR.

I found that the failing plugins all had the following in their tests:
```javascript
import plugin from '../dist/index.js'
```

But all of the passing plugins had:
```javascript
import plugin from '../dist/index.mjs'
```

I added a commit to change this and the build went green.

nicholasdower:
Going to merge this. Happy to follow up if there are any issues. Thanks again to @chri5b for the tests!

PR #1749, opened 2022-01-27
Add missing deps to plugintest (I think/hope)

nicholasdower:
The CI build failed for `packages/plugintest` on #1726 due to:
```shell
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@freesewing/plugin-banner' imported from /home/runner/work/freesewing/freesewing/packages/plugintest/dist/index.mjs
```

I was able to repro locally by following the steps CI takes. Adding the dependencies referenced in [packages/plugintest/src/index.js](https://github.com/freesewing/freesewing/blob/develop/packages/plugintest/src/index.js) to `package.json` resolved the failure for me locally.


joostdecock:
This would get overwritten as dependencies as centrally managed in `config/dependencies.yaml` and `package.json` is auto-generated.

As it happens, I did add these dependencies, but [I forgot to include the correct indentation](https://github.com/freesewing/freesewing/commit/5049528a4027462d2fb54e661ff1984460ed4044).

Side note, these should go under peer dependencies as they are required for the pattern (and not just required for the development environment.
In general the rule is:

- Dependencies: Will get bundled/shipped with the package (eg: `bezier-js` in `@freesewing/core`)
- Peer Dependencies: Required, but not bundled. This is a use-case common in libraries. It would be very inefficient if every library would install its own copy of some commonly used package. So peer dependency means: We need this, but you need to install it yourself. In rollup we use the `peer-deps-external` plugin to automatically keep them out of the bundle. Eg: all sorts of plugins in freesewing patterns
- Dev Dependecies: Needed for the development environment. Eg: rollup in freesewing patterns

Sorry if you know all this. Just trying to explain where I'm coming from.

nicholasdower:
> This would get overwritten as dependencies as centrally managed in `config/dependencies.yaml` and `package.json` is auto-generated.
> 
> As it happens, I did add these dependencies, but [I forgot to include the correct indentation](https://github.com/freesewing/freesewing/commit/5049528a4027462d2fb54e661ff1984460ed4044).
> 
> Side note, these should go under peer dependencies as they are required for the pattern (and not just required for the development environment. In general the rule is:
> 
> * Dependencies: Will get bundled/shipped with the package (eg: `bezier-js` in `@freesewing/core`)
> * Peer Dependencies: Required, but not bundled. This is a use-case common in libraries. It would be very inefficient if every library would install its own copy of some commonly used package. So peer dependency means: We need this, but you need to install it yourself. In rollup we use the `peer-deps-external` plugin to automatically keep them out of the bundle. Eg: all sorts of plugins in freesewing patterns
> * Dev Dependecies: Needed for the development environment. Eg: rollup in freesewing patterns
> 
> Sorry if you know all this. Just trying to explain where I'm coming from.

Oh I didn't know it! :)

I once edited this central config but I think it was an accident that I found it. I didn't realize that it was used to generate package.json. Thanks for taking the time to explain, especially about the types of dependencies. I hadn't figured that out yet either.

Issue #1656, opened 2022-01-15
Remove SASS prebuild step from plugins by migrating to plain CSS


joostdecock:
**What is it that you would like to see happen?**

As discussed in #1649 we'd like to avoid the SASS prebuild step by converting the paperless (and sample) style to plain CSS in plugin-theme.

joostdecock:
Implemented by @nicholasdower :tada:

PR #1750, opened 2022-01-27
Remove Sass from plugin-theme.

nicholasdower:
See #1656

Tested this by loading the change in `plugintest`. I could see my changes to paperless in the exported svg.


joostdecock:
Love getting rid of things :)

PR #1755, opened 2022-01-28
Update backend example.env and README.

nicholasdower:
I had some trouble getting the backend running locally with `docker-compose up`. The first error was about missing `FS_JWT_ISSUER`. The second was a failure to connect to the local database.


joostdecock:
Yeah, those docs are a bit outdated. Let's keep this issue open and try to give them some TLC.

joostdecock:
Don't know why I thought this was an issue. Probably because I was doing several things at once :see_no_evil: 

Go ahead and merge this. We can create an issue, but given that I don't think we'll have time for it soon, perhaps we'll just keep it in mind.

PR #1768, opened 2022-01-29
Add deps from some Simon parts to back part.

nicholasdower:
I like to use the custom contents option but it doesn't work for all parts in Simon. I believe this is due to the fact that the back part sets values in the Store which other parts need. I first thought about setting these values in shared but some of them are complex, for example backArmholeToArmholePitch.


nicholasdower:
Tried to add a test and found this impacts many patterns. Closing and will reopen when I have all tests passing.

nicholasdower:
Now in https://github.com/freesewing/freesewing/pull/1771.

PR #1770, opened 2022-01-29
Updated tests for individual part drafting (aka the content option)

joostdecock:
This is inspired by @nicholasdower 's work on #1768

 - Step 1: Realize we should have unit tests to check that all individual patterns draft
 - Step 2: Realize you already added such tests
 - Step 3: Realize you configured them wrong
 
Submitting this as a PR because this breaks a bunch of tests. And while that is a good thing (1) it would not be great of we're stuck for the foreseeable future with a bunch of tests failing and people shrugging it off because *it's those `only` tests*

(1) Good in the sense that we can start chipping away at fixing those patterns that don't have their part dependencies properly configured (which is always the reason for this to fail)

**Does this catch all problems?**

Nick mentioned he wanted to use this with Simon, and it did not work as expected. Yet simon does not throw any errors.
Which makes me wonder if there's perhaps more to this than the draft failing.

Perhaps he can clarify what he expected and what happened instead.

### DIFF
--- a/packages/css-theme/src/components/_pattern.scss
+++ b/packages/css-theme/src/components/_pattern.scss
@@ -72,6 +72,12 @@ svg.freesewing.pattern {
   .stroke-2xl {
     stroke-width: calc(var(--freesewing-pattern-scale) * 4);
   }
+  .stroke-3xl {
+    stroke-width: calc(var(--freesewing-pattern-scale) * 6);
+  }
+  .stroke-4xl {
+    stroke-width: calc(var(--freesewing-pattern-scale) * 10);
+  }
 
   .sa {
     stroke-dasharray: 1, 3;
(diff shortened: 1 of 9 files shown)


joostdecock:
Oops. Didn't realize I had a commit from late last night in here :see_no_evil:

nicholasdower:
@joostdecock I was also trying to tackle this here: https://github.com/freesewing/freesewing/compare/nickd/simon-deps#files_bucket

nicholasdower:
The problem seems to be that we aren't specifying the only setting correctly and that we aren't calling render. I made those changes and attempted to run all tests. Many failed so I am fixing them one by one.

Am I on the right track?

nicholasdower:
All passing locally now. Opened https://github.com/freesewing/freesewing/pull/1771.

PR #1771, opened 2022-01-29
Update some pattern deps. Add render test.

nicholasdower:
Followup to: https://github.com/freesewing/freesewing/pull/1768
See also: https://github.com/freesewing/freesewing/pull/1770


nicholasdower:
@joostdecock no hard feelings if these dependency changes are no good. Happy to spend more time on this to make it right, or alternatively, tackle these one by one by only running the new render test in patterns which are known to be good. lmk

tests/patterns/drafting.mjs
```diff
@@ -127,7 +125,7 @@ export const testPatternDrafting = (design, Pattern, expect, models, patterns, l
     for (const sa of [0,10]) {
       it(`  - Drafting paperless non-detailed pattern for size-40 (${breasts ? 'with' : 'no'} breasts) sa: ${sa}`, () => {
         expect(
-          doesItDraft(
+          doesItDraftAndRender(
```
joostdecock, line 128: This entire `settings` prop needs to go. It's there for god knows what reason, but has no effect.

joostdecock: Note that this remark is about the lines below, not what you changed.

joostdecock (approved):
We should get rid of this block of code because it's dead wood : https://github.com/freesewing/freesewing/blob/243cf8efa08a2dc5491d3db011bdcdbd4f02b9f5/tests/patterns/drafting.mjs#L136-L140

LGTM

tests/patterns/drafting.mjs
nicholasdower: Gone!

PR #1779, opened 2022-01-30
update(docs) Tidy up Pattern Options after full-split

bobgeorgethe3rd:
Yay the split worked! And then oh dear there is a lot of tidying up to do now that I can see the correct pattern options on the surface. For the most part this is just removing defunct brian diagrams but then you get to Jaeger and it is adding some of the docs in, which I won't say is perfect but better than nothing. There was a sneaky Wahid I hadn't noticed before but then you get to the final boss of Yuri. I don't exactly know why but poor Yuri had a lot of Huey's sample images in it's folders and in it's en.md files so those have now been fixed. I have left the collar info for Carlton and Carlita as that is planned in their documentation update.


joostdecock:
Good work @bobgeorgethe3rd 👍

I suspect when we initially de-coupled the shared options, things got copied over without thinking about the diagram names.

Issue #1723, opened 2022-01-25
Provide a standard way for conditional display of options in the frontend


joostdecock:
This is feature request from [our roadmap](https://github.com/freesewing/freesewing/discussions/1278).

The idea is that we would hide options from the user/frontend/UI when those options are not relevant.
For example: If you say you don't want back pockets on charlie, there's no point in asking whether the user wants back pocket facing.

**How would it work**

We can't magically figure it out, so somehow this should be part of the option configuration.
I also feel it's too complex to put this into some DSL, so what I think makes most sense is a method that can be called and will return the answer to the question: *should we show you?*. 
To make that decision, the method should receive the current pattern options/config.

**Example**

This is from the top of my head, but something like:

```js
options: {
  myAwesomeOption: {
    pct: 50,
    min: 0,
    max: 100,
    show: function(settings) => {
      if (settings.options.myOtherOption === 2) return true
      else return false
    }
  }
}
```

Or, with fat arrow notation and object destructuring:

```js
options: {
  myAwesomeOption: {
    pct: 50,
    min: 0,
    max: 100,
    show: ({ options }) => (options.myOtherOption === 2) ? true : false
  }
}
```

Then in the frontend, we could check if the method exists on the option, and only show it if it either does not exists, or returns true.

PS: Naming things is hard, so perhaps `show` is not the best name? 

**First steps**

We could try this out with one pattern, and then integrate it into the development environment.

bobgeorgethe3rd:
Mm this would be very useful. I know for a fact Bee will benefit from this and clear up the confusion between the Default and Cross Back Ties variations.

joostdecock:
While not all frontend code is updated, the *standard way* as requested in this issue is implemented. Closing.

PR #1798, opened 2022-02-02
Enable conditional display of options (Use in Simon)

nicholasdower:
See #1723

conditional-options-simon

- Update Simon to always honor `seperateButtonPlacket` and `seperateButtonholePlacket`.
- Update `DraftConfig` to pass the pattern to `DraftConfigurator` so that it can be used when configuring options to display.
- Update `OptionGroup` to hide options based on configuration.
- Update Simon to hide `buttonPlacketStyle` and `buttonholePlacketStyle` based on `seperateButtonPlacket` and `seperateButtonholePlacket`.

If this approach ends up being reasonable, I think the next steps will be to:
- Add tests (I couldn't figure out where to do this).
- Maybe add this functionality to other patterns which would benefit.
- Update the options docs to describe this functionality.

### DIFF
--- a/packages/components/src/DraftConfigurator/OptionGroup/index.js
+++ b/packages/components/src/DraftConfigurator/OptionGroup/index.js
@@ -66,13 +66,19 @@ const OptionGroup = (props) => {
 
   return (
     <>
-      {props.options.map((name) => {
+      {props.options.map(nameOrGroup => {
         let output = []
-        if (typeof name === 'object') {
+        if (typeof nameOrGroup === 'object') {
+          const group = nameOrGroup
           // Subgroup
-          for (let subGroup of Object.keys(name)) {
+          for (const subGroup in group) {
             let children = []
-            for (let option of name[subGroup]) children.push(renderOption(option, true))
+            for (const name of group[subGroup]) {
+              const option = props.config.options[name]
+              if (!option.hide(props.pattern.settings.options)) {
+                children.push(renderOption(name, true))
+              }
+            }
             output.push(
               <OptionSubGroup
                 title={<FormattedMessage id={'optiongroups.' + subGroup} />}
@@ -81,7 +87,13 @@ const OptionGroup = (props) => {
               />
             )
           }
-        } else output.push(renderOption(name))
+        } else {
+          const name = nameOrGroup
+          const option = props.config.options[name]
+          if (!option.hide(props.pattern.settings.options)) {
+            output.push(renderOption(name))
+          }
+        }
 
         return output
       })}
--- a/packages/simon/config/index.js
+++ b/packages/simon/config/index.js
@@ -167,10 +167,12 @@ export default {
     buttonPlacketStyle: {
       list: ['classic', 'seamless'],
       dflt: 'classic',
+      hide: options => options.seperateButtonPlacket
     },
     buttonholePlacketStyle: {
       list: ['classic', 'seamless'],
       dflt: 'seamless',
+      hide: options => options.seperateButtonholePlacket
     },
     cuffStyle: {
       list: [
--- a/packages/simon/src/frontleft.js
+++ b/packages/simon/src/frontleft.js
@@ -15,9 +15,13 @@ export default (part) => {
     })
   }
 
-  return options.buttonholePlacketStyle === 'seamless'
-    ? frontLeftSeamless(part)
-    : options.seperateButtonholePlacket
-    ? frontLeftClassicSeperate(part)
-    : frontLeftClassicCuton(part)
+  if (options.seperateButtonholePlacket) {
+    return frontLeftClassicSeperate(part)
+  } else if (options.buttonholePlacketStyle === 'seamless') {
+    return frontLeftSeamless(part)
+  } else if (options.buttonholePlacketStyle === 'classic') {
+    return frontLeftClassicCuton(part)
+  } else {
+    throw `Unexpected buttonholePlacketStyle: ${options.buttonholePlacketStyle}`
+  }
 }
--- a/packages/components/src/Workbench/DraftConfig/index.js
+++ b/packages/components/src/Workbench/DraftConfig/index.js
@@ -122,6 +122,7 @@ const DraftPattern = (props) => {
         updatePatternData={props.updateGist}
         raiseEvent={props.raiseEvent}
         freesewing={props.freesewing}
+        pattern={props.pattern}
         units={props.units || 'metric'}
       />
       <li className='zoombox'>
--- a/packages/simon/src/buttonholeplacket.js
+++ b/packages/simon/src/buttonholeplacket.js
@@ -17,7 +17,7 @@ export default (part) => {
     store,
   } = part.shorthand()
 
-  if (!options.seperateButtonholePlacket || options.buttonholePlacketStyle !== 'classic') {
+  if (!options.seperateButtonholePlacket) {
     part.paths = {}
     part.snippets = {}
     part.points = {}
--- a/packages/simon/src/buttonplacket.js
+++ b/packages/simon/src/buttonplacket.js
@@ -17,7 +17,7 @@ export default (part) => {
     store,
   } = part.shorthand()
 
-  if (!options.seperateButtonPlacket || options.buttonPlacketStyle !== 'classic') {
+  if (!options.seperateButtonPlacket) {
     part.paths = {}
     part.snippets = {}
     part.points = {}
--- a/packages/simon/src/frontright.js
+++ b/packages/simon/src/frontright.js
@@ -23,9 +23,13 @@ export default (part) => {
     }
   }
 
-  return options.buttonPlacketStyle === 'seamless'
-    ? frontRightSeamless(part)
-    : options.seperateButtonPlacket
-    ? frontRightClassicSeperate(part)
-    : frontRightClassicCuton(part)
+  if (options.seperateButtonPlacket) {
+    return frontRightClassicSeperate(part)
+  } else if (options.buttonPlacketStyle === 'seamless') {
+    return frontRightSeamless(part)
+  } else if (options.buttonPlacketStyle === 'classic') {
+    return frontRightClassicCuton(part)
+  } else {
+    throw `Unexpected buttonPlacketStyle: ${options.buttonPlacketStyle}`
+  }
 }


nicholasdower:
Added a gif. Sorry it is sooooo sloooooow!

joostdecock:
Hey @nicholasdower I'm sorry for the lack of feedback here. My day job has been all-consuming this week and I might need to defer this to the weekend. Sorry 🙏🙈

nicholasdower:
> Hey @nicholasdower I'm sorry for the lack of feedback here. My day job has been all-consuming this week and I might need to defer this to the weekend. Sorry 🙏🙈

@joostdecock one of the many nice things about this project is that it isn't work and there isn't any pressure. Thanks for taking a look whenever you have time.

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -66,13 +66,19 @@ const OptionGroup = (props) => {
 
   return (
     <>
-      {props.options.map((name) => {
+      {props.options.map((nameOrGroup) => {
         let output = []
```
joostdecock, line 70: You're a brave man for wading into this mess ;)

I'm putting this comment on a code line you did not touch, because it is something that my older code is very guilty of, but: I find code is easier to read when you only use `let` for variables that need to change. And `const` for everything else.

So I'm going to add comments each time I see a `let` that could/should be a `const` knowing full well that I have (at least historically) not been very good at that either.

So see this more as a generic code style feedback that I want to share with you, and not as criticism

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -66,13 +66,19 @@ const OptionGroup = (props) => {
 
   return (
     <>
-      {props.options.map((name) => {
+      {props.options.map((nameOrGroup) => {
```
joostdecock, line 69: These brackets around `nameOrGroup` are not required here. Not wrong, but this is fine too:

```js
{props.options.map(nameOrGroup => {
```

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -66,13 +66,19 @@ const OptionGroup = (props) => {
...
         let output = []
-        if (typeof name === 'object') {
+        if (typeof nameOrGroup === 'object') {
+          let group = nameOrGroup
```
joostdecock, line 72: Can be a `const`

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -66,13 +66,19 @@ const OptionGroup = (props) => {
...
           // Subgroup
-          for (let subGroup of Object.keys(name)) {
+          for (let subGroup of Object.keys(group)) {
```
joostdecock, line 74: Can be a `const` (yes, even in for loops :) Something I learned later in life)

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -66,13 +66,19 @@ const OptionGroup = (props) => {
...
             let children = []
-            for (let option of name[subGroup]) children.push(renderOption(option, true))
+            for (let name of group[subGroup]) {
```
joostdecock, line 76: Can be a `const`

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -66,13 +66,19 @@ const OptionGroup = (props) => {
...
+              let option = props.config.options[name]
```
joostdecock, line 77: Can be a `const`

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -66,13 +66,19 @@ const OptionGroup = (props) => {
...
+              if (!(option.hide instanceof Function) || !option.hide(props.pattern.settings.options)) {
```
joostdecock, line 78: While the goal is to hide options when they're not needed, is it just me or is having an (optional) `show()` method on the option more intuitive than a `hide()` method?

It would be the same thing, just returning false if it should be hidden. But just `return false from show() to not show` instead of `return true from hide() to hide`.

There's obviously no right or wrong. Just thinking out loud.

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -81,7 +87,13 @@ const OptionGroup = (props) => {
               />
             )
           }
-        } else output.push(renderOption(name))
+        } else {
+          let name = nameOrGroup
```
joostdecock, line 91: Can be a `const`

packages/components/src/DraftConfigurator/OptionGroup/index.js
```diff
@@ -81,7 +87,13 @@ const OptionGroup = (props) => {
...
+          let option = props.config.options[name]
```
joostdecock, line 92: Can be a `const`

packages/simon/config/index.js
```diff
@@ -167,10 +167,16 @@ export default {
     buttonPlacketStyle: {
       list: ['classic', 'seamless'],
       dflt: 'classic',
+      hide: (options)  => {
+        return options.seperateButtonPlacket
+      }
```
joostdecock, line 172: Perhaps is a coding style preference, but a fat arrow function that immediately returns does not need curly braces.
This is equivalent:
```js
hide: options => options.seperateButtonPlacket
```

joostdecock (commented):
I've made some comments about the code style, but I don't particularly care too much about those.

The only thing I feel some unease with is the use of `hide()` when I for some reason had `show()` in my head.
It's very likely I just need some time to adapt to this, but I wanted to hear what you had to say about it too.

joostdecock:
I was thinking that we could save ourselves a lot of checks if core would just set a `show()` (or `hide()`) method on each option.

Something like this (but with a check for options that are not objects (constants):

```js
for (const key in options) options[key] = {
  show: () => true,
  ...options[key]
}
```

or:
```js
for (const key in options) options[key] = {
  hide: () => false,
  ...options[key]
}
```

That would save us the trouble of checking *does it have a show/hide method* in the various places that need that check and if you something like `if(option.show())` or `if (!option.hide())`

Thoughts?

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: Done

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: Sounds great. I think I actually improved a bit on this since I sent this PR. My latest PR is full of `const`!

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: Done

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: I tried a few iterations of this and had some trouble. Thanks for the advice.

nicholasdower: Done/

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: Done

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: Done

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: I don't have a strong preference and think show looks great. Changed.

My original reason for using hide is that the default is to show. So any show function that is defined is only there to return false. So the explanation of when to implement show is when you want to hide something. Somehow that felt backward to me.

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: Done

packages/components/src/DraftConfigurator/OptionGroup/index.js
nicholasdower: Done

packages/simon/config/index.js
nicholasdower: It wasn't my preference! I just didn't know. I like the one-liner, thanks!
Done.

packages/simon/src/frontleft.js
```diff
@@ -15,9 +15,13 @@ export default (part) => {
...
-    : frontLeftClassicCuton(part)
+  if (options.seperateButtonholePlacket) {
+    return frontLeftClassicSeperate(part)
+  } else if (options.buttonholePlacketStyle === 'seamless') {
+    return frontLeftSeamless(part)
+  } else if (options.buttonholePlacketStyle === 'classic') {
+    return frontLeftClassicCuton(part)
+  } else {
+    throw `Unexpected buttonholePlacketStyle: ${options.buttonholePlacketStyle}`
+  }
```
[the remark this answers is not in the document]
nicholasdower: I thought this part deserved to be a bit more verbose. But yeah, I think I tend toward that. I also really wanted to throw otherwise and that didn't work the first time I tried with the ternary. Maybe I just needed some parens.

nicholasdower:
@joostdecock thanks for the thorough review!

I made all the suggested changes but haven't tested yet. In case it looks ok, please don't merge.

nicholasdower:
> I was thinking that we could save ourselves a lot of checks if core would just set a default `show()` (or `hide()`) method on each option.
> 
> Something like this (but with a check for options that are not objects (constants):
> 
> ```js
> for (const key in options) options[key] = {
>   show: () => true,
>   ...options[key]
> }
> ```
> 
> or:
> 
> ```js
> for (const key in options) options[key] = {
>   hide: () => false,
>   ...options[key]
> }
> ```
> 
> That would save us the trouble of checking _does it have a show/hide method_ in the various places that need that check and if you something like `if(option.show())` or `if (!option.hide())`
> 
> Thoughts?

Oh just saw this comment. I'd much prefer this. Will look for a good place to put this code.

joostdecock:
Actually, ignore that comment. That code runs when the pattern is instantiated.

But the config is statically exported, so it should happen in de Design constructor: https://github.com/freesewing/freesewing/blob/develop/packages/core/src/design.js

joostdecock:
I've submitted #1809 for your review. It adds a default the `hide()` method to options.

joostdecock:
I also woke up this morning in agreement with you initial proposal that `hide()` is the more sensible choice (in comparison to `show()`).

So that's what I've implemented, but I do feel bad since you've adapted to `show()` and I guess it looks like I'm trolling you :trollface: but I just need some time to let things sink in from time to time. If that's frustrating, I can hold off giving feedback until my thoughts are fully formed :see_no_evil:

nicholasdower:
> I also woke up this morning in agreement with you initial proposal that `hide()` is the more sensible choice (in comparison to `show()`).
> 
> So that's what I've implemented, but I do feel bad since you've adapted to `show()` and I guess it looks like I'm trolling you :trollface: but I just need some time to let things sink in from time to time. If that's frustrating, I can hold off giving feedback until my thoughts are fully formed 🙈

No way! I think it is much better if we feel comfortable to share ideas early. I'm very happy with the discussions on this PR.

joostdecock:
Thanks, I also much prefer it when I'm allowed to think out loud.

I think this is ready to merge now? :shipit:

nicholasdower:
> Thanks, I also much prefer it when I'm allowed to think out loud.
> 
> I think this is ready to merge now? :shipit:

I believe so. Just tested manually with your changes and everything looks great. I much prefer just being able to call hide without having to know it might not exist. Thanks for the collab! Merging.

PR #1799, opened 2022-02-02
Revert dateformat and remark-parse upgrades.

nicholasdower:
I am getting `ERR_REQUIRE_ESM` while trying to run `yarn develop` in `packages/backend`. I think it is due to these two libraries which have both recently been updated (#1682 & #1782).


nicholasdower:
I am able to start the backend now with this change plus building a new version of i18n (due to a missing translation which has since been added) plus updating `docker-compose.yml` to expose port 27017.

My hope was to be able to run the backend tests (which I think requires a running, local backend) but only a small number of them are passing. Most are failing with 404s and 401s.

joostdecock:
Guess I didn't have coffee when I merged that :roll_eyes: 

I've added both dependencies to the list of modules being held back because ESM: #1685 

I've let the backend code drift from the tests, I have no excuse :)
You'll need to run the backend locally (you can just `yarn develop` it) and you need a local mongo setup. I typically run the `run-mongo.sh` script, because I'm used to running a bunch of containers for various development purposes.

As such, the Dockerfile and docker-compose are a bit unmaintained because I don't use them and up until now that means nobody does ;)

That being said, the `yarn clear:reboot` script will create the required data in the local mongo instance to run the tests. So that's something.
The `testall` script also tests sending out emails. For that to work you need to run [mailhog](https://hub.docker.com/r/mailhog/mailhog/) locally. But I'm afraid that's not documented.

I'm sorry for not being more helpful, but these are areas of the code that are rarely visited by anyone but myself, so I always push off documenting it for more pressing matters.

joostdecock (approved):
LGTM

nicholasdower:
> Guess I didn't have coffee when I merged that 🙄
> 
> I've added both dependencies to the list of modules being held back because ESM: #1685
> 
> I've let the backend code drift from the tests, I have no excuse :) You'll need to run the backend locally (you can just `yarn develop` it) and you need a local mongo setup. I typically run the `run-mongo.sh` script, because I'm used to running a bunch of containers for various development purposes.
> 
> As such, the Dockerfile and docker-compose are a bit unmaintained because I don't use them and up until now that means nobody does ;)
> 
> That being said, the `yarn clear:reboot` script will create the required data in the local mongo instance to run the tests. So that's something. The `testall` script also tests sending out emails. For that to work you need to run [mailhog](https://hub.docker.com/r/mailhog/mailhog/) locally. But I'm afraid that's not documented.
> 
> I'm sorry for not being more helpful, but these are areas of the code that are rarely visited by anyone but myself, so I always push off documenting it for more pressing matters.

This is very helpful! `yarn clear:reboot` was the missing piece.

The Docker setup is working fine for me with a small change. I will take a stab at updating it and the README.

PR #1800, opened 2022-02-02
Enable tests in i18n. Add missing testci to core.

nicholasdower:
Randomly noticed there were tests in the i18n dir that couldn't be run. Also core was missing a testci script.


nicholasdower:
Actually I think I can just add a file like `.github/workflows/tests.i18n.yml`. Will try that tomorrow.

joostdecock (approved):
LGTM

PR #1802, opened 2022-02-03
sleeve grainline alteration

SeaZeeZee:
Grainline point altered so grainline is straight.


nicholasdower:
Looks great @SeaZeeZee! Thank you!

FYI your change only works for me if I revert the modification to `packages/simon/example/src/pattern`. Maybe that was included by mistake?

SeaZeeZee:
Yep you're right, let me change that!

nicholasdower:
Not sure whether I have the magic in me, but I think someone is supposed to say:

@all-contributors please add @SeaZeeZee for code

SeaZeeZee:
Closed it because I cant quite figure out what happened and why it wanted to push that second commit.

PR #1806, opened 2022-02-04
Update testall to report all failures. Add All Tests workflow.

nicholasdower:
I recently made a cross-cutting change and needed to run all tests in order to verify the change was safe. While doing this I noticed that the `testall` script exits after one or more tests fail in a single package. This means you must fix issues one package at a time, then rerun all tests to find the next failure. I tried using the `--no-bail` flag to avoid this problem, but that does not make it easy to find which tests failed as failures are buried in thousands of lines of logs.

In this change I am suggesting we do the following:
 - Use the `--no-bail` option by default (all tests only take 50 seconds to run).
 - Introduce a way to aggregate failures from all packages (writing to a file).
 - Print a summary of failures after all tests complete.
 - Introduce an All Tests workflow which essentially runs kickstart and testall.

The output from `yarn testall` would now look like the following when one or more tests fails:
image

Example success build: https://github.com/freesewing/freesewing/runs/5082640538
image

Example failure build: https://github.com/freesewing/freesewing/runs/5082693681
image

### DIFF
--- a/.github/workflows/tests.all.yml
+++ b/.github/workflows/tests.all.yml
@@ -1,0 +1,41 @@
+name: All Tests
+
+on:
+  push:
+    branches:
+      - develop
+    paths:
+      - '**'
+      - '!**/README.md'
+  pull_request:
+    branches:
+      - develop
+    paths:
+      - '**'
+      - '!**/README.md'
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install dependencies
+        run: npx lerna bootstrap
+        env:
+          CI: true
+      - name: Build all
+        run: npm run buildall
+      - name: Run all tests
+        run: npm run testall
+
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,5 @@ scripts/verdaccio.sh
 
 # Don't ignore the specific Points.dist() docs folder
 !markdown/dev/reference/api/point/dist
+
+.test-failures.log
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "reconfigure": "all-contributors generate && node scripts/reconfigure.js",
     "prerelease": "lerna version --no-git-tag-version --no-push && yarn reconfigure && yarn buildall",
     "buildall": "lerna run cibuild_step1 && lerna run cibuild_step2",
-    "testall": "lerna run testci",
+    "testall": "node scripts/testall.js",
     "release": "lerna exec -- npm publish",
     "postrelease": "git add . && git commit -m ':bookmark: v$npm_package_version' && git tag -a v$npm_package_version -m ':bookmark: FreeSewing v$npm_package_version'",
     "ship": "lerna exec --no-bail -- npm publish",
--- a/scripts/test-failure-collector.js
+++ b/scripts/test-failure-collector.js
@@ -1,0 +1,69 @@
+/*
+ * Used to collect test failures in a file. Use by specifying --file to Mocha.
+ *
+ * See https://mochajs.org/#command-line-usage
+ */
+
+const path = require('path');
+const projectRoot = path.normalize(path.join(__dirname, '..'));
+const outputLog = path.join(projectRoot, '.test-failures.log');
+
+const red = function(string) {
+  return `\x1b[31m${string}\x1b[0m`;
+};
+
+const green = function(string) {
+  return `\x1b[32m${string}\x1b[0m`;
+};
+
+const dim = function(string) {
+  return `\x1b[2m${string}\x1b[0m`;
+};
+
+// Mapping of test file name to array of failing tests.
+const failuresPerFile = {};
+
+afterEach(function () {
+  if (this.currentTest.state === "failed") {
+    failuresPerFile[this.currentTest.file] = failuresPerFile[this.currentTest.file] || [];
+    failuresPerFile[this.currentTest.file].push(this.currentTest);
+  }
+});
+
+after(function () {
+  if (Object.keys(failuresPerFile).length === 0) return;
+
+  const fs = require('fs')
+  const logger = fs.createWriteStream(outputLog, { flags: 'a' });
+  const writeLine = (line) => logger.write(`${line}\n`);
+
+  for (let file in failuresPerFile) {
+    const failures = failuresPerFile[file];
+
+    // Remove project root from file path to keep log lines shorter.
+    if (file.startsWith(projectRoot)) {
+      file = file.substr(projectRoot.length + 1, file.length - projectRoot.length - 1)
+    }
+
+    // Print each failure.
+    failures.forEach(function (failure, i) {
+      const stack = failure.err.stack.split('\n');
+      writeLine(`${file}:  ${i + 1}\) ${failure.title}:`);
+      writeLine(`${file}:`);
+      writeLine(`${file}:      ${red(stack[0].trim())}`);
+      writeLine(`${file}:      ${green('+ expected')} ${red('- actual')}`);
+      writeLine(`${file}:`);
+      writeLine(`${file}:      ${red("-" + failure.err.actual)}`);
+      writeLine(`${file}:      ${green("+" + failure.err.expected)}`);
+      writeLine(`${file}:`);
+      stack.slice(1).forEach(function (stackLine) {
+        writeLine(`${file}:      ${dim(stackLine.trim())}`);
+      });
+      if (i < failures.length - 1) {
+        writeLine(`${file}:`);
+      }
+    });
+  }
+
+  logger.end();
+});
--- a/scripts/testall.js
+++ b/scripts/testall.js
@@ -1,0 +1,25 @@
+const fs = require('fs')
+const path = require('path');
+const spawn = require('child_process').spawn
+
+const projectRoot = path.normalize(path.join(__dirname, '..'));
+const outputLog = path.join(projectRoot, '.test-failures.log');
+const collectorScript = path.join(projectRoot, 'scripts', 'test-failure-collector.js');
+
+// Start with a fresh output log on each run.
+if (fs.existsSync(outputLog)) {
+  fs.unlinkSync(outputLog);
+}
+
+// Run all tests, specifying the collector script.
+spawn('lerna', ['run', '--no-bail', 'testci', '--', '--file', `${collectorScript}`], { stdio: 'inherit' })
+  .on('exit', function(code) {
+    // If a failure occurred, the log file will have been created. Print it.
+    if (fs.existsSync(outputLog)) {
+      console.error(fs.readFileSync(outputLog, 'utf8').trim());
+    }
+
+    // Propagate the exit code.
+    process.exit(code);
+  });
+


joostdecock (commented):
First: I love this plan. It's actually something that has annoyed me at times, but I've never gotten around to dealing with it.

My only remark is that perhaps we should try to keep it pure JS?
What I mean is, we're now running a shell script. Will that make it harder for people on Windows?

Although I do realize that redirecting error makes this probably much easier to do in the shell than in JS :thinking:

nicholasdower:
> First: I love this plan. It's actually something that has annoyed me at times, but I've never gotten around to dealing with it.
> 
> My only remark is that perhaps we should try to keep it pure JS? What I mean is, we're now running a shell script. Will that make it harder for people on Windows?
> 
> Although I do realize that redirecting error makes this probably much easier to do in the shell than in JS 🤔

It never even occurred to me to write this in JavaScript! I guess because I started out by extending the shell script in package.json then when the line got too long I extracted it to a file.

Anyway, I took a shot at it. ptal. It wasn't that hard after all and was a really good learning experience. It also forced me to ditch the use of env variables which I didn't like. So yay!

nicholasdower:
btw, does much of this run on Windows? We use the wrong path separator everywhere, for instance:

```shell
"reconfigure": "all-contributors generate && node scripts/reconfigure.js",
```

and

```shell
"test": "BABEL_ENV=production ../../node_modules/.bin/_mocha tests/*.test.js --require @babel/register",
```

I was careful to avoid this issue in the new JavaScript version of testall. Maybe we can make similar changes to the other scripts that don't look like they should work on Windows.

joostdecock:
> btw, does much of this run on Windows? We use the wrong path separator everywhere, for instance:

Historically, I haven't given to much of a :hankey: about Windows support. 
I don't use it, there's not even any Windows machines in this household, and I have a non-trivial amount of PTSD from all the pain inflicted on me by Windows (and its advocates) in the workplace :wink: 

However, as more people are starting to wade in and contribute, I try to be more mindful of it.
So yeah, you are absolutely right that there's a lot that is not windows-friendly at the moment, and we'll have to do some refactoring to tackle some of that technical debt.

But, when writing new things I like to try to keep platform-agnostic. (typically by wrapping paths with `path.resove()` like in this nextjs config: https://github.com/freesewing/freesewing/blob/develop/packages/freesewing.shared/config/next.mjs

nicholasdower:
> > btw, does much of this run on Windows? We use the wrong path separator everywhere, for instance:
> 
> Historically, I haven't given too much of a 💩 about Windows support. I don't use it, there's not even any Windows machines in this household, and I have a non-trivial amount of PTSD from all the pain inflicted on me by Windows (and its advocates) in the workplace 😉
> 
> However, as more people are starting to wade in and contribute, I try to be more mindful of it. So yeah, you are absolutely right that there's a lot that is not windows-friendly at the moment, and we'll have to do some refactoring to tackle some of that technical debt.
> 
> But, when writing new things I like to try to keep platform-agnostic. (typically by wrapping paths with `path.resove()` like in this nextjs config: https://github.com/freesewing/freesewing/blob/develop/packages/freesewing.shared/config/next.mjs

Makes sense. Now I wish I had a Windows computer so I could make sure my changes actually work. Maybe some Window's lover will come and give freesewing some TLC.

nicholasdower:
I've added an All Tests workflow. It will take around 8 minutes. Things to consider:

- Is 8 minutes too long for a PR build?
- If we decide it is worth having an All Tests workflow, should we remove the plugin-specific and pattern-specific workflows since they would be redundant?

nicholasdower:
I'm trying out a failing build now by purposely breaking one of the tests.

The original successful build is here: https://github.com/freesewing/freesewing/runs/5082640538

joostdecock:
I think this is a great addition and perhaps all we need is some optimisation.

- 5.5 minutes is spent installing dependencies. Are there some yarn flags that could speed this up? Perhaps `yarn install --frozen-lockfile `
- the build does not take that long, but do we need it? We're running these tests in node 14, perhaps we can look into running the tests from the source directly rather than feom the transpieled build code.
- perhaps we could start tracking individual tests timing (per package). Tests for some patterns clock in around the 20 seconds mark. Perhaps it's worth looking at why they take so long and whether they can be optimized.

While thinking about this an outlandish idea popped in my mind: what if we started tracking some of these metrics over time?
Wouldn't be too hard to spin up an Elasticsearch container or use a hosted offering, and push some metrics to it. Then we could have pretty graphs are things like *this PR makes simon tests 0?5 seconds slower* or whatever.

It's perhaps silly and complete overkill. Buy it's perhaps also just a fun thing to do 🤔

nicholasdower:
> I think this is a great addition and perhaps all we need is some optimisation.
> 
> * 5.5 minutes is spent installing dependencies. Are there some yarn flags that could speed this up? Perhaps `yarn install --frozen-lockfile `

I tried adding this to `lerna.json` to see whether it sped up a clean build locally. Unfortunately it didn't. Trying it on this PR now but it will be hard to interpret the results since I am seeing a lot of variability in build times on GitHub.

Is it possible that `yarn install --frozen-lockfile` is the same as `npm ci`? I found this: https://stackoverflow.com/questions/63187000/what-is-the-npm-equivalent-of-yarn-install-frozen-lockfile

See also: https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

Lerna has a `--ci` flag for this. See: https://www.npmjs.com/package/@lerna/bootstrap

But, as far as I can tell this is the default in CI environments and already happening in our builds. From the logs:
```shell
lerna info ci enabled
```

Side note: we are specifying `CI=true` in the workflow configs but it looks like `CI=true` is now the default in GitHub actions. See https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/

> * the build does not take that long, but do we need it? We're running these tests in node 14, perhaps we can look into running the tests from the source directly rather than feom the transpieled build code.

This one is way over my head. Happy to look into this but I might need some tips on what to search for. That said, the build step is less than 10% of the total time. Is there any way to speed up the install? It accounts for more than half the time.

I tried playing around with scopes (e.g. `lerna boostrap --scope @freesewing/simon`) thinking that maybe I could split the build into shards. But this seems to have no impact on install times.

> * perhaps we could start tracking individual tests timing (per package). Tests for some patterns clock in around the 20 seconds mark. Perhaps it's worth looking at why they take so long and whether they can be optimized.
>
> While thinking about this an outlandish idea popped in my mind: what if we started tracking some of these metrics over time? Wouldn't be too hard to spin up an Elasticsearch container or use a hosted offering, and push some metrics to it. Then we could have pretty graphs are things like _this PR makes simon tests 0?5 seconds slower_ or whatever.
> 
> It's perhaps silly and complete overkill. Buy it's perhaps also just a fun thing to do 🤔

I think metrics would be cool.

nicholasdower:
It seems `yarn install --frozen-lockfile` did not help. I tried it in 184fda9b3c9580ac0a36ef7c136987dc330ccb56. Here is the build: https://github.com/freesewing/freesewing/runs/5083631191. Installing took 5:46 and running the tests took 2:21.

Compare that to an earlier build without this flag: https://github.com/freesewing/freesewing/runs/5082640538. In this case installing took 4:56 and running the tests took 2:12.

joostdecock:
Guess there's no magic solution to making the install faster. However, I do think we should see a speed-up once we migrate to the new development environment, as we'll essentially be able to rip out the old one.

Looking at a typical pattern `package.json` these are the dev dependencies:

```
"react": "^17.0.2",
"react-dom": "^17.0.2",
"@babel/plugin-proposal-class-properties": "^7.13.0",
"@freesewing/components": "^2.20.4",
"@freesewing/css-theme": "^2.20.4",
"@freesewing/i18n": "^2.20.4",
"@freesewing/mui-theme": "^2.20.4",
"@freesewing/plugin-bust": "^2.20.4",
"@freesewing/plugin-buttons": "^2.20.4",
"@freesewing/plugin-flip": "^2.20.4",
"@freesewing/utils": "^2.20.4",
"react-scripts": "^5.0.0",
"webpack": "^5.67.0",
"rollup": "^2.66.1",
"@rollup/plugin-babel": "^5.3.0",
"@rollup/plugin-commonjs": "^21.0.1",
"@rollup/plugin-json": "^4.1.0",
"@rollup/plugin-node-resolve": "^13.1.3",
"rollup-plugin-peer-deps-external": "^2.2.4",
"@material-ui/core": "^4.12.3",
"@material-ui/icons": "^4.11.2",
"@material-ui/lab": "^v4.0.0-alpha.60",
"axios": "^0.25.0",
"react-intl": "^5.24.4",
"prop-types": "^15.8.1",
"mocha": "^9.1.1",
"chai": "^4.2.0",
"@babel/register": "^7.10.5"
```

I think that except for Rollup (bundler), mocha and chai (tests),all of these can go once we make the switch.
Then, once we've got that done and dusted, we can look at migrating to [Esbuild](https://esbuild.github.io/) (which is also on our roadmap). That should further cut down the build time and reduce the number of dependencies.

So perhaps we put a pin in this and try to optimize further once we get there?
If nothing else, the total install time might be a good indicator to see whether it was worth the hassle ;)

nicholasdower:
> Guess there's no magic solution to making the install faster. However, I do think we should see a speed-up once we migrate to the new development environment, as we'll essentially be able to rip out the old one.
> 
> Looking at a typical pattern `package.json` these are the dev dependencies:
> 
> ```
> "react": "^17.0.2",
> "react-dom": "^17.0.2",
> "@babel/plugin-proposal-class-properties": "^7.13.0",
> "@freesewing/components": "^2.20.4",
> "@freesewing/css-theme": "^2.20.4",
> "@freesewing/i18n": "^2.20.4",
> "@freesewing/mui-theme": "^2.20.4",
> "@freesewing/plugin-bust": "^2.20.4",
> "@freesewing/plugin-buttons": "^2.20.4",
> "@freesewing/plugin-flip": "^2.20.4",
> "@freesewing/utils": "^2.20.4",
> "react-scripts": "^5.0.0",
> "webpack": "^5.67.0",
> "rollup": "^2.66.1",
> "@rollup/plugin-babel": "^5.3.0",
> "@rollup/plugin-commonjs": "^21.0.1",
> "@rollup/plugin-json": "^4.1.0",
> "@rollup/plugin-node-resolve": "^13.1.3",
> "rollup-plugin-peer-deps-external": "^2.2.4",
> "@material-ui/core": "^4.12.3",
> "@material-ui/icons": "^4.11.2",
> "@material-ui/lab": "^v4.0.0-alpha.60",
> "axios": "^0.25.0",
> "react-intl": "^5.24.4",
> "prop-types": "^15.8.1",
> "mocha": "^9.1.1",
> "chai": "^4.2.0",
> "@babel/register": "^7.10.5"
> ```
> 
> I think that except for Rollup (bundler), mocha and chai (tests),all of these can go once we make the switch. Then, once we've got that done and dusted, we can look at migrating to [Esbuild](https://esbuild.github.io/) (which is also on our roadmap). That should further cut down the build time and reduce the number of dependencies.
> 
> So perhaps we put a pin in this and try to optimize further once we get there? If nothing else, the total install time might be a good indicator to see whether it was worth the hassle ;)

Not sure whether you meant put a pin in this change or the discussion for how to speed up the build.

I think at least parts of this change can be salvaged. Options I see:

A. Close this PR and revisit in the future.
B. Revert the workflow bit but keep the testall changes. I don't think continuing on failure and printing the errors is controversial but please let me know.
C. Keep the testall changes and the workflow bit but only have the workflow run on pushes to develop, not on PRs. Here I don't see a downside to a longer build since no one is waiting on it. This won't help prevent build breaks but would allow us to see what changes break the build.
D. Like C but also add a way to trigger the build on PRs by leaving a specific comment. (It at least gives a way to verify risky changes).
E. Merge as is. (PR builds will be longer but hopefully the build will be broken less frequently).

joostdecock:
> Not sure whether you meant put a pin in this change or the discussion for how to speed up the build.

I meant: this is a great contribution. Let's not hold it back just because PR builds/tests take longer than we'd perhaps like to. So let's put a pin in our efforts to speed up the build by tweaking build settings until we can address the root cause: reducing the number of dependencies.

I'm sorry for the misunderstanding, it must have been disappointing if it gave you the feeling I was dismissive of your work. That was not my intent. I'm a big fan of your contributions 🤗

Going forward, I'll try to be more clear about my intentions.

**Summary**: :shipit:

nicholasdower:
> > Not sure whether you meant put a pin in this change or the discussion for how to speed up the build.
> 
> I meant: this is a great contribution. Let's not hold it back just because PR builds/tests take longer than we'd perhaps like to. So let's put a pin in our efforts to speed up the build by tweaking build settings until we can address the root cause: reducing the number of dependencies.
> 
> I'm sorry for the misunderstanding, it must have been disappointing if it gave you the feeling I was dismissive of your work. That was not my intent. I'm a big fan of your contributions 🤗
> 
> Going forward, I'll try to be more clear about my intentions.
> 
> **Summary**: :shipit:

Thank you! And thanks for the thoughtful review. I love this process and wasn't at all disappointed. Now even less so knowing my change is going it! 😛

Issue #1513, opened 2021-11-23
Add frequently made mistakes to markdown guide


lucibytes:
Add @joostdecock's list of overlooked mistakes to the markdown guide. This won't be an extensive guide to documentation.

I will do this but also I am slow 🐌

darigovresearch:
Thanks @lucibytes for having a go at this, have added extra tags to encourage any people looking at the issue to have a go if they are interested in contributing and think they may be able to resolve this faster

lucibytes:
Awesome, thanks!

SeaZeeZee:
Closing this issue since the documentation got merged!

PR #1808, opened 2022-02-04
Markdown Documentation Frequent Mistakes

SeaZeeZee:
Added documentation on frequent mistakes in Markdown #1513

### DIFF
--- a/markdown/dev/guides/markdown/frequent-mistakes/en.md
+++ b/markdown/dev/guides/markdown/frequent-mistakes/en.md
@@ -6,13 +6,15 @@ Some things to keep in mind when working in Markdown are:
 
 <Note>
 Github itself also allows working in Markdown and will give you a handy preview!
-</Note>
+</Note>  
 
 If you get lost or have a question about how to do something, feel free to come ask on the Discord. We've all had to learn Markdown at some point and would be delighted to pass knowledge on. 
 
 - When adding links please do not link them using a structure like: Link [here](https://www.youtube.com/watch?v=dQw4w9WgXcQ). Instead put the link under the relevant term. See first list item for an example.
 
-- When you are linking within freesewing you can use a relative link from the site root. Think ```markdown/dev/guides/markdown/frequent-mistakes``` instead of ```https://github.com/freesewing/freesewing/tree/develop/markdown/dev/guides/markdown/frequent-mistakes```  
+- When you are linking within freesewing you can use a relative link from the site root.  
+Think ```markdown/dev/guides/markdown/frequent-mistakes``` instead of ```https://github.com/freesewing/freesewing/tree/develop/markdown/dev/guides/markdown/frequent-mistakes```  
+
 Images can be put in the same folder you are working on with a link to the filename. 
 
 - If you're writing documentation that involves steps, please do not mix levels of steps. Steps written out in documentation are there to facilitate brainless execution. Don't be afraid to repeat yourself.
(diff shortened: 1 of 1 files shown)


markdown/dev/guides/markdown/frequent-mistakes/en.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Frequent mistakes
+order: 100
+---
+
+Some things to keep in mind when working in Markdown are:
+
+- There is no need to add a *glossary* section to documentation. We use a plugin called [remark-jargon](https://github.com/freesewing/freesewing/blob/develop/packages/remark-jargon/README.md) to explain terms. Information can be found at the link.
+
```
nicholasdower, line 9: I think something is going wrong because of the blank lines between each bullet line (lines 9 and 11 for example.

This is what it looks like for me:
image

Removing the lines resolves the issue.

SeaZeeZee: Huh, it didn't do that in the preview! I'll fix it :)

nicholasdower: Not sure what is going on. Some other parts of the page are also being rendered strangely on my local freesewing.dev but they look fine in GitHub: https://github.com/freesewing/freesewing/blob/1c8afb2100e58c4cde71b652975b208c847ab955/markdown/dev/guides/markdown/frequent-mistakes/en.md

Will play around with this tomorrow.

joostdecock:
Going to merge this as this won't get published anyway and I have some other docs to write, so I'll give this a closer look while I'm at it.

SeaZeeZee:
Let us know if you figure out what is causing it?

joostdecock:
Ok, so @nicholasdower was right that it's the extra white space (line breaks) that cause this to happen.

But the more general problem is that you're using a bulleted list for what essentially is a subtitle.
Like this:

```md
- Do this not that because yadayadayada

Trust me, it's better this way 

- Also make sure to bla bla
```

What happens is that the markdown parser sees a line that starts with `-` (aha, a list) but also a line that stands on its own (aha, a paragraph) so it creates a list and each list item is its own `<p>` tag which is of course a block-level element in HTML with it will be rendered below the bullet of the list.

The *correct* approach here is to use subtitles for subtitles, which I've done in [5036723](https://github.com/freesewing/freesewing/commit/50367237f122ebd554ab95294e6aec96b097d390)

I'll be making a few more tweaks to show some other improvements :)

joostdecock:
Forgot to that this in this commit: https://github.com/freesewing/freesewing/commit/8ab7bf584080697a158bfc6719acb0f78628a2ae

joostdecock:
Ok, so I've made some changes to this, not all super important, but just to sort of underscore best practices:

- Changed to order (in frontmatter) to push this to the end of the markdown guide
- Changed from bullet list to subheadings (which we will provide anchor links for in the future and maybe a TOC)
- Replaced some very long links with a link reference as that makes it easier to read the raw markdown
- Limited line length to 80 characters as that makes it easier to read the raw markdown
- Changed the inline lengthy code examples to a code block,as inline doesn't wrap and will break the layout on mobile
- Added an example for adding an image
- Updated the link example to link to a freesewing website rather than github

Which results in:
Screenshot from 2022-02-05 10-18-19

SeaZeeZee:
Awesome! Things for me to keep in mind writing next time :)

PR #1809, opened 2022-02-05
feat(core): Add a default hide() method to options

joostdecock:
This adds a default `hide()` method to all pattern options
(that are not a constant) which always returns false.

The idea is that this can/will be used for conditional display
of options based on the user's input.

See Issue #1723 and PR #1798 for background info


nicholasdower (approved):
Nice. Love the test too. :shipit:

PR #1810, opened 2022-02-05
Updating freesewing.dev - Test for new build flow

joostdecock:
I noticed that some of the more recent content on freesewing.dev did not show up in the search. And sure enough, we're not updating the Algolia index when deploying.

I've added this to the build script, and this PR should give us a good test-run to see if it behaves as expected.


SeaZeeZee:
How would I go about reviewing this code?

joostdecock:
@SeaZeeZee the first thing is: Do all the checks pass. I've recently added some extra bots to check code quality, and they picked up some issues here (like no having rate limited our backend endpoint that generates PDFs).

So first I want to get those to run properly.

Then, does the Netlify build pass? If it does, check out the preview, does it look ok?
It it does: :shipit:  (this `shipit` emoji is shorthand for **Shit it!** (shipping being a term used for releasing software).

SeaZeeZee:
I think I follow!

joostdecock:
@SeaZeeZee So all looks good now. Netlify will post the deploy preview. You need to click the `details` links at the **Deploy preview ready!** line.

You can see it's the latest content because it includes the docs I just wrote:
https://deploy-preview-1810--fs-dev-en.netlify.app/reference/api/config/options/pct/snap

Also note the `#` signs in front of the titles with anchor links, that's new too.
Also new is the smartypants plugin which allows you to write `--` for an mdash. 

That being said, there's still two issues:

- The [deploy log](https://app.netlify.com/teams/freesewing/builds/61febef583ea5b000781e4cd) has an error for the indexing to Algolia (this is the service that powers the website search). It doesn't break the build but the error is `Index not allowed with this API key` so it's just a permission problem
- Related: The Algolia index should only be updated when deploying in production. Not on deploy previews. So that's something I should figure out too.

```

Issue #1360, opened 2021-10-02
Add armhole to waist measurement


joostdecock:
This would help with (womenswear) blocks. Question is, where does this measurement start on the armhole side?

This is a todo from #1353

bobgeorgethe3rd:
Would you do it from where the chest circumference is at the armpit?

SeaZeeZee:
I dont have a drafting textbook here, but I wonder where that says to measure from, there must be a standard for it!

stsydow:
I tried something like that in #1817 for Brian.
I previously used the paperless mode and where counting gird units quite a lot.
So I added all measurements I would use:
![pattern](https://user-images.githubusercontent.com/961256/152705739-88245aed-81cb-4bea-bde6-3d01e3416773.svg)

SeaZeeZee:
I did some research on archive.org because I don't have a modern book on pattern drafting available. Bearing in mind these are older sources armhole to waist is usually measured from the armpit to the waist down with the arm down at the side. Sometimes with the aid of a square. Notes are to not have people clamp so hard they raise the shoulder.

Image is from a drafting manual from 1888. 

![armhole](https://user-images.githubusercontent.com/86711383/152819631-90577317-fe90-4038-93f2-4b286d06f90b.png)

stsydow:
Oh I get it (sorry, I got the original question wrong). So `hpsToWaist - armpitToWaist` would be the minimal armhole depth?
Than we could apply the biceps ease and don't use the `armholeDepthFactor`?

joostdecock:
FYI: Work for this started in #2060

stsydow:
I appreciate that you introduced `armholeHeight` to brain.

I fear that it is quite indirect and therefore error-prone.
It depends on the measurements high-shoulder-to-waist, shoulder-slope, and armhole-to-waist.
I fear that the accumulated error for armholeHeight will be a problem for tight fitting cloth.

But it is a step forward in any case!

joostdecock:
@stsydow If you have a more resilient way to determine hps, shoulder, and armhole I'd love to hear it.

But I feel that this change will make it somehow more unforegiving, and perhaps that makes it easier to spot issues with faulty measurments.

What I mean is, let's say that you measure your shoulder slope as being too steep. Currently, that will push down the entire armhole, and so you will need a bit of a trained eye to realize the underlying problem is the shoulder slope. Most people will just see a bunch of wrinkles and conclude *this fit sucks*.

[...]

PR #1817, opened 2022-02-06
Draft: Extend brain for people with breasts and waist.

stsydow:
Extend brain for people with breasts and fit to waist.
Intended as a base block for stretch shirts and turtlenecks too.

This is a pretty extensive rework and a request for feedback.
It which would break other patterns if merged now.

Should this become a pattern of its own?

Also reworked the paperless mode.


joostdecock:
Keeping an eye on this one :eyes: very happy to see some work on this

> Should this become a pattern of its own?

I feel that a block for stretch should probably become its own pattern. 
If for no other reason that it's more intuitive that way for people who are looking for a stretch block.

Perhaps a good middle ground would be to:

- Back-port those improvements that are meaningful for a woven block to Brian 
- Make the stretch block extend Brian so you can limit it to what matters

But I'm just thinking out loud here :)

joostdecock:
@stsydow Are you keeping this in draft because you're still working on it, or to prevent it from accidentally getting merged (since it would indeed break a bunch of stuff).

I'm asking because I'm rather busy lately so I'll probably put this on the long finger for a closer look until I know you're ready with it :)

stsydow:
All off the above :) This is a mixed bag because I want to make a pullover for my wife (the third one?) and this time I did not want to mess the paper draft to make it fit.

The following things are on my mind still:
 - [ ] a stretch fabric option 0% - 20% 
 - [ ] a pullover option with guaranties on  a wide enough waist and hip to pull it over in one piece. (waist without ease + stretch > chest/bust/(shoulder?)  
 - [ ] see what happens to other patterns based on Brian. 
 - [ ] `highBust` can become optional
 - [ ] fix docs
 - [ ] make a pullover and test out the realthing
 - [ ] seam allowance on the hem is broken

So I'm happy for anyone giving it a try, a rough review for things that sick out or even looking the example in #1360 to tell if the paperless mode looks good or needs improvement, is appreciated.

joostdecock:
Hi @stsydow I've been trying to catch-up with some of the PRs that have gone somewhat stale, and this is one of them.

What do you propose to do with this effort?

stsydow:
Sorry for the delay. Life has gotten in the way.
I will work on in but probably not in the next couple of weeks. 
So I'm fine when you want to close it.
If someone want's to reuse parts for other purposes that's also fine.
I could also explain things via phone if someone is interested, but I will not make it to  write up things 
in the meantime.

joostdecock:
Since this is still v2 code, I'm closing this now as we're working towards a v3 release.
Thanks for your efforts @stsydow perhaps we can revisit this at some point in the future.

PR #1824, opened 2022-02-07
Moved cutting list to top

SeaZeeZee:
Moved the cutting list to the top of the page, as that reads easier than having it interspersed in the text.


bobgeorgethe3rd:
Query on this? Doesn't it make more sense to keep this at the top as it makes the user question their cutting before the cutting list?

PR #1837, opened 2022-02-09
Run backend tests in CI.

nicholasdower:
The idea is to make it easier to run the backend tests so that they can be run in CI. Right now running the backend tests requires creating a `.env` file, starting the DB, seeding it with data, starting the server, and then running the test script.

The approach here is to introduce a `testci` script which makes it possible to run the tests in one command without any setup. The `testci` script:
 - Loads default environment variables
 - Starts an in-memory database
 - Seeds the database with the required test data
 - Starts a development server in the same process
 - Runs the tests in a new process

I originally planned to run the tests in the same process but ran into issues with import vs require. I assume there is a way around this but didn't think running the tests in a new process was a bad idea anyway.


packages/backend/src/index.js
```diff
@@ -1,57 +1,8 @@
-import './env'
```
nicholasdower, line 1: The new file is simply:
```javascript
import { connectToDb, startApp } from './app'

connectToDb();

const app = startApp();

export default app;
```

packages/backend/src/app.js
```diff
@@ -0,0 +1,64 @@
+import './env'
```
nicholasdower, line 1: This diff is misleading. I moved the old index.js to app.js and moved the code into two functions. The real diff is something like:
```diff
+export const startApp = () => {
...
-  if (__DEV__) console.log(chalk.yellow('> in development'))
+  if (process.env.NODE_ENV === 'development') console.log(chalk.yellow('> in development'))
...
+  return app;
+};

+export const connectToDb = () => {
...
+};
```

packages/backend/package.json
```diff
@@ -19,6 +19,7 @@
     "major": "npm version major -m ':bookmark: v%s' && npm run build",
```
nicholasdower, line 19: Hope I'm not wrong but I don't think this file is generated.

packages/backend/tests/env.js
```diff
@@ -0,0 +1,20 @@
+process.env.FS_BACKEND = "http://localhost:3000";
```
nicholasdower, line 1: This file is just a copy of the example `.env` file. I don't know how this is normally done but I want to the tests to run the same regardless of the environment.

nicholasdower:
Note that the tests are currently passing locally but failing in CI. I just learned that when using --no-bail lerna does not emit stderr. I've just added a commit to run the tests serially and output stderr so that I can debug. Depending on how much slower this is, I might suggest we leave this in. See https://github.com/lerna/lerna/issues/2199.

nicholasdower:
> Note that the tests are currently passing locally but failing in CI. I just learned that when using --no-bail lerna does not emit stderr. I've just added a commit to run the tests serially and output stderr so that I can debug. Depending on how much slower this is, I might suggest we leave this in. See [lerna/lerna#2199](https://github.com/lerna/lerna/issues/2199).

Fixed this by updating the i18n dependency in backend as it was pinned to a broken version.

joostdecock:
This is great work and a wonderful contribution. I'd like to see this merged :ship:

Issue #1843, opened 2022-02-10
SVG styling is broken on Safari


joostdecock:
**What seems to be the problem?**

This:
![IMG_7133](https://user-images.githubusercontent.com/1708494/153365011-3a493097-a194-4202-a93f-93a7544e4fa1.jpeg)

Only seems to happen on Safari.

@nicholasdower points out that https://simon.freesewing.dev/ is not broken so this is probably related to the recent CSS changes for the `scale` factor, making this a regression.

nicholasdower:
@joostdecock do you mean specifically on the drawing-circles page you linked to? That page is not broken on Safari but it isn't using css vars.

Are you thinking this isn't simply the units thing?

What do you mean by "old site" and "new site"?

Also, what is react render? I only ever see SVGs in the DOM.

Sorry if I am being dense. I am a newb.

nicholasdower:
I am still piecing this together but is it fair to say that the "new site" or freesewing.dev involves code like:
```xml
<Example pattern="tutorial" part="step6">
```

or

```xml
<Example pattern="rendertest" options_only="snippets">
```

?

joostdecock:
I don't always know what you don't know, so forgive me if this is repetitive:

```jsx
<Example pattern="tutorial" part="step6">
```

That's [JSX](https://reactjs.org/docs/introducing-jsx.html) syntax, which is a [React](https://reactjs.org/) thing.

So React is all about **components**. A component is (under the hood) just a method that takes parameters and should return something React can render (a string, or array or strings or another method/JSX).

JSX is a way to write out your UI in a more declarative way. So this:

```
<Example pattern="tutorial" part="step6">
```

Will under the hood be:
```js
function Example(props) {
  /* props looks like:
   * {
   *    pattern: "tutorial",
   *    part: 'step6",
   * }
   */
   
   // Do something smart here
   
   return "Hello"
 }
 ```
  
Those components allow us to split things up in smaller re-usable parts.
Whereas core spits out SVG as one string, React allows us to have a component just to render a path (for example).
And in that component, we can easily do things like add extra features when the path is clicked on and so on.

Since React is Javascript, all this works only in the browser. For export, we need the plain simple SVG string that's spit out by core.

So both the new site (freesewing.dev) and the (current) old site (freesewing.org) use React to render a pattern.
Ultimately, in the DOM you will see SVG. But it's SVG that's rendered by Javascript, and not by our core library.

Here's what our core library does to render an SVG path:
[...]

PR #1847, opened 2022-02-10
#1843 Use units in css scale var.

nicholasdower:
Attempting to address https://github.com/freesewing/freesewing/issues/1843.

See https://stackoverflow.com/questions/65849051/calc-not-working-with-stroke-dashoffset-in-safari-and-firefox

This might not get us all the way but it at least resulting in paths being rendered in my local Simon.


joostdecock (approved):
:shipit:

PR #1850, opened 2022-02-12
fix(docs) Holmes missing context

bobgeorgethe3rd:
literally step above talks about the faced visor.
I also re-added the capitals. Ik it is not grammatically correct but the point is to make people double check what they are reading but also realise it is a sewing term. Now I am thinking I need to check the glossary and make sure some of the ones not showing in remark are on there. Mm.

### DIFF
--- a/markdown/org/docs/patterns/holmes/instructions/en.md
+++ b/markdown/org/docs/patterns/holmes/instructions/en.md
@@ -98,7 +98,7 @@ There are two methods for constructing the visors. This is due to the different 
 #### The Visors Method 1
 
 - Place visor insert on one visor piece inside seam lines.  
-- Temporarily secure visor insert to visor with temporary _pad Stitches_. This will be refered to as "faced visor" going forward.  
+- Temporarily secure visor insert to visor with temporary _Pad Stitches_. This will be refered to as "faced visor" going forward.  
 - _Baste_ along the seam line of the inner curve of the faced visor, making sure not to catch the visor insert.  
 - With right sides together, matching seam lines and centre fronts, sew the faced visor to another visor piece along the outer curve close to the visor insert, making sure not to catch the visor Insert.  
 - Notch and trim the outer curve making sure not to clip the stitching. (You may wish to turn and check the shape before this step)  
--- a/markdown/org/docs/patterns/holmes/instructions/en.md
+++ b/markdown/org/docs/patterns/holmes/instructions/en.md
@@ -49,7 +49,7 @@ Depending on your seam allowance you may want to trim seams as you go to reduce 
 - With right sides together, matching seam lines, sew a tie ear flap to a non-tie ear flap.  
 - Clip and trim seam allowance.  
 - Turn inside out and press.  
-- (Optional) _Topstitch_ or _edgestitch_ the folded edge.  
+- (Optional) _Topstitch_ or _Edgestitch_ the folded edge.  
 - _Baste_ the raw edges together.  
 - Repeat for the remaining ear flap.
 
@@ -87,7 +87,7 @@ The ear flaps will need to be long enough to go over the top of the peak so they
 
 **For both Ear Flaps with Ties and Button Holes**  
 
-It is recommended to at least _topstitch_ the ear flaps but this may not be desirable with certain patterns so has been marked as optional.
+It is recommended to at least _Topstitch_ the ear flaps but this may not be desirable with certain patterns so has been marked as optional.
 
 </Note>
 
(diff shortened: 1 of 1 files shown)


nicholasdower:
The capitalization feels awkward to me for a few reasons:
 - The terms aren't proper nouns.
 - Other people writing and editing these files are unlikely to use this capitalization style, leading to inconsistency. For instance we use "baste" without capitalization around 50 times in instructions but only capitalize it 4 times.
 - The terms are already emphasized by being italicized and highlighted in blue by the Remark plugin.

If you do think we should use this capitalization style, is it reasonable to document it as a guideline and set a goal to migrate existing documentation to the new style?

joostdecock:
We can style jargon however we want, including forcing capitalization through CSS.

I'm not saying I'm in favor of capitalization (I am in favor of writing what is grammatically correct in the context). But if we want to enforce a given style on terms/jargon, CSS is where to do that, not the content itself.

Also keep in mind that you don't really have a lot of control to how this ends up getting translated.

Long story short, I think the best approach is to make sure words are properly defined/emphasized. Not only in the jargon terms file itself, but also as documentation under `sewing` so they can be linked to.

bobgeorgethe3rd:
Initally they were Bold, Italics and Captilized as I was connecting them to the glossary that was not needed. (You can see what this looks like in Yuri). I was going to make them full capitals at one point when initally writing but that didn't feel like it fit in with the rest. If we do decide to change it I will do around and update/re-format the other docs. One of my things in the ever growing to-do list is to update/re-format any of the markdown into the standard though there is a bit of confusion on what that standard would be as I have definitely strayed from Aaron with all docs that I have added and only recently realised that I had done so. I also think Aaron needs a bit re-formatting as well but that might just be me. If we figure out then I do want to create templates as you need something that will have info for all variations.

joostdecock:
I feel the smart approach here is to use/write some remark-lint rules and integrate them with the automated tests: https://github.com/remarkjs/remark-lint

Perhaps @nicholasdower or @stsydow would like to help setting that up.
If not, I'll get around to it after merging #1834

stsydow:
You can add the following and do `yarn checkdocs` but its slow(2min + a ton of warnings) for the whole markdown tree.
 
```diff --git a/package.json b/package.json
index 9c3514536c..f988fdb23d 100644
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate",
     "famadd": "all-contributors add",
-    "famgen": "all-contributors generate"
+    "famgen": "all-contributors generate",
+    "checkdocs": "remark ./markdown --quiet --frail"
   },
   "repository": {
     "type": "git",
@@ -104,7 +105,12 @@
     "react-markdown": "^6.0.2",
     "react-scripts": "^5.0.0",
     "remark": "^13.0.0",
+    "remark-cli": "^10.0.1",
     "remark-html": "^13.0.1",
+    "remark-lint-emphasis-marker": "^3.1.1",
+    "remark-lint-list-item-indent": "^3.1.1",
+    "remark-preset-lint-consistent": "^5.1.1",
+    "remark-preset-lint-recommended": "^6.1.2",
     "rimraf": "^3.0.2",
     "rollup": "^2.48.0",
     "rollup-plugin-peer-deps-external": "^2.2.3",
@@ -130,5 +136,15 @@
     "jsonfile": "^6.1.0",
     "postcss": "^8.4.5",
     "tailwindcss": "^3.0.7"
+  },
+  "remarkConfig": {
+    "plugins": [
+      "remark-preset-lint-consistent",
+      "remark-preset-lint-recommended",
+      "remark-lint-emphasis-marker",
+      ["remark-lint-list-item-indent", "space"],
+      ["remark-lint-emphasis-marker", "_"],
+      ["remark-lint-strong-marker", "*"]
+    ]
   }
 }
```
To have it make corrections you could use:
```
$ npx remark ... -o
```
I'm bit busy the next days but I hope it helps

joostdecock:
Thanks@stsydow this is a great starting point 🙏

nicholasdower:
I definitely think we should do this. I filed #1867 so we don't forget.

nicholasdower:
@bobgeorgethe3rd I suggest we revert the capitalization changes from this PR and merge the rest. If we want to capitalize terms we can do that via the remark style as @joostdecock suggested. wdyt?

joostdecock:
Leaving out the capitalization it seems the only change is adding the word `faced` so I've added that in ab44023
Closing

Issue #1868, opened 2022-02-14
Scale setting is not respected on the website


joostdecock:
**What seems to be the problem?**

![Screenshot from 2022-02-14 18-02-31](https://user-images.githubusercontent.com/1708494/153911145-874e6bec-4092-4647-9bb8-f16e416207e0.png)

**Additional context**

Not sure whether this is a regression or we somehow missed that this never looked good, but it's not great.

nicholasdower:
@joostdecock what are the steps to reproduce? The screenshot looks like what I see if I increase the scale setting while configuring a pattern. But I think that is expected so I'm assuming this is something else.

joostdecock:
Must admit, I didn't think to check the scale. It's a url shared by a user: https://freesewing.org/patterns/jkscx/

joostdecock:
Exported gist here: https://gist.github.com/c1f8047f349a179708af77247ed262e5

I don't see scale set, but I am on my phone so take it with a grain of salt

nicholasdower:
I can easily reproduce by opening any of my own patterns. It is only happening when viewing a pattern on the website, not when editing.

nicholasdower:
The scale variable is set to an empty string. This resolves the issue locally:
```javascript
document.documentElement.style.setProperty('--freesewing-pattern-scale', '1');
```

nicholasdower:
I guess it isn't set to an empty string. It just isn't set. I don't see it in the DOM.

nicholasdower:
Ah, maybe this belongs in freesewing.org! I see https://github.com/freesewing/freesewing.org/commit/a2162034cc0098f8a1271d436ca7110a399189db added it to draft. But I don't see it in preview. 

https://github.com/freesewing/freesewing.org/pull/1585

nicholasdower:
Before:
![image](https://user-images.githubusercontent.com/9117775/153923139-a809f244-3611-4293-8c87-6b5ae0fa7c0e.png)
After:
![image](https://user-images.githubusercontent.com/9117775/153923209-a0b3bb6e-123c-4c57-8dc4-0b0112f49e49.png)

Btw, this is not working on Safari and adding `px` units does not help. See #1843.

nicholasdower:
Ah, I think it is broken in Safari because we use units in the math like:
```css
font-size: calc(22px * var(--freesewing-pattern-scale));
```

joostdecock:
I'm AFK but this is some quality digging ⛏️

[...]

PR #1869, opened 2022-02-14
Remove units from calc(var(...

nicholasdower:
Units must be included in the css var in order for this to render properly on Safari. But including units in the rest of the calculation breaks Safari too. This seems to work.

Corresponding change in https://github.com/freesewing/freesewing.org/pull/1585

See https://jsfiddle.net/tw2se8ub/

See #1868 and #1843.


stsydow (approved):

joostdecock:
Thanks @nicholasdower

PR #1877, opened 2022-02-15
fix(yml) Warlaee Pattern Options

bobgeorgethe3rd:
I'm not sure when the options were updated but the yml hadn't been. This should bring it upto date.


joostdecock:
Thanks @bobgeorgethe3rd :+1:

PR #1879, opened 2022-02-15
Chore: Updating freesewing.org

joostdecock:
After @nicholasdower 's fix for #1868 in #1869 we should update the website.

There's been a bunch of (mostly docs) changes too, so let's get a deploy preview going and check that everything looks fine.


nicholasdower:
@joostdecock I don't know all the required steps but I think we also need https://github.com/freesewing/freesewing.org/pull/1585.

joostdecock:
Wow, this PR is a nice testament to the kind of week I'm having :see_no_evil: 
What a brainfart. Getting freesewing.org updated is not done in this repo. Closing.

Issue #1867, opened 2022-02-14
Add lint checks for Markdown


nicholasdower:
From @joostdecock on #1834:
>I feel the smart approach here is to use/write some remark-lint rules and integrate them with the automated tests: https://github.com/remarkjs/remark-lint
>
>Perhaps @nicholasdower or @stsydow would like to help setting that up.
If not, I'll get around to it after merging https://github.com/freesewing/freesewing/pull/1834

And from @stsydow:
>You can add the following and do `yarn checkdocs` but its slow(2min + a ton of warnings) for the whole markdown tree.
 
```diff --git a/package.json b/package.json
index 9c3514536c..f988fdb23d 100644
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate",
     "famadd": "all-contributors add",
-    "famgen": "all-contributors generate"
+    "famgen": "all-contributors generate",
+    "checkdocs": "remark ./markdown --quiet --frail"
   },
   "repository": {
     "type": "git",
@@ -104,7 +105,12 @@
     "react-markdown": "^6.0.2",
     "react-scripts": "^5.0.0",
     "remark": "^13.0.0",
+    "remark-cli": "^10.0.1",
     "remark-html": "^13.0.1",
+    "remark-lint-emphasis-marker": "^3.1.1",
+    "remark-lint-list-item-indent": "^3.1.1",
+    "remark-preset-lint-consistent": "^5.1.1",
+    "remark-preset-lint-recommended": "^6.1.2",
     "rimraf": "^3.0.2",
     "rollup": "^2.48.0",
     "rollup-plugin-peer-deps-external": "^2.2.3",
@@ -130,5 +136,15 @@
     "jsonfile": "^6.1.0",
     "postcss": "^8.4.5",
     "tailwindcss": "^3.0.7"
+  },
+  "remarkConfig": {
+    "plugins": [
+      "remark-preset-lint-consistent",
+      "remark-preset-lint-recommended",
+      "remark-lint-emphasis-marker",
+      ["remark-lint-list-item-indent", "space"],
+      ["remark-lint-emphasis-marker", "_"],
+      ["remark-lint-strong-marker", "*"]
+    ]
   }
 }

```
>To have it make corrections you could use:
```
$ npx remark ... -o
```

joostdecock:
This has been implemented, with #1900 and #1901 waiting for review

PR #1898, opened 2022-02-18
Initial test of markdown linting (and auto-fixing)

joostdecock:
So after the initial setup of markdown linting (see #1867) I did some tests, and we need the `remark-frontmatter` plugin since frontmatter is not part of standard markdown, so without this plugin, it will causes all sort of false positives.

After that, I did a small test (small because I limited it to markdown under `markdown/dev/guides/best-practices`) to auto-fix the markdown.
That commit is in this PR (along with the installed plugin) and I think it looks good.
But before running this over more markdown content, I'd like a second opinion :)

Thanks to @stsydow for the initial legwork for setting this up and @nicholasdower for creating an issue for it.

### DIFF
--- a/markdown/.remarkrc.yaml
+++ b/markdown/.remarkrc.yaml
@@ -1,6 +1,7 @@
 settings:
   bullet: "-"
 plugins:
+  - remark-frontmatter
   - remark-preset-lint-consistent
   - remark-preset-lint-recommended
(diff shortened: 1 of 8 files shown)


nicholasdower (approved):
Love this. I'm excited to see the massive change when you run this on all files. :shipit:

PR #1899, opened 2022-02-18
Add smoke test for create-freesewing-pattern.

nicholasdower:
#1891 breaks the `create-freesewing-pattern` package but All Tests is green. Adding a test which simply runs `node ./lib/cli.js -h` and asserts that the command runs successfully.


joostdecock:
Good stuff

PR #1904, opened 2022-02-19
chore: Changes to test script settings

joostdecock:
I've added a custom mocha reporter that only produces output when tests fail.
I've updated the script config to use this reporter for `testci` runs, so that CI output is more terse.

This also includes a port of the `i18n` tests to ESM so that we can use the same run-script there.

I'm making a mental note that the core tests should also be ported to ESM.


joostdecock:
I can't seem to bend Github actions to me will :shrug: 

This works:
```
BABEL_ENV=production npx mocha tests/*.test.mjs --require @babel/register --jobs 4 --parallel --reporter ../../tests/reporters/terse.js
```

This does not:

```
BABEL_ENV=production npx mocha tests/*.test.mjs --require @babel/register --jobs 4 --parallel --reporter ../../tests/reporters/terse.js
```

So it's en environment thing. It's seems to be mostly patterns vs pluings. Either an environment thing, dependency config, or the way the tests themselves are configured.

nicholasdower:
Attempted to address this in https://github.com/freesewing/freesewing/pull/1906.

nicholasdower:
Pushed the changes to this branch. Will merge on green.

nicholasdower (approved):

PR #1906, opened 2022-02-19
WIP - DO NOT MERGE

nicholasdower:
Attempting to address test failures in #1904.

All tests was broken because `--parallel` is not compatible with `--file`. My suggestion is to remove --parallel since we already have concurrency in the all tests run (packages are run in parallel).

I don't know why the other workflows are broken. They just run `npm run testci` in each package and that seems to run fine for me locally. But, I'm not sure it is worth figuring that out. We planned to delete these workflows eventually in favor of all tests. Maybe we just do that now.

This PR isn't meant to be merged. Any useful bits should be pushed to #1904.


nicholasdower:
Pushed changes to #1904. Closing.

PR #1913, opened 2022-02-19
Create locale dir automatically.

nicholasdower:
I am unable to run freesewing.dev unless I create the locales dir manually. I think this can be done automatically.


joostdecock:
Agree that things should*just work*.
That being said, I went with the low-tech approach for the lab equivalent: https://github.com/freesewing/freesewing/tree/develop/packages%2Ffreesewing.lab%2Fpublic

I think I slightly prefer that since it's a one time thing.

We could even drop a README.md there explaining what those folders do and how they will be populated.

Perhaps there are benefits to doing it programmatically that I'm overlooking?

nicholasdower:
I hadn't noticed this approach. I have updated this PR to do the same. 

I guess the only advantage is that someone has to remember to do this if they add a new dir. After all, freesewing.dev was broken. That said, this is a rare event. I like your approach.

joostdecock (approved):

PR #1919, opened 2022-02-20
DO NOT MERGE - Testing Remark

nicholasdower:


joostdecock:
Hey, hear me out: What if those 10.000 warnings are actually the correct result?
Like, what if it is actually picking up the config file. I think it actually does.
Because there's tons of warning like this: ``Emphasis should use `_` as a marker`` and well, that's what's configured in the config file. And it's different from the default.

So the previous run changes all emphasis to use `*` and now it's being changed back to `_`.
So the run with the 10000+ warnings is actually the one we want.

nicholasdower:
> Hey, hear me out: What if those 10.000 warnings are actually the correct result? Like, what if it is actually picking up the config file. I think it actually does. Because there's tons of warning like this: `` Emphasis should use `_` as a marker `` and well, that's what's configured in the config file. And it's different from the default.
> 
> So the previous run changes all emphasis to use `*` and now it's being changed back to `_`. So the run with the 10000+ warnings is actually the one we want.

I think you're right!

I am still confused though. When you sent your dev and org changes, I assumed you ran fixdocs to generate them and there should therefore be zero warnings left. Instead, there are a zillion warnings left and `fixdocs` doesn't fix them.

Many of the warnings are like:
```
warning  Incorrect list-item indent: remove 2 spaces  list-item-indent  remark-lint
```

It is complaining about bullets having 2 spaces after them. But those are being added by `fixdocs`. If you remove the spaces manually and run checkdocs, it passes. But if you then run fixdocs, then run checkdocs again, it fails.

nicholasdower:
Reverted the workflow in https://github.com/freesewing/freesewing/pull/1921.

PR #1920, opened 2022-02-20
fix(workbench): Type-check optionalMeasurements. Closes #1918

### LINKED ISSUE #1918: optionalMeasurements key not type-checked before accessing its properties
**What seems to be the problem?**

On Discord, **redpandasnugs** writes: 

> I fired up a new development environment tonight and on version 2.20.6 it's throwing a props.optional is undefined error if I add a measurement to the index file. Looking at another pattern on 2.19.8 I can't reproduce the same behavior.
>
> Adding the optionalMeasurements key to the index file clears the error, regardless of whether I set a value for the key or leave it empty.
>
> If that behavior isn't expected I'm not sure where the error is. If it is maybe it would be worth adding optional Measurements as an empty key to the default starting index file?

joostdecock:
This fixes a regression bug introduced in 283fa4f884dce01b007f11a84718da870ed6feb8 and in doing so will close #1918.

The issue is that we just copied the code block from the required measurements.
But those are always set, whereas `optionalMeasurements` is not.

We could do the whole of making sure they are always set, but I went for the simple solution by just adding a check here.
My reasons for doing so is that once the new website is up, the entire `components` package will get deprecated and dropped :fire:

### DIFF
--- a/packages/components/src/Workbench/Measurements/index.js
+++ b/packages/components/src/Workbench/Measurements/index.js
@@ -148,7 +148,7 @@ const Measurements = (props) => {
             updateValue={props.updateMeasurement}
           />
         ))}
-        {props.optional.map((m) => (
+        {props.optional && props.optional.map((m) => (
           <FormFieldMeasurement
             key={m}
             name={m}


nicholasdower (approved):

joostdecock:
Tagging @stsydow because he did the initial work. Just as an FYI, I didn't realize this was about to cause problems when I committed it either.

stsydow:
Thanks and sorry! I hit it yesterday night too, but you where faster.

joostdecock:
Don't worry about it, I didn't realize it either. These things happen to people who do the work :)

I've released [2.20.7](https://github.com/freesewing/freesewing/releases/tag/2.20.7) just now with this fix in it, so it's all forgotten.

PR #1922, opened 2022-02-20
Checkdocs workflow take 2

nicholasdower:
Add remark settings so that fixdocs doesn't break itself.
Add workflow.
Run fixdocs.


nicholasdower:
Closing in favor of https://github.com/freesewing/freesewing/pull/1923.

PR #1923, opened 2022-02-20
chore: Tweaks to remark settings and linter run

joostdecock:
This holds two commits:

- b6dab9f8fa9d50307dc15fd70607c4f61be416c4 makes some tweaks to the remark config, specifically:
  - Using `./markdown` rather than `markdown` seems to work better for some reason
  - Ignoring non-English files as they are auto-generated from the English content, so updating them should happen via Crowdin (this should also speed things up)
- e6f1189017651803b2023742834952b0b2f1393d holds the result of the actual linting

This works, but for one thing: 
```
warning  Incorrect list-item indent: remove 2 spaces  list-item-indent  remark-lint
```
There are **2144** warnings, but the **all** seem to be this warning.
So it seems to me that there's an issue with [remark-lint-list-item-indent](https://www.npmjs.com/package/remark-lint-list-item-indent) where it's not fixing the files.

I will open an issue upstream. In the meanwhile, I think we can merge this?

### DIFF
--- a/markdown/.remarkignore
+++ b/markdown/.remarkignore
@@ -1,0 +1,5 @@
+# Don't bother with auto-generated files
+de.md
+nl.md
+es.md
+fr.md
--- a/markdown/.remarkrc.yaml
+++ b/markdown/.remarkrc.yaml
@@ -1,5 +1,7 @@
 settings:
   bullet: "-"
+  emphasis: "_"
+  strong: "*"
 plugins:
   - remark-frontmatter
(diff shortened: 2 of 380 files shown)


nicholasdower:
@joostdecock I think you are missing `listItemIndent: one`. Also in https://github.com/freesewing/freesewing/pull/1922/commits/9b83eac7d0cde60c61ee7a8843d9eefd76a55c88.

Issue #1936, opened 2022-02-22
Teagan's sleeve title says 'back' instead of 'sleeve'


joostdecock:
**What seems to be the problem?**

On Discord, **kalniete** writes: 

> Today I printed a Teagan, and the sleeve piece is labeled "back." The front and back are correctly labeled. Is the sleeve as printed by default the right side of the body? It looks like it is, but I don't see match marks to confirm.

PR #1942, opened 2022-02-23
Update Teagan left/right sleeve instructions.

nicholasdower:
https://github.com/freesewing/freesewing/issues/1936


joostdecock:
🙏

PR #1963, opened 2022-02-27
brian - dependency updates + minor fix

stsydow:


packages/brian/example/package.json
```diff
@@ -15,21 +15,21 @@
     "@freesewing/mui-theme": "latest",
     "@freesewing/pattern-info": "latest",
     "@freesewing/plugin-bundle": "latest",
-    "@freesewing/plugin-theme": "latest",
     "@freesewing/plugin-i18n": "latest",
     "@freesewing/plugin-svgattr": "latest",
+    "@freesewing/plugin-theme": "latest",
```
nicholasdower, line 20: I'm sorry to say that this file is generated via `yarn reconfigure`. So any manual changes will eventually be overwritten. To make changes to dependencies you have to update [config/dependencies.yaml](https://github.com/freesewing/freesewing/blob/develop/config/dependencies.yaml), then run `yarn reconfigure`.

Side note: I think we should add a PR build step that checks whether yarn reconfigure changes anything. It would also be nice to have a comment in this file but I'm not sure that is possible.

packages/brian/package.json
```diff
@@ -43,11 +43,12 @@
     "@freesewing/plugin-bundle": "^2.20.7",
     "@freesewing/plugin-mirror": "^2.20.7"
   },
-  "dependencies": {},
+  "dependencies": {
```
nicholasdower, line 46: Same comment about this file.

packages/brian/src/front.js
```diff
@@ -85,8 +85,8 @@ export default (part) => {
     points.s3ArmholeSplit = utils.curveIntersectsY(
       points.shoulder,
       points.mirroredShoulderCp1,
-      points.mirroredFrontArmholePitchCp2,
-      points.mirroredFrontArmholePitch,
+      points.mirroredArmholePitchCp2,
```
nicholasdower, line 88: I think I understand this change but want to confirm. `armholePitch` and `armholePitchCp2` are identical to `frontArmholePitch` and `frontArmholePitchCp2` since they are copied at the top of this file. So this change is about being consistent and shouldn't impact the pattern. Is that right?

stsydow: There is an option to offset the `frontArmholePitch` from `armholePitch` look like it got bolted on later.

nicholasdower: So sorry but I don't understand. Maybe you could just briefly explain the issue this change addresses?

stsydow: Ah sorry - I was also on the the `back` part in my head not the front But still:

`mirroredArmholePitch*` means actually `backMirroredArmholePitch*` which is correct as we extend the front here into the back part. (and not back into the front as it was)

stsydow: It looked correct when testing, but you got be thinking if I missed something.

joostdecock: The `base` part defines both `armholePitch` and `frontArmholePitch` points (and their control points). They are not the same, as the front is cut deeper, depending on the `frontArmholeDeeper` option:

https://github.com/freesewing/freesewing/blob/d0d16af60c32defecaa9c64694dc1e557b650c2e/packages/brian/src/base.js#L128-L132

The `front` part overwrites the `armholePitch` points with the values of the `frontArmholePitch`. That is so that people extending the pattern can just use the same `armholePitch` points without having to worry about back or front parts.

This change does not change anything. Because those points are the same.
However, I do think it's better to be explicit about the intent. This is the front part. We're working with the front armhole pitch point here. So as such, I'm not in favor of this change.

What I mean is, this change would mean that just looking at the line, there's a lot of room for misinterpretation. You have to be aware of the line at the top of the file to know that, actually, it's front pitch point anyway.

So while I think it's good to make sure there's a good reason for using these other names. I think in this case there is.

I would be open to have the `base` part define both `frontArmholePitch` and `backArmholePitch` points, and then have a similar mechanism in the `back` part where it copies the relevant points. I guess that would be the most explicit way to go and would require less cognitive overhead for people reading the code.

joostdecock:
Hey team, where are we with this? Is @stsydow going to take a stab at those armholePitch suggestions, or is this as far as you'll carry this effort?

stsydow:
I was away due to cold the last week. 

What I got and would like to address "soon TM" is moving 
the calculation of back and front pitch should move to the base part.

But I am wondering: Should we couple back and front pitch so that the armhole width stays constant? 
(moving in at the front would ease in the back)

nicholasdower:
I suggest we close this PR for now and open a new PR / reopen when it is time to take another look. If I don't hear any objections I will close this in a couple of days.

stsydow:
I guess you can close it. 
I won't make it in the next week to build a proper fix and can open a new PR when ready.
Do you want to open an issue for the missing `backArmholePitch` or just wait of a PR?

joostdecock:
Issue for `backArmholePitch` here: #2057 

Closing this, but happy to see more improvements from @stsydow going forward.

PR #1967, opened 2022-03-01
updated side seam calculation

anna-puk:
should work for a larger variety of proportions;
this adds two required measurements; consider making them optional


nicholasdower:
Hi @anna-puk! Nice to meet you and thank you for contributing!

Unfortunately, I am seeing some unexpected behavior when I try out your changes. Could you tell me whether you were able to start the development environment locally and if so, whether you see the same issues I am seeing?
image

Compare this to the site without these changes:
image

packages/ursula/config/index.js
```diff
@@ -38,15 +39,16 @@ export default {
     gussetRatio: 0.7, // Relationship between front and back gusset widths
 
     // Percentages
-    gussetWidth: { pct: 7.2, min: 4, max: 12 }, // Gusset width in relation to seat
+    gussetWidth: { pct: 7.2, min: 2, max: 12 }, // Gusset width in relation to seat
     gussetLength: { pct: 12.7, min: 10, max: 16 }, // Gusset length in relation to seat
-    fabricStretch: { pct: 15, min: 5, max: 25 },
-    rise: { pct: 46, min: 30, max: 100 },
+    fabricStretch: { pct: 15, min: 0, max: 100 }, // temporary extension for testing
```
nicholasdower, line 44: Maybe you meant to remove this?

packages/ursula/example/package.json
```diff
@@ -15,21 +15,22 @@
     "@freesewing/mui-theme": "latest",
```
nicholasdower, line 15: FYI, changes to this file will be overwritten the next time `yarn reconfigure` is run.

Were these changes required for you to start the example app? If not, I suggest reverting this file. If so, we should look into why. Note that I am able to start the example app on your branch even without these changes.

packages/ursula/src/front.js
```diff
@@ -21,52 +21,71 @@ export default function (part) {
   store.set('xScale', utils.stretchToScale(options.fabricStretch))
 
   // Design pattern here
+  
+  // determine height of front part: use cross seam (and cross seam front) if selected and available
+  if (options.useCrossSeam && measurements.crossSeam) {
+	store.set('crossSeam',measurements.crossSeam)
```
nicholasdower, line 27: It looks like there are some tabs in here. Would you mind replacing them with 2 spaces?
```suggestion
    store.set('crossSeam',measurements.crossSeam)
```

packages/ursula/src/front.js
```diff
@@ -21,52 +21,71 @@ export default function (part) {
...
+  } else { // use original approximation: front and back are roughly waistToUpperLeg high, plus gusset length
+	  store.set('crossSeam',measurements.waistToUpperLeg * (1 + options.backToFrontLength) + options.gussetLength * measurements.seat)
```
nicholasdower, line 29:
```suggestion
    store.set('crossSeam',measurements.waistToUpperLeg * (1 + options.backToFrontLength) + options.gussetLength * measurements.seat)
```

packages/ursula/src/front.js
```diff
@@ -21,52 +21,71 @@ export default function (part) {
...
+  }
+  // optionally use crossSeamFront to determine relative length of front and back
+  if (options.useCrossSeam && measurements.crossSeamFront) { // subtract half the gusset length from cross seam front
+    store.set('frontHeight',measurements.crossSeamFront - measurements.seat*0.5*options.gussetLength)
+  } else { // subtract gusset length, divide by roughly 2
+	store.set('frontHeight',(store.get('crossSeam') - options.gussetLength * measurements.seat)/(1 + options.backToFrontLength))
```
nicholasdower, line 35:
```suggestion
    store.set('frontHeight',(store.get('crossSeam') - options.gussetLength * measurements.seat)/(1 + options.backToFrontLength))
```

packages/ursula/src/front.js
```diff
@@ -21,52 +21,71 @@ export default function (part) {
...
+  }
+
```
nicholasdower, line 37: There is xtra blank line here containing a tab character. I suggest simply removing the line
```suggestion
  }
```

anna-puk:
> Hi @anna-puk! Nice to meet you and thank you for contributing!
> 
> Unfortunately, I am seeing some unexpected behavior when I try out your changes. Could you tell me whether you were able to start the development environment locally and if so, whether you see the same issues I am seeing?

Yes, I was able to start the development environment. I actually saw the same issue you've shown here, but thought it was due to previously selected options being sticky... (By the way, the pattern as drafted here is an extreme case: the leg opening is set higher than the rise. I decided to deal with that by giving the side seam a height of zero, which seems reasonable as an option but not as a default.)

I can update the default values for the options to get to more-or-less the same shape as before.

Is that the right thing to do, though? How would it affect users who have saved a previous version of the pattern, with options configured to their liking?

Also, I can probably work out the conversion factors so that the same percentages will give roughly the same shape as before, but that would look rather odd in the code. Is there a guideline for this kind of thing? If I only changes the default values, how would that affect users who have saved a previous version of the pattern, with options configured to their liking?

packages/ursula/config/index.js
anna-puk: Oh, that comment was not as descriptive as I thought... I think we *do* need to extend the range, though probably not quite this far. While the pattern is (hopefully) being tested, I'd like to leave it like this.

packages/ursula/src/front.js
anna-puk: Done!

packages/ursula/src/front.js
anna-puk: (also done)

nicholasdower:
> > Hi @anna-puk! Nice to meet you and thank you for contributing!
> > Unfortunately, I am seeing some unexpected behavior when I try out your changes. Could you tell me whether you were able to start the development environment locally and if so, whether you see the same issues I am seeing?
> 
> Yes, I was able to start the development environment. I actually saw the same issue you've shown here, but thought it was due to previously selected options being sticky... (By the way, the pattern as drafted here is an extreme case: the leg opening is set higher than the rise. I decided to deal with that by giving the side seam a height of zero, which seems reasonable as an option but not as a default.)

I didn't get the part about it being an extreme case. I was seeing this behavior for all of the default size. That said, I only see this behavior on the original commit (7faaf27dd4da07424bb4c08642bc65e75b993fd4). It is no longer happening on the current commit (c96b681986bb1e2658494ab51024a4442812e5cb).
> 
> I can update the default values for the options to get to more-or-less the same shape as before.
> 
> Is that the right thing to do, though? How would it affect users who have saved a previous version of the pattern, with options configured to their liking?

Sorry, I also didn't understand this. But maybe this is no longer relevant? This is the diff I see now for size 36 (for instance):
image

> 
> Also, I can probably work out the conversion factors so that the same percentages will give roughly the same shape as before, but that would look rather odd in the code. Is there a guideline for this kind of thing? If I only changes the default values, how would that affect users who have saved a previous version of the pattern, with options configured to their liking?

packages/ursula/config/index.js
nicholasdower: Do you mean tested while reviewing this PR?

nicholasdower:
I just realized that there was [a discussion](https://discord.com/channels/698854858052075530/757631205804998759/947628651992739851) in #pattern-design about this change involving @nataliabarnhart and @joostdecock. I imagine they will want to review.

packages/ursula/config/index.js
nicholasdower: I just found the thread about this in Discord. I get it now.

nataliasayang:
> I just realized that there was [a discussion](https://discord.com/channels/698854858052075530/757631205804998759/947628651992739851) in #pattern-design about this change involving @nataliabarnhart and @joostdecock. I imagine they will want to review.

Hey Nick! Would you mind editing your comment to include my current username, @nataliasayang ?

nataliasayang:
Hey @nicholasdower ! Would you mind editing your comment to include my current username, @nataliasayang ?

joostdecock:
Hi @anna-puk pardon me for being so slow to respond, I am struggling to handle all that's on my plate.

I appreciate you took the time to improve this pattern, and I'd very much would like to merge this.
Before we do though, a few remarks. They might not all be from your PR as I just walked through the various option tests.

**Farbic stretch**

The fabric stretch option is not taken into account when determining the control points for those points where waistline meets side seam:
Screenshot 2022-03-17 at 18-23-53 Screenshot

**Elastic stretch**

This option seems to do nothing. Should we drop it?

**Rise**

The rise option falls apart at lower settings and does seem to have an outsized effect on the shape of the leg opening:
Screenshot 2022-03-17 at 18-27-41 Screenshot

**Leg opening**

The leg opening option should always keep at least some side seam:
Screenshot 2022-03-17 at 18-28-57 Screenshot

**Back exposure**

The back exposure option should be constrained a bit so that it doesn't make those weird wings:
Screenshot 2022-03-17 at 18-31-58 Screenshot


I suspect most of these can be resolved with tweaking the min and max for options in the config, for others there's perhaps some small tweaks to be made.

If you feel up to the task, it would be great if you could take a stab at it :)

anna-puk:
@joostdecock: Thanks for having a look! I'm not sure I can fix _all_ of these, but I'm happy to have a look while I'm still familiar with the code!

For the **Fabric Stretch** I've implemented a fix (see my most recent commit).

**Elastic Stretch**: no corresponding pattern piece is drawn, but the pattern contains two lines of text indicating the required lengths of elastic. These values are affected by the Elastic Stretch option, so it seems to be working as intended.

**general remark about options**
I think this design may be a bit unusual in that a) the same design is used for rather different styles and b) the drafted pattern (even the thumbnail version) is a reasonable representation of the end product. Visually comparing the pattern pieces to your favorite brand of underwear seems like it would work well here. Because of that, I think we can anticipate the user playing around with the various options, and then I'd prefer to set the option limits rather wide, so they don't become limiting. On the other hand, I can also imagine a design rule saying that we only want to generate patterns which are somewhat reasonable... Do we have any rules/guidelines for this? (Might also be a good discussion for the Discord channel.)

**Rise**: I need to look into this more; I agree that there are some unintended consequences here. On the other hand, as a user, one could adjust the Rise first and then fiddle with the Leg Opening until it looked right... 

**Leg Opening**: I wanted to allow side seams of any length, even really short ones (which is roughly like a thong, not entirely unbelievable). Other than zero, it's hard to define a minimum while respecting "use percentage options where possible". I tried it out and found that while playing with the sliders, it feels pretty natural when you start from a leg opening below the rise. If the leg opening is already set rather high and then you lower the rise, that one does feel pretty awkward.

**Back Exposure**: I agree that it looks odd for the negative values. I do see a bit of that 'wing' effect on mass-produced underwear with high coverage, so there might be something to it after all. We may need some fit testing to see how it works out.

joostdecock:
> On the other hand, I can also imagine a design rule saying that we only want to generate patterns which are somewhat reasonable... Do we have any rules/guidelines for this? (Might also be a good discussion for the Discord channel.)

In general, although I doubt this is written down somewhere, we should strive to avoid *footguns*. As in, making it all too easy for people to come up with ways to mess it up.

> Other than zero, it's hard to define a minimum while respecting "use percentage options where possible". I tried it out and found that while playing with the sliders

What I typically do in a case like this is to put a hard limit in the code. Like, we'll respect the option but if to goes below X then option be damned, we'll set it to X.

nataliasayang:
> The back exposure option should be constrained a bit so that it doesn't make those weird wings:

@joostdecock, see:

* https://www.ae.com/us/en/p/aerie/undies/multipack-underwear/aerie-ribbed-seamless-boybrief-underwear-3/1774_7397_900
* https://tommyjohn.com/products/womens-cool-cotton-high-rise-brief?variant=124968
* https://www.uwilawarrior.com/products/no-brainer

Eliminate it if you want, but it was very much intentional in the original design.

joostdecock:
Thank you @nataliasayang it goes to show how much I have to learn about women's underwear 🙈

anna-puk:
> [...] it was very much intentional in the original design.

Oh, seeing those images is very helpful, thanks! I could tell from the code/comments that the 'wings' were intentional, so I didn't want to eliminate the option.

With the changes from my latest commit, the flare is positioned better and will bulge out less when the leg opening is smaller. (Sorry, I'm not sure how to include an image. :( )

anna-puk:
I think we need to figure out (as a community) whether this ought to be an update to ursula or a new design. Pending the outcome of that discussion, I'm closing the pull request (i.e. retracting it).

PR #1978, opened 2022-03-03
Add missing pattern symlinks in Bee and Ursula

nicholasdower:


nicholasdower:
I'm thinking this one isn't too controversial. Merging.

PR #1990, opened 2022-03-10
Making it more clear you need four pocket facings

TriploidTree:
I cut two, not paying close attention and then halfway through construction realised I needed 2x2 = 4

### DIFF
--- a/markdown/org/docs/patterns/charlie/cutting/en.md
+++ b/markdown/org/docs/patterns/charlie/cutting/en.md
@@ -4,7 +4,7 @@
 - Part **2**: **2 x** _with good sides together_.
 - Part **4**: **4 x**
 - Part **6**: **2 x**
-- Part **8**: **2 x** 2 _with good sides together_
+- Part **8**: **4 x** 2 pairs _with good sides together_
 - Part **9**: **2 x** _with good sides together_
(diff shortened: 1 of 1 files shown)


joostdecock (changes requested):
This would make it say `4 x 2 pairs` which is 8 in total and unless I'm mistaken, that's incorrect.

So if we want to make it clearer, I feel we should do something like ` 4x (2 x 2 pairs)`

TriploidTree:
I think the challenge is whether what's in bold can be used without the following text, so I read "x2" rather than "x2 pairs". 

Maybe `4x (2 matching pairs)`. or we move pairs into the bold so it reads as "x2 pairs" in bold, with the remainder unformatted?

TriploidTree:
I think *4X* (2 x  2 sets with good sides together) might be a good compromise. 
If I read "2 pairs", I think 4 pieces, so "2x2 pairs" is 8, as a pair is already 2 parts. Isn't English fun. 

That said, I'm more interested in "4X" when reading a cutting list, so maybe 4X (2 x 2 pairs) would be fine as I'd ignore the part in brackets. I'm not sure whether that's an experienced sewist thing though.

joostdecock:
Thank @TriploidTree

PR #1995, opened 2022-03-12
feat: First stab at custom layout

joostdecock:
This adds a React component to handle custom layouts.
This React component is a long way from perfect, but it's a start.
    
There are several reasons that (at least in my opinion) implementing this is non-trivial:
    
**React re-render vs DOM updates**

For performance reasons, we can't re-render with React when the user drags a
pattern part (or rotates it). It would kill performance.
So, we don't re-render with React upon dragging/rotating, but instead manipulate the DOM directly.
    
So far so good, but of course we don't want a pattern that's only correctly laid out in the DOM. We want to updat the pattern gist so that the new layout is stored. For this, we re-render with React on the end of the drag (or rotate).
    
Handling this balance between DOM updates and React re-renders is a first contributing factor to why this component is non-trivial
    
**SVG vs DOM coordinates**
    
When we drag or rotate with the mouse, all the events are giving us coordinates of where the mouse is in the DOM.
    
The layout uses coordinates from the embedded SVG which are completely different.
So we need to make this translation and that adds complexity.
    
**Part-level transforms**
    
It's not just that the DOM coordinates and SVG coordinate system is different, each part also has it's own transforms applied and because of this behaves as if they have their own coordinate system.
    
In other words, a point (0,0) in the part is not the top-left corner of the page. In the best-case scenario, it's the top-left corner of the part. But even this is often not the case as parts will have transforms applied to them.
    
**Flip along X or Y axis**
    
Parts can be flipped along the X or Y axis to facilitate a custom layout. This is handled in a transform, so the part's coordinate's don't actually change. They are flipped late into the rendering process (by the browser displaying the SVG).
    
Handling this adds yet more mental overhead
    
**Bounding box**
    
While moving and rotating parts around is one thing. Recalculating the bounding box (think auto-cropping the pattern) gets kinda complicated because of the reasons outlined above.
    
We are currently handling a lot in the frontend code. It might be more elegant to move some of this to core. For example, core expects the custom layout to set the widht and height rather than figuring it out on its own as it does for auto-generated layouts.
    
**Known issues**
    
- Rotating gets a little weird sometimes as the part rotates around it's center in the SVG coordinate system, but the mouse uses it's own coordinates as the center point that's used to calculate the angle of the rotation
    
- Moving parts into the negative space (minus X or Y coordinated) does not extend the bounding box.
    
- Rotation gets weird when a part is mirrored
    
- The bounding box update when a part is rotated is not entirely accurate
    
I've sort of left it at this because I'm starting to wonder if we should perhaps re-think how custom layouts are supported in the core. And I would like to discuss this with the @freesewing/developers .


nicholasdower:
Wow! Really cool and very fun to use. I would definitely use this every time.

Note that I wasn't able to compile with your latest lint commit. I had to revert it locally.

Also, moving flipped parts seems a little strange. The part reverts to it's original orientation while dragging.

nicholasdower:
One more thing I noticed, updating the pattern by switching to draft and changing some option sometimes has unexpected impact on the layout.

packages/freesewing.shared/components/workbench/layout/draft.js
```diff
@@ -0,0 +1,419 @@
...
+
+  // State variable to switch between moving or rotating the part
+  const [rotate, setRotate] = useState(false)
+
+  // Initialize drag handler
+  useEffect(() => {
+    handleDrag(select(partRef.current))
+  }, [rotate])
+
+  // These are kepts as vars because re-rendering on drag would kill performance
```
nicholasdower, line 169:
```suggestion
  // These are kept as vars because re-rendering on drag would kill performance
```

packages/freesewing.shared/components/workbench/layout/print/pagesize-picker.js
```diff
@@ -0,0 +1,78 @@
+import PageSizeIcon from 'shared/components/icons/page-size'
+import { useTranslation } from 'next-i18next'
+import Popout from 'shared/components/popout'
+
+
+const sizes = ['a4', 'a3', 'a2', 'a1', 'a0', 'letter', 'tabloid']
+
+
```
nicholasdower, line 8:
```suggestion
```

packages/freesewing.shared/components/workbench/layout/print/pagesize-picker.js
```diff
@@ -0,0 +1,78 @@
+import PageSizeIcon from 'shared/components/icons/page-size'
+import { useTranslation } from 'next-i18next'
+import Popout from 'shared/components/popout'
+
+
```
nicholasdower, line 5:
```suggestion
```

nicholasdower:
This looks good to me and IMO should be 🚢 . I get that it might not be completely ready but it is already useful. Maybe it should be tagged as experimental? I don't feel strongly.

I didn't review this very carefully because there is a ton of code and a ton of complexity. I am happy to do a more thorough review overall or of specific parts if you think it would be useful, but it's also cool with me if you just want to merge.

nicholasdower (approved):

joostdecock:
Yeah, I know there's rough edges, but I sort of wanted to put it out there because I feel there's been other stuff that's been piling up.

Since the entire new dev environment isn't really used in anything user-facing (yet) I'm happy to merge it as-is and further polish it down the line.

nicholasdower:
> Yeah, I know there's rough edges, but I sort of wanted to put it out there because I feel there's been other stuff that's been piling up.
> 
> Since the entire new dev environment isn't really used in anything user-facing (yet) I'm happy to merge it as-is and further polish it down the line.

I say go for it. It already looks awesome.

joostdecock:
Tests pass now, merging

PR #2036, opened 2022-03-27
Woutervdub bella shoulder ease

woutervdub:
Changing constant option shoulderToShoulderCorrection to percentage (advanced) option shoulderToShoulderEase.


nicholasdower:
@woutervdub FYI there is also a place to add a title and description here: https://github.com/freesewing/freesewing/blob/develop/packages/i18n/src/locales/en/options/bella.yml

packages/bella/config/index.js
```diff
@@ -52,7 +52,7 @@ export default {
     shoulderSlopeBack: 1.23,
     neckWidthBack: 0.197,
     neckWidthFront: 0.17,
-    shoulderToShoulderCorrection: 0.995,
+    // shoulderToShoulderCorrection: 0.995,
```
nicholasdower, line 55: Ok to remove?
```suggestion
```

packages/bella/src/back.js
```diff
@@ -21,8 +21,8 @@ export default function (part) {
   points.cbNeckCp1 = new Point(points.hps.x * 0.8, points.cbNeck.y)
   let slope = measurements.shoulderSlope * options.shoulderSlopeBack * -1
   points.shoulder = utils.beamsIntersect(
-    new Point((measurements.shoulderToShoulder * options.shoulderToShoulderCorrection) / 2, 0),
-    new Point((measurements.shoulderToShoulder * options.shoulderToShoulderCorrection) / 2, 100),
+    new Point((measurements.shoulderToShoulder * (1+options.shoulderToShoulderEase)) / 2, 0),
+    new Point((measurements.shoulderToShoulder * (1+options.shoulderToShoulderEase)) / 2, 100),
```
nicholasdower, line 25:
```suggestion
    new Point((measurements.shoulderToShoulder * (1 + options.shoulderToShoulderEase)) / 2, 0),
    new Point((measurements.shoulderToShoulder * (1 + options.shoulderToShoulderEase)) / 2, 100),
```

packages/bella/src/front-side-dart.js
```diff
@@ -21,7 +21,7 @@ export default function (part) {
   points.hpsCp2 = new Point(points.hps.x, points.cfNeck.y / 2)
   let slope = measurements.shoulderSlope * (2 - options.shoulderSlopeBack) * -1
   let xShoulder =
-    ((measurements.shoulderToShoulder * options.shoulderToShoulderCorrection) / 2) *
+    ((measurements.shoulderToShoulder * (1+options.shoulderToShoulderEase)) / 2) *
```
nicholasdower, line 24:
```suggestion
    ((measurements.shoulderToShoulder * (1 + options.shoulderToShoulderEase)) / 2) *
```

nicholasdower (approved):
Looks great! A few small suggestions.

nicholasdower (changes requested):
I just noticed the build is failing. It looks like bee now expects to have this property defined.

Also, I'm not sure whether this is intentional but when setting this property to the minimum value there is a noticeable jump in the pattern. This is even more noticeable if I add the property to bee.

packages/bella/config/index.js
woutervdub: I left it in place, commented out, for history sake. I like to keep removals of code in the file for a while as an easy indication of what was changed. I guess with github keeping track, we don't really have to.

packages/bella/src/back.js
woutervdub: Isn't there some formatting process that makes everything uniform? I bothered doing so quite a while ago (since it doesn't like my code formatting preferences ;-)

woutervdub:
> I just noticed the build is failing. It looks like bee now expects to have this property defined.
> 
> Also, I'm not sure whether this is intentional but when setting this property to the minimum value there is a noticeable jump in the pattern. This is even more noticeable if I add the property to bee.

I don't see anything weird with different values for the option. Doing a 'test `bella`' results in a normal distribution of the shoulder point.

`bee` only has the `shoulderToShoulderCorrection` option in the config file, no where in the actual code. There's something wrong with my dev environment, cause I can't run `bee` locally (it doesn't like the config file, and neither does VS Code).

I could keep the old option, and just add the new ease option:
```
 let xShoulder =
    ((measurements.shoulderToShoulder * options.shoulderToShoulderCorrection * (1 + options.shoulderToShoulderEase)) / 2) *
```
And just set the default for `shoulderToShoulderEase` to 0.

packages/bella/config/index.js
nicholasdower: My preference is to never allow commented code but I am happy to follow an existing convention.

woutervdub: I have no idea if there is a convention. ;-)

packages/bella/src/back.js
nicholasdower: > Isn't there some formatting process that makes everything uniform? I bothered doing so quite a while ago (since it doesn't like my code formatting preferences ;-)

In Visual Studio Code you can run "Format Document". In this file it will make the suggested changes.

On Mac you can accomplish this with any of the following:

- To format an entire file, while in the file type ⇧ ⌥ F
- Alternatively, type ⇧ ⌘ P then start typing "format document" and hit enter once it autocompletes.
- To format a highlighted selection, highlight some code and then type: ⌘ K, then ⌘ F
- Alternatively, type ⇧ ⌘ P then start typing "format selection" and hit enter Enter once it autocompletes.

packages/bella/config/index.js
joostdecock: Code style is enforced by prettier, but comments are comments. They will not be removed.

Git is where history lives. I don't mind leaving comments in while things are a WIP, but I also feel we should strip them prior to merging.

packages/bella/src/back.js
joostdecock: Or run `yarn prettier` in the monorepo root :)

packages/bella/src/front-side-dart.js
joostdecock: This seems like Nick fixed a bug here. Wouter?

woutervdub: That's just a couple of extra spaces (as `prettier` will overrule, or not, anyway)
Or am I missing something?

woutervdub:
Ah, yes, that makes sense. 

It is somewhat confusing to have to specify all the options in an inherited pattern, specially when the parent is modified. I expected it to grab whatever the value of the option is in the parent if the inherited child doesn't specify it.

packages/bella/src/back.js
nicholasdower: I grepped for "format" before posting this. Didn't know about "prettier"! Thanks!

nicholasdower:
> Also, I'm not sure whether this is intentional but when setting this property to the minimum value there is a noticeable jump in the pattern. This is even more noticeable if I add the property to bee.

I played around with this a bit and it seems it is not an issue. It looks like the jump I was seeing was simply the result of the following code requiring one more step to reach `target`:
```javascript
   while (points.bustSide.x < target && steps < 80) {
     for (const p of rot) points[p] = points[p].rotate(increment, points.armholePitch)
     angle += increment
     points.bustSide = utils.beamIntersectsY(points.armhole, points.sideHem, points.bust.y)
     steps++
   }
```

woutervdub:
> @woutervdub FYI there is also a place to add a title and description here: https://github.com/freesewing/freesewing/blob/develop/packages/i18n/src/locales/en/options/bella.yml

Implemented

woutervdub:
> -      { bellaAdvanced: ['frontShoulderWidth', 'fullChestEaseReduction', 'highBustWidth'] },
> +      { bellaAdvanced: ['frontShoulderWidth', 'fullChestEaseReduction', 'highBustWidth', 'shoulderToShoulderEase'] },

Implemented

packages/bella/src/back.js
woutervdub: That's why I stopped bothering with formatting. I have prettier integrated with VS Code (my preferred editor), but my settings are different ;-)

packages/bee/config/index.js
```diff
@@ -38,7 +38,7 @@ export default {
     advanced: [
       { bellaDarts: ['backDartHeight'] },
       { bellaArmhole: ['armholeDepth', 'frontArmholePitchDepth'] },
-      { bellaAdvanced: ['frontShoulderWidth', 'fullChestEaseReduction', 'highBustWidth'] },
+      { bellaAdvanced: ['frontShoulderWidth', 'fullChestEaseReduction', 'highBustWidth'], 'shoulderToShoulderEase' },
```
nicholasdower, line 41:
```suggestion
      { bellaAdvanced: ['frontShoulderWidth', 'fullChestEaseReduction', 'highBustWidth', 'shoulderToShoulderEase'] },
```

packages/i18n/src/locales/en/options/bella.yml
```diff
@@ -10,6 +10,10 @@ bustSpanEase:
   title: Bust span ease
   description: Controls the amount of (horizontal) ease added to your bust span when locating the bust point.
 
+shoulderToShoulderEase:
+  title: Shoulder to Shoulder ease
+  description: Controls the amount of ease between you shoulders. Initially set to -.5% because Bella implements a block that is used in the industry.
```
nicholasdower, line 15:
```suggestion
  description: Controls the amount of ease between your shoulders. Initially set to -.5% because Bella implements a block that is used in the industry.
```

packages/bee/config/index.js
woutervdub: Oops. Thanks!

packages/i18n/src/locales/en/options/bella.yml
```diff
@@ -10,6 +10,10 @@ bustSpanEase:
   title: Bust span ease
   description: Controls the amount of (horizontal) ease added to your bust span when locating the bust point.
 
+shoulderToShoulderEase:
```
nicholasdower, line 13: Looks like this needs to be added to all locales. I believe that is why the build is failing. Added a commit.

bobgeorgethe3rd:
> shoulderToShoulderEase may be a constant for Bee. I will check when the Bella update goes through whether it affects Bee enough/at all to be a varaible.

Got to round to checking, It causes minor changes to the draft so can be left as a variable.

PR #2044, opened 2022-03-29
fix(core): Handle path split on start/end of path segments

joostdecock:
Wouter found a novel way to break things. I fixed in in his fork, and this is the backported fix.

Essentially, it checks that the point we're splitting a path on lies on the start/end of a path segment (a curve or line) and if so doesn't split the path, but rather divides it at that point.

This is needed because splitting on the actual endpoint (or a point very close to it) will result in a path operation that draws a line or curve between two (almost) identical points.
And in the `path.offset()` method, we thrown out those sort of things, resulting in weird behaviour in `path.offset` (and the `pd` macro).

Since offsetting Bezier paths is dark voodoo stuff, best to handle this prior to that, and don't split beziers (or lines) when the split point is one of its end/start points.

### DIFF
--- a/packages/core/src/path.js
+++ b/packages/core/src/path.js
@@ -747,12 +747,26 @@ Path.prototype.split = function (point) {
   if (point instanceof Point !== true)
     this.raise.error('Called `Path.split(point)` but `point` is not a `Point` object')
   let divided = this.divide()
-  let firstHalf = false
-  let secondHalf = false
+  let firstHalf = []
+  let secondHalf = []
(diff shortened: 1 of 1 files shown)


woutervdub:
I didn't realise that could be an issue. I just looked at the problem and figured that path splitting was the easiest and most self-documenting solution to a problem. Sorry to have caused you more work.

nicholasdower (approved):

joostdecock:
No worries, finding novel ways to break it is an important skill ;)

PR #2060, opened 2022-04-03
WIP: Conversion to the new `waistToArmhole` measurement

joostdecock:
This PR is a work in progress to add the new `waistToArmhole` measurement, and convert patterns to adopt its use.

High-level overview of changes:

- Added the `waistToArmhole` measurement to modes, utils (neckstimate) and i18n
 - Brian now uses/requires the `waistToArmhole` measuremnt
 - Brian uses this measurement to calculate the armhole depth/height
 - The `armholeDepthFactor` in brian is deprecated, I've kept it in the config but it's no longer used
 - Instead, armhole dept is controlled by the new `arhmoleHeight` factor
 - The sleevecap height is no longer depending on the biceps circumference, but now is based on the armhole height
 - The sleevecap height is further controlled by the new `sleevecapHeight` option which controls the sleevecap height independently of the armhole depth/height
 - There's a new `sleeveBulge` option in Brian (unrelated but while I was there I added it)

**Things not to forget**
- Update the sleevecap docs to mention the new `sleevecapHeight` option

### DIFF
--- a/markdown/org/docs/patterns/brian/options/sleevecapheight/en.md
+++ b/markdown/org/docs/patterns/brian/options/sleevecapheight/en.md
@@ -1,0 +1,11 @@
+---
+title: "Sleevecap height"
+---
+
+This option controls the height of the sleevelcap.
+
+The sleevecap height controls how formal a sleeve looks. A sleeve with a higher sleevecap will
+be more dowward pointing (like a jacket) and give a more formal look, but allow for less movement.
+A sleeve with a lower sleevecap will be more outward pointing (like a t-shirt) and give a more informal
+look, with more movement.
+
--- a/markdown/org/docs/patterns/holmes/fabric/en.md
+++ b/markdown/org/docs/patterns/holmes/fabric/en.md
@@ -4,7 +4,9 @@ title: "Holmes deerstalker hat: Fabric Options"
 
 ### Main Fabric
 
-For the main fabric of your hat, a thick fabric will work best to hold the shape but if you wanna go for a lighter fabric you can interface it to give it more body. Generally you want to use **Wools** of coating weight but you can experiment with different weights and fabrics with the relevant interfacing. <Note>
+For the main fabric of your hat, a thick fabric will work best to hold the shape but if you wanna go for a lighter fabric you can interface it to give it more body. Generally you want to use **Wools** of coating weight but you can experiment with different weights and fabrics with the relevant interfacing. 
+
+<Note>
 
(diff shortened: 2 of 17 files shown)


markdown/org/docs/patterns/brian/options/sleevecapheight/en.md
```diff
@@ -0,0 +1,11 @@
+---
+title: "Sleevecap height"
+---
+
+This option controls the height of the sleevelcap.
+
+The sleevecap height controls how formal a sleeve looks. A sleeve with a higher sleevecap will
+be more dowward pointing (like a jacket) and give a more formal look, but allow for less movement.
```
luizfzs, line 8:
```suggestion
be more downward pointing (like a jacket) and give a more formal look, but allow for less movement.
```

markdown/org/docs/patterns/holmes/fabric/en.md
```diff
@@ -4,7 +4,9 @@ title: "Holmes deerstalker hat: Fabric Options"
 
 ### Main Fabric
 
-For the main fabric of your hat, a thick fabric will work best to hold the shape but if you wanna go for a lighter fabric you can interface it to give it more body. Generally you want to use **Wools** of coating weight but you can experiment with different weights and fabrics with the relevant interfacing. <Note>
+For the main fabric of your hat, a thick fabric will work best to hold the shape but if you wanna go for a lighter fabric you can interface it to give it more body. Generally you want to use **Wools** of coating weight but you can experiment with different weights and fabrics with the relevant interfacing.
```
luizfzs, line 7: Nit: are we generally OK with using informal expressions such as `wanna`?

packages/i18n/src/locales/en/options/brian.yml
```diff
@@ -6,6 +6,10 @@ armholeDepthFactor:
   title: Armhole depth factor
   description: Controls the depth of the armhole. Higher values make a deeper armhole.
 
+armholeHeight:
+  title: Armhole height
+  description: Controls the height of the armhole. Higher values result in a more tight armhole.
```
luizfzs, line 11: My understanding is that this is meant to control the bottom point of the armhole, given that a higher value (which I assume means 'go up more') means a tighter armhole.

However when I read `armholeHeight` I immediately think 'how tall the armhole is' which gets confusing when reading that a higher value means a smaller armhole.

luizfzs (commented):
I have some typo corrections and a comment about naming.

PR #2063, opened 2022-04-04
Fix broken build.

nicholasdower:


nicholasdower:
@joostdecock I didn't intend to modify so many files. Most of the changes are the result of running `yarn build` in the 18n dir. Does this look ok?

joostdecock:
Yeah, there's a differently formatted build output for NextJS. 
Perhaps those files should be gitignored, but it's not a problem. 

I pushed i18n changes but didn't rebuild. You did so you're seeing the side-effects of those changes.

And since these files are the result of a build we could/should perhaps ignore themm

PR #2086, opened 2022-04-07
fix(lunetius-option-desc): to upper leg

starfetch:
changed the option description for `toUpperLeg` from "Below the hips" to "To the upper leg", as that is technically correct (it uses the `waist to upper leg` measurement). It sounds a bit more clunky, so opinions? (I'm also not sure if changing this here impacts anything else apart from the translation files, hence a PR and not a direct commit.)

### DIFF
--- a/packages/i18n/src/locales/en/options/lunetius.yml
+++ b/packages/i18n/src/locales/en/options/lunetius.yml
@@ -13,5 +13,5 @@ length:
     toKnee: On the knee
     toBelowKnee: Below the knee
     toHips: On the hips
-    toUpperLeg: Below the hips
+    toUpperLeg: To the upper leg
     toFloor: To the floor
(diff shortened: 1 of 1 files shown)


packages/i18n/src/locales/en/options/lunetius.yml
```diff
@@ -13,5 +13,5 @@ length:
     toKnee: On the knee
     toBelowKnee: Below the knee
     toHips: On the hips
-    toUpperLeg: Below the hips
+    toUpperLeg: To the upper leg
```
joostdecock, line 16: Since `toHips` is `On the hips` and `toKnee` is `On the knee`, perhaps `toUpperLeg` can be `On the thigh`

starfetch: Oh, great idea! Done.

PR #2103, opened 2022-04-13
fix(link) Broken Links, Counter and ButtonHoles

bobgeorgethe3rd:
Noticed this when trying to look it up. The counter page exists it is just not registered as count so the link needed changing :)

Edit: Noticed the buttonhole pages had broken links as well

Changes
- Changed link from "count" to "counter"
- Added api to buttonhole links

### DIFF
--- a/markdown/dev/reference/api/config/options/en.md
+++ b/markdown/dev/reference/api/config/options/en.md
@@ -129,7 +129,7 @@ to determine whether an option should be shown or not.
 
 [const]: /reference/api/config/options/const
 
-[count]: /reference/api/config/options/count
+[count]: /reference/api/config/options/counter
 
 [deg]: /reference/api/config/options/deg
 
--- a/markdown/dev/reference/api/config/options/en.md
+++ b/markdown/dev/reference/api/config/options/en.md
@@ -27,7 +27,7 @@ There are the five option types that an aspiring pattern designer should be
 familiar with:
 
 1. [**boolean** options][bool] are for yes/no choices
-2. [**counter** options][counter] are for integer values
+2. [**counter** options][count] are for integer values
 3. [**degree** options][deg] are for degrees
 4. [**list** options][list] are for a list of possible choices
 5. [**percentage** options][pct] are for percentages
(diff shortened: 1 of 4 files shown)


markdown/dev/reference/api/config/options/en.md
```diff
@@ -27,7 +27,7 @@ There are the five option types that an aspiring pattern designer should be
 familiar with:
 
 1. [**boolean** options][bool] are for yes/no choices
-2. [**counter** options][count] are for integer values
+2. [**counter** options][counter] are for integer values
```
nicholasdower, line 30: I believe you will also need to update line 132 like:
```diff
-[count]: /reference/api/config/options/count
+[counter]: /reference/api/config/options/counter
```

bobgeorgethe3rd: Done Lol of course I missed the actual link gg me

bobgeorgethe3rd:
Noticed a few more broken links so slipped them in.

markdown/dev/reference/api/config/options/en.md
joostdecock: What's there is not going to result in a broken link.

This is a markdown link using a reference. So:

```md
[link text][link reference]
```

Then you can define the reference elsewhere on the page:

[link reference]: /some/link/here

What's currently in this PR is this:

```md
[**counter** options][counter] are for integer values

[count]: /reference/api/config/options/counter
```

So the link reference that's defined is `count` but the reference used is `counter`.

Long story short: it seems there's some confusion between the URL and the link reference (which is just a unique ID and can be anything)

joostdecock (changes requested):
These are good fixes, but the `count` link reference should be used, not `counter` (but adapting the URL is good)

bobgeorgethe3rd:
I see what you saying have used [count] and just changed the url.

nicholasdower (approved):
Looks great. :shipit:

nicholasdower:
@joostdecock it says you requested a change but I don't think that is true:
image

Going to merge but please let me know if something needs to be followed up on.

joostdecock:
@nicholasdower 

> @joostdecock it says you requested a change but I don't think that is true:

I did, but @bobgeorgethe3rd already made the change:

> I see what you saying have used [count] and just changed the url.

All's good :+1:

PR #2114, opened 2022-04-16
Update back.js

woutervdub:
This addresses the issue #2051. When the HPS to Bust becomes too small, the calculation of 'bustWidthBack' would throw an error because the line along which this was measured would no longer intersect with the side seam. This solution solves this by creating an artificial extension on the side seam that is used to create this point at which the measurement is made.
There is still an issue where if the value keeps decreasing, the side dart itself will no longer be on the side seam. This is an issue in front.js.

### DIFF
--- a/packages/bella/src/back.js
+++ b/packages/bella/src/back.js
@@ -112,14 +112,6 @@ export default function (part) {
   }
 
   // Store the back width at bust level
-  points.bustSide = utils.curveIntersectsY(
-    points.waistSide,
-    points.waistSideCp2,
-    points.armhole,
-    points.armhole,
-    measurements.hpsToBust
-  )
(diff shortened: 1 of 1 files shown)


woutervdub:
How do I do that? I just went to the development branch of Freesewing, and changed the file. Then created a new branch for the pull/commit. 
sorry, not much of a GitHub person.

nicholasdower:
> How do I do that? I just went to the development branch of Freesewing, and changed the file. Then created a new branch for the pull/commit. sorry, not much of a GitHub person.

Please ignore. I just assumed it was due to your branch being behind. Looks like the build was broken in #2108.

nicholasdower:
Looks like the build was broken in #2108 and fixed automatically in #2113.

I rebased you branch and force pushed. Will merge once green (unless you beat me to it).

PR #2127, opened 2022-04-20
Holmes and Bee Config Updates

bobgeorgethe3rd:
Whilst working on some patterns I noticed that the headEase option I use in my hat patterns was no longer functioning correctly and was no longer snapping. Upon inspection realised that toAbs() was now used for absolute value snap options so adjusted and checked Holmes which need changing as well and found Bee's needed the same changes as well. They should now work as intended.

Edit: Also updated the ear slightly as it didn't have a singular single seam path which was an oversight by me when splitting my bad.

Edit2: Also cleaned up Holmes's visor seam allowances as it was not using best practices.

Holmes
Changes
- headEase option now uses toAbs rather than pctBasedOn
- Ear og paths.seam has become paths.saBase
- Ear paths.hem changed to paths.hemBase
- Ear saBase and hemBase setRender to false
- Visor paths.saInner and saOuter now counter clockwise
- Visor paths.seam now uses paths.saInner and saOuter to draft
- Visor paths.sa neatened too use the new paths.saInner and saOuter
- Visor points.sa1 and sa2 plotted with simpler method.

Additions
- Ear paths.seam joining paths.hemBase and saBase

Bee
Changes
- neckTieWidth changed to toAbs() rather than pctBasedOn
- bandTieWidth changed to toAbs() rather than pctBasedOn


bobgeorgethe3rd:
Also updated the ear slightly as it didn't have a singular single seam path

Issue #2165, opened 2022-05-07
Bug report - Shin is missing grain lines


nataliasayang:
**What seems to be the problem?**

As [reported by comixminx in Discord](https://discord.com/channels/698854858052075530/757632180980547686/959835580366192741), the Shin pieces don’t have a grain line on them, though the documentation shows some sample pieces that do have grain lines printed.

**Are you a FreeSewing patron?**

 - [x] Yes, I am :hugs:
 - [ ] No, I am not :thinking:

bobgeorgethe3rd:
Fixed as of #2166

PR #2166, opened 2022-05-07
Shin - Adding Grainlines

bobgeorgethe3rd:
For issue #2165 

Using the grainlines from https://freesewing.org/docs/patterns/shin/cutting/ as reference.

Additions
- Added grainline on back,js
- Added grainline on front.js
- Added grainline on waistband.js

Changes
- Tweaked the horizontal distance of the scalebox on back.js to reduce chance of overlap with the new grainline.

I was up and simple fix, understand though if you don't like the way I have done it.

### DIFF
--- a/packages/shin/src/back.js
+++ b/packages/shin/src/back.js
@@ -117,7 +117,13 @@ export default function (part) {
       nr: 1,
       title: 'back',
     })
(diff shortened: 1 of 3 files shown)


joostdecock:
LGTM

PR #2181, opened 2022-05-16
Typo fix in freq-mistakes markdown guide

SeaZeeZee:
As it says on the tin!


joostdecock:
It's better to not mix typo changes (which are a no-brainer to merge) with more opinionated changes.

In this case, you changed:

```md
## Don't be shy to ask a friend
```

Into:

```md
## Don't be afraid to ask for help!
```

So now I've started a review :joy:

SeaZeeZee:
Oh thats my bad I could have realised that! I can reword my reword to be closer to the original for sure! The only reason I changed it is because I felt the original didn't flow off the tongue as much!

PR #2236, opened 2022-06-06
Fix hydration errors due to localstorage

eriese:
the way localstorage was working was giving me errors with hydration. These new hooks use the react lifecycle to make changes after mounting

### DIFF
--- a/packages/freesewing.dev/hooks/useApp.js
+++ b/packages/freesewing.dev/hooks/useApp.js
@@ -2,6 +2,7 @@ import { useState } from 'react'
 import set from 'lodash.set'
 // Stores state in local storage
 import useLocalStorage from 'shared/hooks/useLocalStorage.js'
+import useTheme from 'shared/hooks/useTheme'
 // Prebuild navigation
 import prebuildNavigation from 'site/prebuild/navigation.js'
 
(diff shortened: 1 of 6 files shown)


joostdecock:
Thanks for the PR @eriese 
Looking at the deloy on https://freesewing-mjzsnywj5-freesewing.vercel.app/ it seems the theme setting is lacking persistence.
When the home page loads, I can switch to the dark theme (for example). When I click **Guides** I'm back to the light theme.
Compare that to https://freesewing.dev where the theme stays, once selected.

Do you see the same? Because that feels like an issue to me?

Also (not strictly related to this PR) the first load is always the SSR page with the light theme. Do you think we should force a re-render on the client?

eriese:
> Do you see the same? Because that feels like an issue to me?
I hadn't seen that, definitely an issue. I'll look into it

> Also (not strictly related to this PR) the first load is always the SSR page with the light theme. Do you think we should force a re-render on the client?
Not sure I follow. If you're asking what I think you're asking, my implementation leaves theme undefined until mount and then defaults to user preference of light or dark, so I think that re-render is happening

joostdecock:
> so I think that re-render is happening

It is not currently happening (not in your code, in what's there now). Or rather, it is but it isn't :joy: 

Like, let's say we switch to dark mode on freesewing.dev, then do a hard reload. What we get is:

Screenshot from 2022-06-07 11-28-11

The theme switcher has correctly updated to show **Dark theme**. But the page is still with a white background (light theme).

So the point I was making -- and once again, this has nothing to do with your PR -- is that it doesn't really work as it should.
So since you're working on it, I figured I'd mention it :)

I assumed this line would take care of it, but it doesn't seem to:

https://github.com/freesewing/freesewing/blob/19db3023bc94e8a1029011a215681f9c9ab0eb45/packages/freesewing.org/components/wrappers/page.js#L57

eriese:
It looks like the failing tests are all in I18n, but please let me know if I missed something my code broke

PR #2237, opened 2022-06-06
Option improvements

eriese:
- add `hide` back to option display logic
- eliminate option sub-group and use recursion instead
- externalize options and values to make it easier to overwrite them if using freesewing as a library


joostdecock:
@all-contributors please add @eriese for code

Issue #2246, opened 2022-06-09
Manually Enter Measurements in imperial units


eriese:
**What is it that you would like to see happen?**

I'd like to be able to put in measurements in imperial to save me time converting it all before entering. 

**Are you a FreeSewing patron?**

 - [ ] Yes, I am :hugs:
 - [X ] No, I am not :thinking:

**Additional context**
The real question is whether folks think it's important to be able to put in fractions. I liked the idea until I started implementing it, and getting the right length of debouncing before formatting seemed like it was going to be a UX issue. Which is to say, I've got code ready to go for this if anyone weighs in on whether they'd prefer fractions or not fractions. Maybe a toggle to turn them on?

joostdecock:
When you say `fractions` doe you mean something like `12.25`? Because you can already input measurements as (imperial) fractions (if by fraction you mean 1/8):

![Screenshot from 2022-06-10 07-52-48](https://user-images.githubusercontent.com/1708494/172999664-61eccb52-eda2-4924-bea7-46cf8bc04fd4.png)

It's not implemented yet in the new measurements input, mostly because I never use imperial :see_no_evil: but obviously it's something we want to support. So perhaps that's what you're referring to?

Anyway, yes we obviously want to support imperial units in all its shapes and forms, but it's not clear to me what measurments input you are referring to. Current website, or future website.

joostdecock:
To be clear: Setting the `mm` state in the example above should also trigger a *save*. By writing to localstorage or sending an update to the backend for example.

eriese:
@joostdecock wait, do you mean?
![sketch viz of joost's method](https://user-images.githubusercontent.com/5298929/173125233-9dd69622-a794-46ec-afa7-aaec1a702d2a.png)

I was doing something similar, but I was also listening to the state, which is where I'm creating the timing issue, but only in the few cases where you're trying to type something like 14 3/16 but it interprets when you've only typed 14 3/1 because that's a valid fraction
![sketch viz of my method](https://user-images.githubusercontent.com/5298929/173126474-ad223010-3a8b-4a02-848d-965b7b619fc7.png)

If you aren't listening to the state, how are you updating the value when it gets changed by, say, pressing one of the preload measurements buttons?

joostdecock:
You've got it all figured out 👍

If you have something like this ready to go, I'd love to see a PR for it.

We can use the units stored in the gist and default to metric for the initial value, right?

eriese:
Wait!

[...]

PR #2248, opened 2022-06-10
#2246 support imperial measurements in new measurement inputs

eriese:
Not quite ready, still need ideas for fixing an issue with imperial:

if the user is slow in typing a decimal, it may be converted to a fraction before they have completed their input.

#2246


joostdecock:
Could we avoid all that by just not formatting the user's input?

Like, we want to display a nice fractions in the UI. But perhaps we leave the input as-is? 
Or perhaps we can trigger the formatting `onBlur` (when the input loses focus)

I'm thinking out loud here. I am happy to let you do it how you think is best. Just throwing ideas out there 😊

eriese:
I'm up for that. Or, to be more specific, I think it works if we format
everything as decimals while allowing fraction inputs as long as the
fraction part of the input is <1

I'd like to hear from others, but I guess I could also just make it and
then it's in until someone asks for different

joostdecock:
We're working on the future UI here. Nothing we do is going to break/change any UX people are familiar with today.

So I would vote for putting in place the things we've come up with here. We can tweak it or deal with corner cases later. It would already represent an great step forward.

So I would say go ahead.

I also wanted to ask if you'd be willing/open to take more responsibility within the project?
My feeling is that you're a better React developer than I am, and this is an area where there aren't many contributions. It would be nice to have someone to bounce ideas of and can also work independently to build a better experience.

I should perhaps work on my sales pitch, but my question is: how do you feel about more responsibility / a larger role in FreeSewing?

joostdecock:
@eriese There's been a lot of changes in the repo structure. I'm still finalizing them and will update the docs. In the meanwhile, things are a bit broken and I'm sorry. Should be ok later today :see_no_evil:

eriese:
No need to apologize! I believe this is properly merged and if you let me
know when it's all working again I can double check that

joostdecock:
We're back in business 😊

New is `yarn tips`. Not related to this but feedback welcome

eriese:
I'm gonna need to get a Vercel account, aren't I?

joostdecock:
@eriese If you DM me your email address I can send you an invite for the Vercel team.

That being said, lab is broken right now because of the changes I did recently. I'm still putting it back on the rails.
In other words: It's not you, it's me :joy:

joostdecock:
Please remind me @eriese where are we with this? Can this be merged?

eriese:
Good to go!

PR #2251, opened 2022-06-12
New pattern: hi

woutervdub:


joostdecock:
Tx Wouter, very excited about this, but I'm going to postpone this merge until I'm back from my trip since I'm in the middle of some stuff right now.

But I'll take care of it.

woutervdub:
I totally expected you to not do anything on this until you return. But since I need to put a stake in the sand at some point, now seemed the right time. And it allowed me to blame any delays on you, so there's that ;-)

PR #2262, opened 2022-06-15
Proposal(hi): Make Hi faster (at generating teeth)

joostdecock:
This pattern uses a method to generate teeth which depends on a lot
of Path.shiftAlong() invocations.
This causes the pattern to be a bit sluggish, with a render taking
around 160ms (which is too slow to feel instantaneous for the user) and
the unit tests timing out because some individual tests take longer than
2 seconds. When extending the timeout, running all tests clocks in just
 over 14 seconds which is slow.

This PR replaces this method with an optimized one that is a lot faster.
(all tests now complete in 2.11 seconds).

To do this,it does some things differnet, and cuts some corners:
- It does not use the Path object, but rather the low-level Bezier
  object directly
- It generates teeth for only half the mouth and will mirror them to the
  other side
- It uses a symmetric cubic Bezier curve so we don't need to walk the
  path in steps that are determined by distance, but can just walk it at
  regular intervals of the t value. As it happens, that can be done in
  one pass by using Bezier.getLUT()

The corners I cut:

- Because of the symmetric Bezier the path is ever so
slightly different, but the difference is minimal.provides ai
- Since we generate half of the mouth/teeth and then mirror, we need and
  even number of teeth, so I've changed the upper teeth from 15 to 14


woutervdub:
The shape of the curve is insignificant IMHO. Mirroring, and therefore requiring an even amount of teeth is also not an issue. And allows us to say that it is not a complete copy. 

Why did the size of the small teeth (extreme sides of the mouth) change?

joostdecock:
@woutervdub I'm going to merge this now because I have another big set of changes and letting things diverge will be a nightmare later.

However, we can continue to fine-tune this prior to publishing it. The variation of tooth size for example is something we can review.

Issue #2281, opened 2022-06-22
useMemo() breaks gists updates in measurements input


joostdecock:
**What seems to be the problem?**

The way `useMemo()` is used breaks gists updates:

https://github.com/freesewing/freesewing/blob/78d22259d922d19dbce8be6c1c8b2ea36550cde4/sites/shared/components/workbench/inputs/measurement.js#L56

**Additional context**

Only one measurement will/can be used per page load.
What I mean is:

- Go to measurments
- Add a new measurement => works fine
- Add another measurement => replaces the previous measurement in the gist
- Navigate away (to draft or whatever)
- Add a new measurement => works fine

I'm going to comment out the use of `useMemo()` for now as that solves the issue.
I defer to @eriese for a more permanent fix since it's her baby so I think she'll do better than me.

joostdecock:
I have (temporary) added debug info to the lab to make this easier to troubleshoot:

![Screenshot from 2022-06-22 15-00-53](https://user-images.githubusercontent.com/1708494/175035337-5b5a11b0-8bd7-41a6-bd29-ab24cf6b89f9.png)

eriese:
@joostdecock is there something in particular that had you identify `useMemo` as the culprit? I have an alternate theory, but I want to have all the facts

woutervdub:
Is there a way to 'clear' the lab? Going to `noble` errors out before I can get to `measurements`. It works fine for other designs I hadn't visited before.

eriese:
> Is there a way to 'clear' the lab? Going to `noble` errors out before I can get to `measurements`. It works fine for other designs I hadn't visited before.

The only thing I can think of is to go into your localstorage and change the `_state` value of `fs_noble_gist` to `{view: 'measurements'}`

eriese:
> How do I get to localstorage?

in your javascript console (in the browser), enter
```
let b = JSON.parse(localStorage.getItem('fs_noble_gist'))
b._state={view: 'measurements'}
localStorage.setItem('fs_noble_gist', b)
```

then refresh the page

PR #2284, opened 2022-06-22
fix measurement gist issues with better state management  #2281

eriese:
fix for #2281 This should be working now, but please double check me?

`useMemo` wasn't the culprit, `gist` was: because it's a deeply nested object, react was struggling to track its internal changes and propagate them to everywhere. I think I've come up with a tidy solution.

I also fixed the safety check that keeps you from going to draft without all the measurements. I want to do some UX fixes around that, but I figure that's a separate PR so we can have things working properly asap

### DIFF
--- a/sites/shared/components/workbench/preload.js
+++ b/sites/shared/components/workbench/preload.js
@@ -14,12 +14,15 @@ const preload = {
 
     if (result.data.files['pattern.yaml'].content) {
       let g = yaml.load(result.data.files['pattern.yaml'].content)
-      if (g.design !== pattern.config.name) return [
+
+      if (g.design !== undefined && g.design !== pattern.config.name) return [
         false, `You tried loading a configuration for ${g.design} into a ${pattern.config.name} development environment`
       ]
 
       return g
     }
+
+    // TODO notify people of these errors
     else return [false, 'This gist does not seem to be a valid pattern configuration']
   }
 }


eriese:
Did the UX fixes I wanted. Open to opinions on whether they are good. For example, it now defaults to having 'Enter Measurements by hand' open, which maybe there was a reason it wasn't.

joostdecock:
There's a few issue that need to be addressed stil:

- Clearing the gist no longer works (I did not check why, but it probably restored the default gist which you removed in favor of the hook, so we'll need to have another look at that
- All the inputs are in error state by default (which I find to be rather *scary* for the user. Arguably, ok state is also misleading when there's no value. So perhaps we should take a clue from the degree input which does have this sort of greyed-out state when there's no value
- I wonder if it would make sense to move methods like `updateGist` and `unsetGist` into the hook file. Just for logically grouping code together I mean. Perhaps as named exports?

That being said, I'm going to merge this because the inability to use measurements right now is blocking people, and this nothing above is a show-stopper.

On a personal note, I love your decisive approach and attention to detail. 5/5 would merge again :heart:

joostdecock:
Forgive me, but I can't stop thinking about this comment:

> useMemo wasn't the culprit, gist was: because it's a deeply nested object, react was struggling to track its internal changes and propagate them to everywhere

All we ever do in the frontend is manipulate the gist. All options or settings or whatnot, are merely an update of the gist, and propagation has never been a problem. It was also never a problem to propagate the measurements before we started adding support for imperial.

Now we've added a bunch of complexity and broken other things in the process. Like `setGist()` (which was a simple state update) now no longer works. 

I've always liked the simplicity of [gist, setGist] (with the added updateGist method) that we can pass down.
So I guess my question is: How confident are you that a simple state update is not sufficient?

eriese:
> So I guess my question is: How confident are you that a simple state update is not sufficient?

A very fair question! 

On reflection, I think that what actually broke everything was my `useLocalStorage` hook changes. There was no trouble with the gist in Freesewing itself, but once I tried to use it as a library in an app that wrapped it so I could customize some things I was getting lots of SSR issues. 

This is, in all honesty, my first React project, so I'm learning as I go, but in my research I figured out that the reason for the SSR issues was because checking and populating from `localStorage` was happening outside of the React lifecycle, so then React freaked out about there being a mismatch from what it believed was the current state and what it saw to be the actual current state. 

So I rewrote that hook to be more within React's way of handling stuff, which eliminated my SSR issues but then made all the other places that state was being handled in a not-quite-the-React-way way have a problem, i.e. basically all callbacks were using stale values. I'm not really sure why it worked before, but once I'd broken it, I started trying to fix things using `useMemo` to make sure things were updated when React knew there had been a change. That was doing ok with simpler state values, but it couldn't work for `gist` because... I don't really know why. What I do know is that as I was trying to figure out this measurement thing, I saw that all the functions that are sprinkled throughout the library that take `gist` and copy it and mutate part of it and send that back up the chain were being instantiated with a frozen `gist` value that wasn't being updated, so even though the `state` reflected changes to the `gist`, any new change from elsewhere was being made on the original `gist`, thus erasing all other changes. 

What's seeming painful about this switchover is that I admittedly did not check to see where gist was being used outside of the places I personally saw broken (so there goes my attention to detail), so anything I don't use in the platform is not fixed. It's sounding like this is maybe a somewhat longer adventure and maybe the solution is to roll back all of my various changes (but keep the `.gitpod.yml`) till I have time to clean everything up properly. Once I do, I hope you'll feel that I've actually made things simpler rather than more complex.

> There's a few issue that need to be addressed stil:
> 
> * Clearing the gist no longer works (I did not check why, but it probably restored the default gist which you removed in favor of the hook, so we'll need to have another look at that

probably I didn't implement it in the place where it's used... that's on me

> * All the inputs are in error state by default (which I find to be rather _scary_ for the user. Arguably, ok state is also misleading when there's no value. So perhaps we should take a clue from the degree input which does have this sort of greyed-out state when there's no value

agreed

> * I wonder if it would make sense to move methods like `updateGist` and `unsetGist` into the hook file. Just for logically grouping code together I mean. Perhaps as named exports?

yes, love it

eriese:
>  I wonder if it would make sense to move methods like `updateGist` and `unsetGist` into the hook file. Just for logically grouping code together I mean. Perhaps as named exports?

one thing to note: 
@joostdecock in my new PR #2296 I moved `unsetGist`, but I left `updateGist` because it uses logic specific to the workbench use case. I don't think these can be named exports, though, because I think they need to be instantiated after `setGist`. I could be wrong about that, but I think what we can consider doing instead (to keep it all in the React way of doing things but still avoid the cumbersome passing of functions down the tree) is making a [context provider](https://reactjs.org/docs/context.html) for the gist, so that child components can pick it up as needed without it having to be passed through parent components that don't use it.

joostdecock:
@eriese I much prefer the simplicity of React's state management over rolling our own state management (aka `useReducer`).
When I think about long-term maintenance the changes we put in place do (at least for me) increase the mental overhead.

But, I also understand that prop-driling is not ideal, and so we can perhaps land on `useContext` as a middle-ground.

That being said, I also would like to return to the root cause for all these changes: The SSR issue.
What is it that you are trying to do? What kind of errors are you having? And what do you mean exactly when you say *using FreeSewing as a library*.

Because rewriting prop-drilling as useContext is better, sure. But it's also refactoring for the refactoring.
The original *gist is too complicated for React to handle in state* is, based on my experience, unlikely to be as simple as that.
So I'd like to understand what lead us down this rabbit hole so that we're addressing the underlying issue.

I can merge #2296 (and I will) but I'd like to return to the drawing board and say: What is it that you are trying to do didn't work?

eriese:
> I much prefer the simplicity of React's state management over rolling our own state management (aka `useReducer`).
> When I think about long-term maintenance the changes we put in place do (at least for me) increase the mental overhead.

hmm... I guess I think of `useReducer` as part of React's state management rather than a roll your own. Would it help reduce the mental overhead if I made `setGist` work under the hood the way it used to?

eriese:
there's also possibly one where `useReducer` is more than is needed for this use case, and that I can switch it back to `useState` or something simpler feeling... I will play with that

eriese:
> there's also possibly one where `useReducer` is more than is needed for this use case, and that I can switch it back to `useState` or something simpler feeling... I will play with that

yes, I played with it, and `useReducer` was overkill, so hopefully I've found a nice balance of not adding ongoing mental overhead and fixing what I broke, and we can leave the `contextProvider` stuff for a refactor down the line

PR #2296, opened 2022-06-23
all gist mutation now uses the new functions

eriese:
continuing, but hopefully ending the #2281 saga. I searched for all mentions of gist in the `sites` folder and made sure everyone was using the new methods correctly. The only things that were having trouble still were ClearThing and seam allowance, and they are now working. I tried to click around everywhere in the lab to make sure it all looked good, and as far as I could tell, it did.

### DIFF
--- a/sites/shared/components/workbench/draft/part/index.js
+++ b/sites/shared/components/workbench/draft/part/index.js
@@ -112,7 +112,7 @@ const XrayPart = props => {
 }
 
 const Part = props => {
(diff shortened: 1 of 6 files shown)


joostdecock:
Merging this, but there's an ongoing discussion about this in #2284

eriese:
@joostdecock take a look and see if you like this implementation any better? I think the main thing I want to do here is just keep all copying of the gist in one place where we know we're getting the latest.

PR #2301, opened 2022-06-25
Change hem allowance on Wahid to standard SA

arigrayzel:
Currently, when seam allowance is drawn on the patterns, a triple allowance is given on the bottom of the Wahid pieces, as they are labeled "hem allowance". However since Wahid has a lining rather than a rolled hem this is unnecessary and potentially confusing.

### DIFF
--- a/designs/wahid/src/back.js
+++ b/designs/wahid/src/back.js
@@ -111,7 +111,7 @@ export default (part) => {
     if (sa) {
       paths.sa = paths.saBase
         .offset(sa)
-        .join(paths.hemBase.offset(sa * 1))
+        .join(paths.hemBase.offset(sa))
         .close()
         .attr('class', 'fabric sa')
     }
--- a/designs/wahid/src/back.js
+++ b/designs/wahid/src/back.js
@@ -111,7 +111,7 @@ export default (part) => {
     if (sa) {
       paths.sa = paths.saBase
         .offset(sa)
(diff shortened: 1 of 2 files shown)


designs/wahid/src/back.js
```diff
@@ -111,7 +111,7 @@ export default (part) => {
     if (sa) {
       paths.sa = paths.saBase
         .offset(sa)
-        .join(paths.hemBase.offset(sa * 3))
+        .join(paths.hemBase.offset(sa * 1))
```
joostdecock, line 114: You can just get rid of the entire `* 1` so it's just `.offset(sa)`

designs/wahid/src/front.js
```diff
@@ -285,7 +285,7 @@ export default (part) => {
     if (sa) {
       paths.sa = paths.saBase
         .offset(sa)
-        .join(paths.hemBase.offset(sa * 3))
+        .join(paths.hemBase.offset(sa * 1))
```
joostdecock, line 288: You can just get rid of the entire `* 1` so it's just `.offset(sa)`

joostdecock (changes requested):

joostdecock (approved):
@all-contributors please add @arigrayzel for code

joostdecock:
@all-contributors please add @arigrayzel for code

PR #2312, opened 2022-06-28
add(optiongroups.yaml) Added Type and Size for Noble & Bob

bobgeorgethe3rd:
Adding type for the new option group created by Noble.

Edit: Noticed Bob had the new option group size as well so added to this.


joostdecock:
Nice work buddy :+1:

PR #2314, opened 2022-06-28
Incorrect Title Format Hi Size Design Option page

bobgeorgethe3rd:
Name of pattern is not present for Design Option titles so have removed.


joostdecock:
Good catch. Pretty sure this was me :see_no_evil:

PR #2316, opened 2022-06-28
Layout Thing Improvements

eriese:
I simplified the existing `sites/shared/workbench/layout/draft.js` by:
- breaking it up into multiple components
- using `transform-origin` to eliminate the need to a lot of the transform math to re-center the parts after flipping and rotating
- using [DOMPointReadOnly](https://developer.mozilla.org/en-US/docs/Web/API/DOMPointReadOnly) with `matrixTransform` to handle conversion from DOM coordinates to SVG coordinates.
- using `getBoundingClientRect` to avoid having to calculate top left and bottom right on rotated parts, and using the above coordinate conversion method to change the returned DOM coordinates into SVG coordinates.
- making a centralized method (in `core/part`) for generating transforms, so that they are done the same way on init as they are after user interaction.
- adding a reset button to the layout interface, which clears the entire layout.

I should admit that I didn't check whether the methods I'm using are more computationally expensive than what I replaced (my guess is that some are), but on my local I didn't experience any performance problems, and I'm hopeful that it will feel like a lot less mental overhead to maintain. 

I also believe that this has solved all known issues, so please let me know if any new ones have arisen.


joostdecock:
Oh @eriese is there anything you can't do :heart:

PR #2352, opened 2022-07-02
A11y improvements

eriese:
- refactored all the "picker" components to now all use the same base `Picker` component, which uses HeadlessUI's menu components. Any visual differences are because the contrast was low and I changed it.
- changed 'neutral-focus' color in dark theme so that focus outline shows better

Needs some cleanup, but I want confirmation that this will be acceptable before continuing

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,6 +138,7 @@
   ],
   "version": "0.0.0",
   "dependencies": {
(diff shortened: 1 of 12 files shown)


joostdecock (approved):
Let's do it

PR #2357, opened 2022-07-05
Unice add markers for elastic by injecting from ursula (pr version)

anna-puk:
add markers, copying as much as possible from ursula


joostdecock:
Closing in favor of #2401

PR #2398, opened 2022-07-12
Fix Bug: some items render in the wrong places in the print layout

eriese:
The partname wasn't getting set properly, which I guess messed up the ids so nothing had unique ids, which messed up the references. this fixes that.


eriese:
Wait, don't merge this, it hasn't fixed everything

eriese:
ok, now it should be good to go

PR #2420, opened 2022-07-16
Menu fixes

eriese:
- add missing design names to mobile nav
- remove pattern picker from mobile nav because the patterns are listed
- add key to pattern picker spans to get rid of no key error
- don't put `dropdown-end` class on pickers on left hand side of header to make sure the whole width of the picker is visible on narrower screens

### DIFF
--- a/sites/lab/components/layouts/lab.js
+++ b/sites/lab/components/layouts/lab.js
@@ -10,7 +10,6 @@ export const BeforeNav = ({ app }) => (
       <LocalePicker app={app} />
     </div>
     <div className="md:hidden flex flex-row flex-wrap sm:flex-nowrap gap-2 mb-2">
-      <PatternPicker app={app} />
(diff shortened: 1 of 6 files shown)


sites/lab/components/layouts/lab.js
```diff
@@ -10,7 +10,6 @@ export const BeforeNav = ({ app }) => (
       <LocalePicker app={app} />
     </div>
     <div className="md:hidden flex flex-row flex-wrap sm:flex-nowrap gap-2 mb-2">
-      <PatternPicker app={app} />
```
joostdecock, line 13: Please also remove the import of this component since it's no longer used

joostdecock (changes requested):
Could you just remove the unused import.
Apart from that, LGTM

joostdecock:

joostdecock (approved):

PR #2421, opened 2022-07-16
Error boundary

eriese:
- add gist history stack
- add errorBoundary component to wrap the lab view
- add default inner component for error boundary using the draft error component as a template
- add stack trace to events error readout
- actually use svg style attribute so plugins can manipulate it
- various additional null checks and try/catches to prevent errors


joostdecock:

joostdecock (approved):

PR #2422, opened 2022-07-16
Preliminary Cutting Layout

eriese:
Currently only fully configured for Bella, but shouldn't break for other designs

The gist now holds each type of layout separately so that the user doesn't have to move things back and forth when switching between types

At this current stage the cutting layout operates much like the print layout with the following additional features:
- when a part has a `cutonfold` macro it will make the part mirror itself along the indicated line so that the full unfolded shape of the part is shown 
- it uses the `cutList` object in the design config, which is structured:
```
{
  [partName]: {
    cut: 2,
    isPair: true
  }
}
```
Parts not present in the cutList default to 'Cut 1', otherwise it will render the number in `cut`. If `isPair` it will flip every other rendered copy of the part by default

Additional changes:
- implemented rotation snapping to 15 degree increments with the shift key
- added rotate90 buttons to part manipulation buttonset
- buttons maintain consistent orientation rather than rotating and flipping with the parts
- new, clearer icons (let me know what you think!) for the part manipulation buttons 
- hovertext for the part manipulation buttons


joostdecock:

That being said, I think we should discuss this way we're going to provide the cutting config.
Providing it in the config is a bit of a *naive* approach that works fine for Bella because it's such a simple pattern. But if we want to support all patterns, we can't have a static config, and instead the pattern should export a method that we can call and pass the user settings, because depending on it it may or may not have certain parts.

I'm all for this, we need to do it right, rather than quick :shrug:

eriese:
> That being said, I think we should discuss this way we're going to provide the cutting config. Providing it in the config is a bit of a _naive_ approach that works fine for Bella because it's such a simple pattern. But if we want to support all patterns, we can't have a static config, and instead the pattern should export a method that we can call and pass the user settings, because depending on it it may or may not have certain parts.

you're absolutely right. the mirrorOnFold plugin (which is poorly named, oops) does accept a method for cutList, but then as I started also using it in other places, I forgot I was doing that, so it's not implemented everywhere. I'll centralize it

eriese:
@joostdecock There's a utility method for generating the cutlist based on the settings. see what you think

eriese:
yes, this branch is super stale

PR #2423, opened 2022-07-16
Prevent attempts at drafting until after gistReady

eriese:
Patterns with options that have a `toAbs` method break the build because the draft view tries to draft them without waiting for gistReady, so they don't have measurements to pass to `toAbs`. They aren't redirected to measurements because that check does wait for `gistReady`. This prevents attempts to draft until the gist is ready


joostdecock:

joostdecock (approved):

PR #2445, opened 2022-07-20
Fix: Actually Ensure locale directories are made in i18n prebuild

eriese:
Well, I wasn't paying enough attention to how promises work, so the race condition wasn't solved. Now we're awaiting the directory creation to make sure that the files really aren't made without the directory


eriese:
Closing because these changes are also in #2449 and would make for a painful merge

PR #2449, opened 2022-07-20
Fix: Prebuild scripts for cibuild_step scripts

eriese:
- reconfigure script now checks if packages have a prebuild script when writing cibuild steps, and adds that as `precibuild_stepx` to the package.json of that package
- `sites/shared/prebuild/i18n.mjs` now calls the build method in `packages/i18n/src/prebuild.mjs` to reduce maintenance

### DIFF
--- a/config/dependencies.yaml
+++ b/config/dependencies.yaml
@@ -192,7 +192,7 @@ react-pattern:
     '@freesewing/i18n': *freesewing
 rehype-jargon:
   _:
-    'unist-util-visit': '^2.0.3'
+    'unist-util-visit': '^4.1.0'
     'hast-util-from-html': '^1.0.0'
 remark-jargon:
   _:
--- a/packages/i18n/src/next.mjs
+++ b/packages/i18n/src/next.mjs
@@ -7,17 +7,20 @@ import enNamespaces from "./next/en/index.mjs"
 import esNamespaces from "./next/es/index.mjs"
 import frNamespaces from "./next/fr/index.mjs"
 import nlNamespaces from "./next/nl/index.mjs"
+import ukNamespaces from "./next/uk/index.mjs"
 
 export const de = deNamespaces
 export const en = enNamespaces
 export const es = esNamespaces
 export const fr = frNamespaces
 export const nl = nlNamespaces
+export const uk = ukNamespaces
 
 export const languages = {
   de: "Deutsch",
   en: "English",
   es: "Español",
   fr: "Français",
-  nl: "Nederlands"
+  nl: "Nederlands",
+  uk: "undefined"
 }
--- a/packages/i18n/src/prebuild.mjs
+++ b/packages/i18n/src/prebuild.mjs
@@ -108,10 +108,10 @@ const getNamespacesFromFileList = async (files, locales, only=false) => {
     let file = files[i]
 
     let loc = localeFromFileName(file);
-    if (locales.indexOf(loc) === -1) return
+    if (locales.indexOf(loc) === -1) continue
 
     let namespace = namespaceFromFile(file);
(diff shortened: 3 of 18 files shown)


packages/i18n/src/prebuild.mjs
```diff
@@ -34,14 +34,19 @@ const getTranslationFileList = async () => {
...
+/** extracts the locale from the file name */
+const localeFromFileName = (fileName) => {
+  const split1 = fileName.split(`${path.sep}locales${path.sep}`).pop()
+  return split1.split(`${path.sep}`).shift()
+}
+
 /*
  * Figures out the list of locales from the list of files
  * (by checking how many version of aaron.yaml exist)
  */
```
joostdecock, line 46: Should we perhaps keep a list of (supported) locales in the repo's config folder? We might need this in more place than one 🤔

joostdecock (approved):

packages/i18n/src/prebuild.mjs
eriese: wow, I was calling `return` instead of `continue` when checking locales, so that's why it was breaking with a filter

config/dependencies.yaml
```diff
@@ -192,7 +192,7 @@ react-pattern:
     '@freesewing/i18n': *freesewing
 rehype-jargon:
   _:
-    'unist-util-visit': '^2.0.3'
+    'unist-util-visit': '^4.1.0'
```
eriese, line 195: remark-jargon had at some point been bumped to `unist-util-visit@^4.1.0`, so I figured that they should both be on the same version, but tell me if I'm wrong?

packages/i18n/src/next.mjs
```diff
@@ -7,17 +7,20 @@ import enNamespaces from "./next/en/index.mjs"
...
+export const uk = ukNamespaces
 
 export const languages = {
   de: "Deutsch",
   en: "English",
   es: "Español",
   fr: "Français",
-  nl: "Nederlands"
+  nl: "Nederlands",
+  uk: "undefined"
```
eriese, line 25: I've been avoiding committing all the Uk files, but probably that needs to happen

packages/remark-jargon/package.json
```diff
@@ -35,7 +35,7 @@
     "remark-html": "^13.0.1"
   },
   "dependencies": {
-    "unist-util-visit": "^2.0.3"
+    "unist-util-visit": "^4.1.0"
```
eriese, line 38: this happened because I ran `yarn reconfigure` and this was the version set in the dependencies file

packages/remark-jargon/src/index.js
```diff
@@ -1,4 +1,4 @@
-import visit from 'unist-util-visit'
+import {visit} from 'unist-util-visit'
```
eriese, line 1: this broke after I ran `yarn reconfigure` because it's now using a new version that only has named exports

packages/i18n/src/prebuild.mjs
joostdecock: > My question is what the use case is for not simply using all the locales that we have?

Well people are currently working on the Ukranian translation. It is 3% done, and we want it to go up to a higher number, and at least have the signup/login flow and so on translated.
When that's done, we'll probably do some sort of *we haz ukranian now* announcement. But until that time, it should be kept out of the build.

For what it's worth, adding a new language is an exceedingly rare event.

config/dependencies.yaml
joostdecock: Not sure why it's different from remark-jargon. Or rather I can't remember.

joostdecock (approved):

PR #2467, opened 2022-07-23
chore(core): Working on test coverage

joostdecock:


BenJamesBen (commented):
While not an official code review _per se_, i did look over the code changes.
I note that some of the added tests are for undocumented functions, for example path.withRaise(), point.withRaise(), utils.isCoord(), utils.sampleStyle(). But, perhaps these are intended to be private functions.

joostdecock:
> perhaps these are intended to be private functions.

@BenJamesBen I think you should see them as such. 
There's a way to make class functions private in JS, but that does not apply to prototype inheritance.

So you should see this more as: If you know what you're doing, you do you.

joostdecock:
96.29% => merging

PR #2469, opened 2022-07-24
Fix: copy locale files to the sites in prebuild

eriese:
I got ahead of myself refactored out the part where the sites/shared prebuilder actually writes the locale json files, so this puts it back in


packages/i18n/scripts/prebuilder.mjs
```diff
@@ -1,4 +1,7 @@
 import {build} from '../src/prebuild.mjs'
 
+// use a deny-list to keep locales that aren't ready out of the build
+export const denyList = ['uk']
```
eriese, line 4: I get what you were saying about needing a list of supported languages. My feeling was that a deny-list was better, but I really think we could go either way on that one. And we probably want to keep it someplace more prominent... let me know your thoughts?

joostdecock: LGTM

joostdecock (approved):

PR #2490, opened 2022-07-27
notch/bnotch snippets come from plugin-notches, not plugin-theme

BenJamesBen:
As I understand it, at one time notch and bnotch were in plugin-theme, but they were subsequently moved to plugin-notches. (A minor documentation issue, but it did cause me some grief spending time looking through the plugin-theme source code, trying to find where notch was defined.)


joostdecock (approved):
You're doing sone great work here @BenJamesBen much appreciated 🙏

PR #2497, opened 2022-07-28
Feature: Export print layout to PDF

eriese:
All the essential bits are implemented and ready for feedback:
- [x] user can export the print layout as they've set it on the print layout page by clicking 'export' from the print layout page or  'pdf' from the export page
- [x] user can export their print layout to other formats using the other 'export to PDF' buttons on the export page
- [x] user can customize the amount of page overlap
- [x] first page upper left corner has a ruler to measure print accuracy in the pattern's units
- [x] cover page shows the entire print layout as a guide
- [ ] cover page is optional
- [ ] cover page is styled  

This uses the not-yet-implemented multiple layouts functionality, so in order to not duplicate effort, this was branched off of my WIP cut layout branch. I rolled back everything that was clearly not needed in this branch, but there's still a bit of new code that's not relevant to this feature.

Going to work on tests and styling and everything, but in the mean time I wanted to get this up for people to test


BenJamesBen:
(I'm not sure whether feedback is better located here or on Discord.)
I was able to cause an error by:
1. Using the Bob pattern, on the LayoutThing: ForPrinting page.
2. Set Paperless to 'yes'.
3. Press Export button.

```
Unhandled Runtime Error
TypeError: Cannot read properties of null (reading 'width')
Source
../shared/components/workbench/exporting/exporter.js (42:27) @ new Exporter

  40 | 	  this.svg = divElem.firstElementChild;
  41 | 
> 42 | 		this.svgWidth = this.svg.width.baseVal.valueInSpecifiedUnits
     | 		                        ^
  43 | 		this.svgHeight = this.svg.height.baseVal.valueInSpecifiedUnits
  44 | 
  45 | 	  this.wPages = Math.ceil(this.svgWidth/pageWidthInPx)
Call Stack
handleExport
../shared/components/workbench/exporting/index.js (70:9)
exportIt
../shared/components/workbench/layout/print/index.js (37:17)
```

BenJamesBen:
I was able to get the ExportThing "YAML" button to cause an error twice, though I am no longer able to reproduce the problem. I am not sure what sequence of steps resulted in the error. Using the standard Bob, I randomly pressed the PDF, SVG, JSON, and YAML buttons a number of times to test them out. Also, sometimes the YAML button does nothing. When I get into the bug state, ClearThing fixes the issue.
```
Unhandled Runtime Error
YAMLException: unacceptable kind of an object to dump [object DOMPoint]

Source
../shared/components/workbench/exporting/index.js (80:25) @ exportYaml

  78 | }
  79 | const exportYaml = gist => {
> 80 |   const blob = new Blob([yaml.dump(gist)], {
     |                         ^
  81 |     type: 'application/x-yaml;charset=utf-8'
  82 |   })
  83 |   fileSaver.saveAs(blob, `freesewing-${gist.design || 'gist'}.yaml`)
Call Stack
handleExport
../shared/components/workbench/exporting/index.js (31:32)
onClick
../shared/components/workbench/exporting/index.js (128:31)
```

I think the steps to reproduce the "YAML button does nothing" behavior is:
1. With Bob set back to default using ClearThing.
2. Go to LayoutThing: ForPrinting
3. Move the piece slightly.
4. In ExportThing, press the YAML button.

Hmm... I was able to reproduce the Runtime Error just now by following those steps.

BenJamesBen:
The ExportThing page "Export for Printing" buttons do not actually print to the page sizes printed on the buttons. Instead, they print to the page size selected in the LayoutThing: ForPrinting page.
1. Using Lucy, on the LayoutThing: ForPrinting page.
2. Select the Letter page size.
3. Press the Export button. The pattern is printed on 4 letter-size pages, as expected.
4. Go to the ExportThing page and press the "A0 PDF" button.
5. The pattern is incorrectly printed on 4 letter-size pages, instead of 1 A0-sized page.
6. (By using ClearThing and repeating the steps instead selecting "A0" on LayoutThing: ForPrinting, you can confirm that the ExportThing page is using the size that was selected in LayoutThing:ForPrinting.)

joostdecock:
> I'm not sure whether feedback is better located here or on Discord

@BenJamesBen I feel this would get snowed under quickly on Discord and it might also be a too abstract for most people there.
Also note that there's a GitHub channel where these omments will also appear.
TL;DR: hete is fine

@eriese with cutList fully implemented in `part` and `pattern.getCutList()` available post-draft, I think some of the methods in core-utils are no longer required, correct?

eriese:
> * We should not have a `utils.js` and `core-utils.js` file in a package named `core`. That's a very confusing situation
Totally see your point here. I separated `core-utils` from `utils` because everything in `utils` gets attached to `Part`, and I wanted utility methods that were used in multiple places by the core but wouldn't make sense to expose as part drafting utilities. I still think it makes sense for those methods to stay separate from the other utilities, but given other things we're discussing, this file may be moot anyway.

> * I don't see why the transform-generating method is not attached to the `part` prototype
`Pattern.prototype.getRenderProps`, the method we call to get all the info to pass to the draft components in our various layout UIs, returns `Objects` with part data rather than `Part` objects, so when the front end needs to update the layout based on user interaction, it needs to generate transforms using only the POJO of part data it has.

> * Part transforms should not be passed as a bunch of that need to come in a specific order, but rather as a layout object as documented here: https://freesewing.dev/reference/api/settings/layout#layout-is-a-layout-object
Do you mean in `generatePartTransform`? Happy to change it to an object. Much better, really.

> * I am not convinced we should handle multiple layouts in core, which leads to a bunch of further questions
It sure does! Perhaps instead of looking for all the layouts in the settings, we just pass them directly to the render functions? It'll have to do basically all the same work, but it might look slightly cleaner...

**Also to note:** A lot of the changes in `pattern.js`  are with two goals in mind, both of which are related to layout handling, so maybe it all ends up moved out
1) it needed to be able to accept a layout type
2) there was a pre-existing issue in the layout code where if you changed a setting and a piece got added, it wasn't packed and you wouldn't see it unless you reset to the auto layout, and I think we don't want that. So where it used to skip packing if it saw that a layout was defined, it now packs if anything has changed in the layout.

eriese:
Some proposals:
1) instead of passing `layoutType` all over the place in `Pattern`, it's the view's job to set `gist.layout` to the correct value from `gist.layouts`. This way, all the layouts are saved individually, but the pattern looks in the same place it's always looked.
2) I think in order to distinguish between layouts that shouldn't be recalculated and layouts that do, we will still need to pass something from the UI, which you can see in proposal 3
3) `Pattern.prototype.pack` is taking on too much and is getting messy. I think we need to break it out into 3 pieces:

```
pack(repack = false) {
  if (repack === false && layout !== undefined) return the layout
  if (layout !== undefined) pack new parts around the existing layout
  else pack from scratch
  return the pack
}

generateAutoLayout(repack = false) {
  set the pattern's size from the pack step result
  If (repack) put that data into the autoLayout object
}


generatePartTransforms() {
  clear part transforms and generate new ones
}
```

eriese:
Not sure I agree that either additional needs are hard to fit in:
> * **Runtime part creation/duplication**: Add parts to the pattern at run-time

Technically this one is already supported with the pages "part" that gets added by the print layout plugin. We could consider putting part duplication into a plugin as well. That plugin uses the postLayout hook, but the cut layout one wouldn't necessarily have to

> * **Incremental packing**: Have a explicit layout for some parts and then fit the other parts around that

I don't think that a lot of additional code is needed to implement this within the current packing code, and it's not a breaking change. I do kind of think that packing could be extracted from the pattern, but I'm not sure what it accomplishes besides trimming down a too-long file

> Alternative approaches:
> 
> * Render each part individually. Handle layout in the DOM rather than in SVG
> * Add methods to the SVG object to add/duplicate parts (this would not be a solution for React-based rendering though)

I need more context on why either of these would be preferable

> Perhaps this is a good occasion to re-visit the way parts are added to a pattern, which is something I'm not really happy with today anyway 🤔

I would love to and I have MANY ideas, but that would definitely bring some breaking changes, so in the mean time, I think the simplest thing is to add incremental packing to the pack method and handle whatever else outside of core. That's at least all I need to get this PR in shape. Cutting layout will want a way to duplicate parts, but there's enough that needs overhaul over there that we can get to it when we come to it.

joostdecock:
> Not sure I agree that either additional needs are hard to fit in:

What I meant is that it's hard to fit in this functionality with the current API surface in core.
I would prefer to extend/adapt the API to support these things, rather than get overly focused on one specific use-case.

> I would love to and I have MANY ideas, but that would definitely bring some breaking changes, so in the mean time, I think the simplest thing is to ...

I think we should strive to make it simple for other people. Not do what's simple for us ;)
So let's hear those MANY ideas :)

As for *for now let's just quickly*, I appreciate you have your own goal/timing. But for FreeSewing, it doesn't really make a lot of sense to cram this in now when realistically, this feature won't be available on freesewing.org until sometime in 2023.

So I prefer to aim for _doing it right_, rather than doing it fast.

eriese:
> So I prefer to aim for _doing it right_, rather than doing it fast.

Of course. I was thinking more in terms of progressive enhancement, and also not letting this code get so stale that reviving it is a huge pain. I think that the way I made this branch added a lot of complexity that is not actually in the scope of the feature it purports to add, and that the only layout question that actually needs to be answered before we can cleanly implement export to pdf is how we will tell the pattern which layout it's supposed to use for rendering. And I believe we have answered that question: set `layout` to the one that's supposed to be used and change nothing in core. All the other questions are either generally necessary or specific to the cutting layout

So I think what I should do is split this PR into two things that can be accepted at this time that probably won't need major re-working when we have more answers:
1) the ui enhancements for layout tools: new buttons, snapped rotation, page margin slider
2) export to PDF. I'll start from `develop` so it's clearer to me and everyone what I'm changing for this feature specifically, and I think it'll turn out to not need to change core maybe at all

Does that sounds ok?

joostdecock:
@eriese with #2586 merged, can this be closed?

eriese:
yeah, it's definitely stale

PR #2498, opened 2022-07-28
Fix #2487: picker menus work in safari

### LINKED ISSUE #2487: [bug]: [safari] Header dropdown menu buttons not working on freesewing.dev and localhost:8000 lab

Using the Safari web browser, the header dropdown menu buttons do not work on these pages:
- On freesewing.dev, the "Light Theme" button.
- On the localhost:8000 version of the lab, the "Designs", "Next Version", "Light Theme", and "English" buttons. (But interestingly, on lab.freesewing.dev all 4 of those buttons work just fine.)

The buttons change color when hovered over. They shrink slightly upon mousedown and go back to original size upon mouseup. However, no menu items appear.

This was seen with macOS Safari 12.5 and iOS Safari 15.6. On iOS Safari, the same behavior was seen where the buttons work on lab.freesewing.dev but not on freesewing.dev. The problem is not seen when using the macOS Chrome browser. With Chrome all buttons on all sites work fine. I have an ad blocker installed, but the problem also occurs with the ad blocker disabled.

_No response_

No, I am not 😞

### Additional context

_No response_

eriese:
fixes #2487 

Something about the Daisy dropdown styling still being on the menus was messing them up in safari, I guess competing concerns? They work now

### DIFF
--- a/sites/shared/components/picker.js
+++ b/sites/shared/components/picker.js
@@ -5,17 +5,17 @@ import Link from 'next/link'
 /** an accessible dropdown menu for use by picker components */
 export const Picker = ({Icon, className, title, ariaLabel, iconOnly=false, children, end}) => {
 
-	return (<Menu as="div" className={`dropdown w-auto ${end ? 'dropdown-end' : ''}`}>
+	return (<Menu as="div" className={`w-auto relative`}>
 		<Menu.Button className={iconOnly
 			? `btn btn-sm`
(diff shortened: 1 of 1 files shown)


BenJamesBen:
I tested it the fix and confirm that it works. With the fix installed and viewing the localhost:8000 lab using desktop Safari, the header menu buttons now work.

joostdecock (approved):

PR #2499, opened 2022-07-28
Octoplushy

woutervdub:


woutervdub:
New design: Octoplushy

markdown/org/docs/patterns/octoplushy/fabric/en.md
```diff
@@ -0,0 +1,12 @@
...
+---
+## Main and Contrast Body Fabrics (Upper body and legs, belly and underside legs)
+
+This plush toy is intended to have two different coloured fabrics, one for the upper body and top side of the legs, one for the belly and underside of the legs. 
+
+Hi is designed for faux fur fabric, which most plush toys are made from, but you could also use polar fleece, felt, denim, corduroy, or anything moderately heavy. Lightweight or drapey fabrics may not hold Hi's shape well when stuffed. Fabrics with a two-way stretch will work better than wovens, the stretch should be perpendicular to the grainline.  
+
+## Stuffing
+
+Since this is a plush toy, it will need to be stuffed with material. Most of the time you'll want to use polyester fiberfill (polyfill), but if you have a lot of fabric scraps, you can use shredded scrap fabric. However this may be heavier and less soft than fiberfill. It's also worth thinking about whether your Hi may be at risk of getting dirty, or want to go for a swim, as not all stuffing options are washable.
```
bobgeorgethe3rd, line 12:
```suggestion
Octoplushy  is designed for faux fur fabric, which most plush toys are made from, but you could also use polar fleece, felt, denim, corduroy, or anything moderately heavy. Lightweight or drapey fabrics may not hold 
Octoplushy's shape well when stuffed. Fabrics with a two-way stretch will work better than wovens, the stretch should be perpendicular to the grainline.  

## Stuffing

Since this is a plush toy, it will need to be stuffed with material. Most of the time you'll want to use polyester fiberfill (polyfill), but if you have a lot of fabric scraps, you can use shredded scrap fabric. However this may be heavier and less soft than fiberfill. It's also worth thinking about whether your Octoplushy may be at risk of getting dirty, or want to go for a swim, as not all stuffing options are washable.
```

bobgeorgethe3rd: Replacing Hi with Octoplushy from where it has been copied.

joostdecock:
Awesome :) 
Please fix the tests so we can merge this with confidence that it won't break things.

I suspect you might need to update your fork to fix the lab deploy.
Then, address the deepscan issues.

joostdecock:
Closing in favor of #2507

PR #2502, opened 2022-07-30
feat(core): Added new cutting methods. Closes #2470

### LINKED ISSUE #2470: [feature]: Handling cutting info in the core library

We need to add handling of the cutting info in the core library for some of the new features we're working on.

**Config format:**
```js
{
  grain: 90,
  cutOnFold: [p1, p2],
  materials: {
    fabric: {
      cut: 2,
      identical: false
    },
    lining: {
      cut 2
    }
  }
}
```

**Methods:**

- [x] `part.addCut(nr of cuts=2, material='fabric',identical=false)`
- [x] `part.addCut(falsy, material)` => removes
- [x] `part.removeCut(material)` => also removes
- [x] `part.setGrain(angle)` => called by macro('grainline'), which is where the points are created and the angle is calculated. does no work aside from storing the value
- [x] `part.setCutOnFold(p1, p2)` => called by macro('cutOnFold'), which is where the points are created. does no work aside from storing the value 

### Additional context

_No response_

joostdecock:


packages/core/src/part.js
```diff
@@ -357,20 +357,68 @@ Part.prototype.generateTransform = function(transforms) {
   }
 }
 
-/** Chainable way to set the grain property */
-Part.prototype.setGrain = function (grain = 90) {
-  this.attributes.set('data-grain', grain)
+/** Chainable way to add the cut info */
+Part.prototype.addCut = function (cut=2, material='fabric', identical=false) {
```
eriese, line 361: change to `cut=1` I don't think it's safe to assume most pieces are pairs

packages/core/src/part.js
```diff
@@ -357,20 +357,68 @@ Part.prototype.generateTransform = function(transforms) {
...
+  if (cut === false) {
+    if (this.cut.materials[material]) delete this.cut.materials[material]
+    else this.context.raise.error(`Tried to remove a material that is not set`)
```
eriese, line 364: I think this should be a warning

packages/core/src/part.js
```diff
@@ -357,20 +357,68 @@ Part.prototype.generateTransform = function(transforms) {
...
+    return this
+  }
+  if (typeof material !== 'string') {
+    this.context.raise.error(`Tried to set material to a value that is not a string`)
+    return this
+  }
+  if (this.cut.materials[material]) delete this.cut.materials[material]
+  else this.context.raise.error(`Tried to remove a material that is not set`)
 
   return this
```
eriese, line 395: This should internally call `this.setCut(false, material)` rather than doing any of its own business

packages/core/src/part.js
```diff
@@ -357,20 +357,68 @@ Part.prototype.generateTransform = function(transforms) {
...
 
-/** Chainable way to set the grain property */
-Part.prototype.setCut = function (cut = { count: 2, mirror: true, onFold: false }) {
-  this.attributes.set('data-cut', cut)
+/** Chainable way to add the grain info */
+Part.prototype.setGrain = function (grain=false) {
+  if (grain === false) {
+    this.context.raise.warning('Called part.setGrain() without any parameters. Not changing anything')
+    return this
+  }
```
eriese, line 403: I actually think it should remove the grain. I guess a part that doesn't care about grain is pretty rare, but maybe it should be an option?

packages/core/tests/part.test.js
```diff
@@ -273,33 +273,137 @@ it("Should generate the part transforms", () => {
...
+
+it("Should generate an error if material is not a string", () => {
+  let pattern = new freesewing.Pattern();
+  let part = new pattern.Part();
+  part.addCut(3, 4)
+  expect(pattern.events.error.length).to.equal(1)
+  expect(pattern.events.error[0]).to.equal('Tried to set material to a value that is not a string')
+});
+
+it("Should generate an error when removing a material that is not set (through addCut)", () => {
```
eriese, line 303: change to warning

packages/core/tests/part.test.js
```diff
@@ -273,33 +273,137 @@ it("Should generate the part transforms", () => {
...
+
+it("Should remove the part cut through removeCut", () => {
+  let pattern = new freesewing.Pattern();
+  let part = new pattern.Part();
+  part.addCut(4, 'fabric', true)
+  part.removeCut('fabric')
+  expect(typeof part.cut.materials.fabric).to.equal('undefined')
+});
+
+it("Should generate an error when removing a material that is not set (through removeCut)", () => {
```
eriese, line 328: change to warning

packages/core/tests/part.test.js
```diff
@@ -273,33 +273,137 @@ it("Should generate the part transforms", () => {
...
+  let pattern = new freesewing.Pattern();
+  let part = new pattern.Part();
+  part.addCut(4, 'fabric', true)
+  part.removeCut('lining')
+  expect(pattern.events.error.length).to.equal(1)
+  expect(pattern.events.error[0]).to.equal('Tried to remove a material that is not set')
+});
+
+it("Should generate an error when removing a material that is not set (through removeCut)", () => {
```
eriese, line 337: this text doesn't match the test

packages/core/tests/part.test.js
```diff
@@ -273,33 +273,137 @@ it("Should generate the part transforms", () => {
...
+
+it("Should generate a warning when calling removeCut without parameters", () => {
+  let pattern = new freesewing.Pattern();
+  let part = new pattern.Part();
+  part.addCut(4, 'fabric', true)
+  part.removeCut()
+  expect(pattern.events.warning.length).to.equal(1)
+  expect(pattern.events.warning[0]).to.equal('Called part.removeCut() without any parameters. Not removing anything')
+});
+
```
eriese, line 363: this is a duplicate

eriese (changes requested):
Looking really good! Just some little details, I think

packages/core/src/part.js
BenJamesBen: A part that is a circle of fabric? For example, when making circular/round fabric buttons.

packages/core/src/part.js
```diff
@@ -357,20 +357,68 @@ Part.prototype.generateTransform = function(transforms) {
...
+  if (cut === false) {
+    if (this.cut.materials[material]) delete this.cut.materials[material]
+    else this.context.raise.error(`Tried to remove a material that is not set`)
+    return this
+  }
+  if (typeof this.cut.materials[material] === 'undefined') this.cut.materials[material] = {}
+  if (typeof cut !== 'number') {
+    this.context.raise.error(`Tried to set cut to a value that is not a number`)
+    return this
+  }
```
BenJamesBen, line 371: We are permitting customers to set `cut` to negative numbers and floats? I am not sure how `cut` will be used elsewhere in the code.

packages/core/src/part.js
BenJamesBen: Also, not familiar with Javascript or how grain works, but will trying to set a grain of 0 get caught by this? I'm assuming grain is an angle and 0 is a valid value.

packages/core/src/part.js
eriese: Good catch. Cut needs to be positive

packages/core/src/part.js
eriese: In js, `===` is exact equality, so this will only catch false values, not falsy

packages/core/src/part.js
joostdecock: It's a default, and defaults should be sensible. 
My feeling is that cut=2 is by far the most common option.

Quick check for designs in the first three letters of the alphabet seems to confirm that feeling:

| Design | Cut=1 | Cut=2 |
| --------- | -------- | --------- |
| Aaron | 2 | 0 |
| Albert | 3 | 0 |
| Bee | 1 | 2 |
| Benjamin | 1 | 1 |
| Bent | 0 | 4 |
| Bob | 1 | 0 |
| Breanna | 0 | 3 |
| Brian | 0 | 3 |
| Bruce | 1 | 3 |
| Carton | 2 | 14 |
| Carlita | 2 | 15 |
| Cathrin | 0 | 6 |
| Charlie | 2 | 7 |
| Cornelius | 2 | 6 ||
| **Summary** | 17 | 64 |

So leaving 2 as I feel it's the most sensible default.

packages/core/src/part.js
joostdecock: Not sure what to make of *remove the grain* so I've set it to `false`. I think we should always have `grain` as a property, rather than remove it from the object.
Setting it to `false` is explicit, so went with that.

packages/core/tests/part.test.js
joostdecock: Not sure what you mean, but it does. And if it doesn't the test wouldn't pass :thinking:

packages/core/tests/part.test.js
joostdecock: tx

packages/core/src/part.js
joostdecock: I've checked for an int, but not positive. I've also allowed zero. Because some patterns have parts that are not rendered by default. The user can however choose to render them. So I feel it makes sense to set `cut=0` on these, so I've allowed that.

joostdecock:
@BenJamesBen @eriese
Thank you for your input :pray: 
I tried to resolve all issues you guys flagged, thanks a lot. Some exceptions:

- **Set cut to `1` by default**: Didn't do that as `2` is far more common than `1`
- **Remove grain if it's falsy**: Didn't *remove* it from the object, but set it to `false`
- **Test text doesn't match the text**: I think that may have been a mistake. Text does match the text for as far as I (or the test itself) can see
- **Cut should be a positive integer**: I've also allowed zero because this could be used for parts that are not to be cut

joostdecock:
Last minute addition: I've added the ability to remove the `cutOnFold` with `part.setCutOnFold(false)`

packages/core/tests/part.test.js
eriese: What I meant was that either the description of what this test is testing for is incorrect, or this test is a duplicate of the test before it, which has the exact same description. In reading a second time, I lean more towards duplicate, but my first reading I leaned more toward that it might be testing something different

eriese (approved):
Left my one comment about that test, but other than that, I'm all good with your changes, so I'm going to approve because I don't think the name of a test should mean you have to keep waiting for me to look at it again to merge

packages/core/tests/part.test.js
joostdecock: Ah ok, I see. These used to return different warnings depending on whether you try to remove a non-existing material type, or pass in a non-string.

However, since `removeCut` now calls `addCut` under the hood, as you so correctly suggested, the warning is the same.

Still, I think it's good to test both scenarios just in case we touch `addCut` later so I'll leave them bot in, but I'll update the title

joostdecock:
Screenshot from 2022-07-31 15-08-16

Geez Github, you make it sound so rude!

PR #2503, opened 2022-07-31
feat(plugin-grainline): Set part grain in macro

joostdecock:
This requires #2502 to be merged


BenJamesBen (approved):
The changes look fine to me.

Issue #2027, opened 2022-03-23
Add more information to pattern pieces to help distinguish iterations of a pattern 


TriploidTree:
**What is it that you would like to see happen?**

I would like to (optionally?) add the date I exported or created a pattern and the pattern title to each pattern piece or other such unique identifier, so I can trace back to the original pattern and not mix up my pieces.

**Are you a FreeSewing patron?**

 - [ ] Yes, I am :hugs:
 - [x] No, I am not :thinking:

**Additional context**
As a (somewhat untidy) sewist who prints updated patterns to correct issues, I currently have no way to readily distinguish patterns created with a given set of parameters. By including the date of export on each piece, I might have some hope of tracing back to which version of the pattern is in my hands, and also to ensure I don't mix up pieces that belong to different versions.
![image](https://user-images.githubusercontent.com/4170521/159799573-ca2aef3b-dd72-4b04-97d2-411e16d5719d.png)

starfetch:
This is something that I would also love to have (I label my stuff manually, or rather, try to). 
From the dev-side, I'm not sure how complicated it could get to figure out the placement of this "comment" box on each and every pattern piece, since that would (probably) require placing it manually -- unless it's anchored to the numbering.

BenJamesBen:
Adding customer-provided comments to a pattern piece would be a bigger task. However, we could easily add an exported timestamp to the title macro.
![Title Date](https://user-images.githubusercontent.com/109869956/182002240-2aa16bfc-f6c2-4f38-8c4d-838869f4782f.png)
In `plugins/plugin-title/src/index.js`:
```diff
94a95,102
>       shift += 8
>       const dateformat = require('dateformat')
>       const now = new Date()
>       this.points[`_${prefix}_exportDate`] = so.at
>         .shift(-90 - so.rotation, shift * so.scale)
>         .attr('data-text', dateformat(now, 'yyyymmdd"T"HHMMo'))
>         .attr('data-text-class', 'fill-note')
>         .attr('data-text-transform', transform(so.at.shift(-90 - so.rotation, shift * so.scale)))
```

joostdecock:
@BenJamesBen Let's go 🚀

joostdecock:
Date added by @BenJamesBen closing

PR #2527, opened 2022-08-03
Add timestamp to title.

BenJamesBen:
This change adds a timestamp to the Title produced by plugin-title, and it also updates and fixes a typo in its documentation.

This pull request accomplishes part of Enhancement Request #2027. However, because it doesn't completely address all aspects of it (things like adding comments/notes to patterns), I am leaving #2027 open.


markdown/org/docs/various/notation/title/en.md
```diff
@@ -2,7 +2,7 @@
 title: Titles
 ---
 
-Each pattern piece has a title that tells you the number and name of the piece, as well as the pattern name:
+Each pattern piece has a title that tells you the number and name of the piece, as well as the pattern name and a timestamp of when the pattern was rendered:
```
joostdecock, line 5: Perhaps *when the pattern was **drafted*** is clearer to users? To render is kinda nerdy, not sure everyone will get what that means.

plugins/plugin-title/src/index.js
```diff
@@ -92,6 +92,14 @@ export default {
           .attr('data-text-class', 'fill-current font-bold')
           .attr('data-text-transform', transform(so.at.shift(-90 - so.rotation, shift * so.scale)))
       }
+      shift += 8
+      const dateformat = require('dateformat')
+      const now = new Date()
+      this.points[`_${prefix}_exportDate`] = so.at
+        .shift(-90 - so.rotation, shift * so.scale)
+        .attr('data-text', dateformat(now, 'yyyymmdd"T"HHMMo'))
+        .attr('data-text-class', 'fill-note')
```
joostdecock, line 101: Could we make that class `text-sm` so it's a bit smaller and just black?

joostdecock (dismissed):

markdown/org/docs/various/notation/title/en.md
BenJamesBen: I made the suggested change.

plugins/plugin-title/src/index.js
BenJamesBen: Change made.

BenJamesBen:
I made the suggested changes. The title now looks like:
Title timestamp, smaller text

joostdecock (approved):

PR #2528, opened 2022-08-03
Fix SA slider to work when SA set to zero. Fixes #2501.

### LINKED ISSUE #2501: [bug]: [lab] Sliding to 0 cm SA invisibly toggles Seam Allowance to "no"

In the lab, if you slide the seam allowance slider to 0 cm, it seems to toggle the value of Seam Allowance to be "no" in an invisible and semi-persistent manner. That is, when the seam allowance is slider is subsequently set to a non-0cm value, even though Seam Allowance is set to "yes", the tool incorrectly behaves as if Seam Allowance was set to "no".

Steps to Reproduce:
1. On lab.freesewing.dev start a new pattern, for example *Simon* without breasts 40, and go to the DraftDesign page.
2. Under Settings, set Seam Allowance to "Yes". The seam allowance appears in the pattern, as expected.
3. Slide the Seam Allowance slider back and forth. The seam allowance grows and shrinks, as expected.
4. Slide the Seam Allowance slider to 0 cm. The seam allowance disappears, as expected.
5. Slide the Seam Allowance slider back to a non-0cm value, for example 1.5 cm, and slide the slider back and forth. Note that Seam Allowance is still set to "Yes". However, unexpectedly the seam allowance does not reappear in the pattern. This is the bug.
6. Slide the Seam Allowance slider to a non-0cm value, for example 1.5 cm. Toggle Seam Allowance to "No" and then back to "Yes". Note that the seam allowance reappears in the pattern. This is the workaround.

When in the bug state, switching to/from the LayoutThing page does not fix the problem. The bug is also seen in the localhost:8000 version of the lab. The bug is also seen when using Imperial units. The bug mainly affects customers who accidentally slide the SA slider to 0, slide it back to a non-zero value, and then wonder why the seam allowance didn't reappear as expected.

_No response_

No, I am not 😞

### Additional context

_No response_

BenJamesBen:
This fixes #2501, where seam allowance would temporarily stop working when the SA slider was set to zero. 

This pull request also changes the lab UI saBool setting label from "Seam allowance" to "Seam Allowance enabled". Previously there were 2 different settings with the same "Seam allowance" label, making it confusing when trying to discuss both settings in the same sentence.
Seam allowance enabled


BenJamesBen:
(The tests are giving errors because I added new strings to packages/i18n/src/locales/en/settings.yml but didn't change any of the other locale files. If I also need to include those other locale files in this PR, changing them by adding some sort of placeholder text, please let me know. I am not sure what the workflow is for this type of i18n-affected change.)

packages/i18n/src/locales/en/settings.yml
```diff
@@ -6,6 +6,10 @@ paperless:
   title: Paperless
   description: Drafts a pattern with all dimensions included so you can transfer it on fabric or another medium without the need to print
 
+sabool:
+  title: Seam allowance enabled
```
joostdecock, line 10: Perhaps `Include Seam Allowance` is more clear for this?

packages/i18n/src/locales/en/settings.yml
```diff
@@ -6,6 +6,10 @@ paperless:
...
+  description: Controls whether or not to use seam allowance in your pattern
```
joostdecock, line 11: And then here *whether or not to **include** seam allowance in your pattern*

sites/shared/components/workbench/menu/core-settings/core-setting-sa-bool.js
```diff
@@ -21,7 +21,7 @@ const CoreSettingSaBool = props => {
       <SumButton onClick={toggle}>
         <SumDiv>
           <Deg />
-          <span>{ t('settings:sa.t') }</span>
+          <span>{ t('settings:sabool.t') }</span>
           <span className="ml-4 opacity-50">
```
joostdecock, line 25: With the phrasing above, I think we can drop the yes/no.
I would add a translation string for `saSize` that reads `Seam allowance size` and use that for the other label

joostdecock (approved):
You should add the strings to the other locale files if you want the tests to pass.
You can use the English translations, it will be overwritten later by the translation software.

It's a bit clunky, I know.

joostdecock:
@BenJamesBen I approved it, because my remarks are minor. 
I just in general prefer to use two different labels:

- Include seam allowance
- Seam allowance size

Because your remark is correct, it's confusing to have the same twice with a different meaning.

BenJamesBen:
I will rework the settings the other way, to have `sa` be the seam allowance size and `saBool` to be whether or not to include the seam allowance. I'll resubmit the PR after doing so.

(I did initially debate which way the semantics should go. I thought that the term "include" could be unclear if someone were to believe that "all patterns have/include a seam allowance value, even if that value is zero".

BenJamesBen:
I made the suggested changes. The UI now looks like:
Include seam allowance

The previous code used locale translations for "yes", "no", and "Size". The new code no longer uses these. However, i have left the locale translation files intact. If the text should be removed from those files, please let me know.

joostdecock (approved):

PR #2549, opened 2022-08-07
Update Getting Started tutorials

BenJamesBen:
Some small changes and corrections to the Mac and Linux Getting Started tutorials, and a large change to the Windows tutorial.
- From observation while testing the tutorials, the commands are now `npm run dev` and `yarn dev` (instead of `npm run start` and `yarn start`).
- I changed the "Windows is unsupported" comment to a warning, with wording changes that are hopefully both clearer and friendlier.
- (We could further change the wording from "happy to assist you" to "happy to try to assist you", if we need to promise less.)
- In the Windows tutorial I re-arranged the sections to have a single "Setup/Start the Development Environment" section at the end. For those sections, I updated the instructions (copied from the Mac/Linux tutorials).

I have not tested the Windows tutorial instructions on a Windows system, but I think that the instructions I added should be correct. (I am still not sure about the remainder of the instructions which were in the original document.)

(To view the updated web pages in a browser, I learned that in the PR repository you can `cd sites/dev && yarn start`.)

### DIFF
--- a/markdown/dev/tutorials/getting-started-linux/dev-setup/en.md
+++ b/markdown/dev/tutorials/getting-started-linux/dev-setup/en.md
@@ -86,11 +86,11 @@ npx @freesewing/new-design
 ```
 
 After you've answered [some questions](#questions), it will take a while to set everything up.
-When it's done, you will have a new folder with the development environent inside.
+When it's done, you will have a new folder with the development environment inside.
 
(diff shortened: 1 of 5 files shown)


joostdecock (approved):
TIL I can't write environment without tripping over my own fingers :)

PR #2551, opened 2022-08-07
Fix Simon notch typos

joeschofield0:
I replaced "single" and "double" notch with the symbols for these notches (to reflect the with the pattern notation guide https://freesewing.org/docs/various/notation/notches/ ).
I also tweaked the word order of a couple of sentences for clarity.

### DIFF
--- a/markdown/org/docs/patterns/simon/instructions/en.md
+++ b/markdown/org/docs/patterns/simon/instructions/en.md
@@ -85,7 +85,7 @@ You can now put your cuffs aside, we'll attach them later.
 
 ![Sew collar and undercollar together](3a.png)
 
-Place both collar parts with the good sides together, and sew them together at the standard seam allowance.
+Place both collar parts with the _good sides together_, and sew them together at the standard seam allowance.
 
 #### Trim seam allowance
 
--- a/markdown/org/docs/patterns/simon/instructions/en.md
+++ b/markdown/org/docs/patterns/simon/instructions/en.md
@@ -85,7 +85,7 @@ You can now put your cuffs aside, we'll attach them later.
 
 ![Sew collar and undercollar together](3a.png)
 
-Place both collar parts together with the good sides, and sew them together at the standard seam allowance.
+Place both collar parts with the good sides together, and sew them together at the standard seam allowance.
 
 #### Trim seam allowance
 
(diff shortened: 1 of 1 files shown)


BenJamesBen:
I'm not sure how I feel about having the "single/double" words replaced with symbols. I find the web page easier to read with words rather than symbols, since the two symbols are very similar at first glance. I would be okay with including the symbols if they were in addition to the "single/double" words.

joeschofield0:
> I'm not sure how I feel about having the "single/double" words replaced with symbols. I find the web page easier to read with words rather than symbols, since the two symbols are very similar at first glance. I would be okay with including the symbols if they were in addition to the "single/double" words.

Hey Ben :)  Yes, I take your point and realise the symbols might not be accessible for people.  The main thing for me is that the pattern doesn't include any _double notches_, so this bit of text really confused me when I was doing my first 2 Simons.

markdown/org/docs/patterns/simon/instructions/en.md
```diff
@@ -85,7 +85,7 @@ You can now put your cuffs aside, we'll attach them later.
 
 ![Sew collar and undercollar together](3a.png)
 
-Place both collar parts together with the good sides, and sew them together at the standard seam allowance.
+Place both collar parts with the good sides together, and sew them together at the standard seam allowance.
```
joostdecock, line 88: Can you use it like this:
```
with _good sides together_
```

That way it will get picked up by our jargon plugin

markdown/org/docs/patterns/simon/instructions/en.md
```diff
@@ -461,9 +461,9 @@ Spread your shirt open so that both fronts and back lie flat with their good sid
 
 Now place your sleeve on top with the good side down, matching the top of the sleeve with the shoulder seam.
 
-> Be careful, your sleeve is not symmetric, so be sure to have the double notch (and cuff guard) on the back side, and the single notch on the front side.
+> Be careful, your sleeve is not symmetric, so be sure to have the ⊗ notch (and cuff guard) on the back side, and the ⊙ notch on the front side.
```
joostdecock, line 464: Can we do:
```
have the back notch (⊗) ... and the front notch (⊙) on the ...
```
so that we have both the words and the symbols?

markdown/org/docs/patterns/simon/instructions/en.md
```diff
@@ -461,9 +461,9 @@ Spread your shirt open so that both fronts and back lie flat with their good sid
...
 >
-> Also note that the double notch on the back of your sleeve does not have a corresponding notch on the back. That's because it should match up with the seam where your yoke joins the back.
+> Also note that the ⊗ notch on the back of your sleeve does not have a corresponding notch on the back of the shirt. That's because it should match up with the seam where your yoke joins the back.
```
joostdecock, line 466: Same here, words + symbol:
```
that the back notch (⊗) on the back...
```

joostdecock (changes requested):

joostdecock:
@joeschofield0 thanks so much for this. Nice catch, double notches is clearly wrong.
I've suggested some tweaks to accommodate the remark by @BenJamesBen that symbols only might be harder to read.
Apart from that, I'm happy to merge this :)

markdown/org/docs/patterns/simon/instructions/en.md
joeschofield0: Done :)

markdown/org/docs/patterns/simon/instructions/en.md
joeschofield0: Done.

markdown/org/docs/patterns/simon/instructions/en.md
joeschofield0: Done

joostdecock (approved):

joostdecock:
@all-contributors please add @joeschofield0 for docs

Issue #2558, opened 2022-08-08
[feature]: Add notches to the Titan block to indicate specific points and aid alignment


TriploidTree:
### What is it that you would like to see happen? 🤔

As the Titan block is used for generating subsequent patterns (whether by hand or through FreeSewing), it would be useful to mark key notches on the block.

Hip and knee would be useful notches to help when choosing where to place style elements. It would also be helpful to verify the block is landing at the correct points during fitting.

Notches additionally help with alignment of pattern pieces during sewing. A double notch on the back seam would help identify front from back quickly (and help with alignment when sewing the crotch curve).

### Are you a FreeSewing patron? 😃

No, I am not 😞

bobgeorgethe3rd:
I will look into adding these. Though I suspect the hip is going to be a bit tricksy.

bobgeorgethe3rd:
Closed as of #2568

PR #2568, opened 2022-08-10
Titan Notches for #2558

bobgeorgethe3rd:
As mentioned in #2558 this update adds the following notches.

- Cross and Crotch Seam notches
- Hip notches (when the rise is high enough)
- Seat Notches
- Knee notches

Changed logoAnchor.y to fork.y to minimise overlap of new paths.

Another thing I added was Hip Line and Seat Line paths to better identify where the hip line is with the notches.

I have also changed the waistbandWidth option to allow 0%, whilst the waistband being included is awesome for extending it is not the best for those who are using Titan as a physical block or for those who are testing Titan to check their measures.
Whilst not strictly necessary for extending Titan as some of these notches may be in a different place when extended, it will benefit those who are using Titan as it is described in its instructions. 

This has made me think about another possible addition to Titan which will benefit those extending it so may be another update soon. 

Added options.fitGuides so it does not display on Charlie and Paco.

Edit: Added seat notches.
Edit2: Added fitGuides

### DIFF
--- a/designs/titan/config/index.js
+++ b/designs/titan/config/index.js
@@ -49,6 +49,7 @@ export default {
     fitCrossSeam: true,
     fitCrossSeamFront: true,
     fitCrossSeamBack: true,
+    fitGuides: true,
 
     // Fit
     waistEase: { pct: 2, min: 0, max: 10 },
--- a/designs/titan/config/index.js
+++ b/designs/titan/config/index.js
@@ -71,6 +71,6 @@ export default {
     crotchSeamCurveAngle: { deg: 25, min: 0, max: 35 },
     waistBalance: { pct: 60, min: 30, max: 90 },
     grainlinePosition: { pct: 45, min: 30, max: 60 },
-    waistbandWidth: { pct: 3, min: 1, max: 6, snap: elastics, ...pctBasedOn('waistToFloor') },
+    waistbandWidth: { pct: 3, min: 0, max: 6, snap: elastics, ...pctBasedOn('waistToFloor') },
   },
 }
--- a/designs/titan/config/index.js
+++ b/designs/titan/config/index.js
@@ -72,6 +72,6 @@ export default {
     crotchSeamCurveAngle: { deg: 25, min: 0, max: 35 },
     waistBalance: { pct: 60, min: 30, max: 90 },
     grainlinePosition: { pct: 45, min: 30, max: 60 },
-    waistbandWidth: { pct: 3, min: 0, max: 6, snap: elastics, ...pctBasedOn('waistToFloor') },
+    waistbandWidth: { pct: 3, min: 1, max: 6, snap: elastics, ...pctBasedOn('waistToFloor') },
   },
 }
--- a/designs/titan/src/back.js
+++ b/designs/titan/src/back.js
@@ -221,7 +221,7 @@ export default (part) => {
       to: points.grainlineBottom,
     })
     macro('scalebox', { at: points.knee })
-    points.logoAnchor = new Point(points.crossSeamCurveStart.x / 2, points.crossSeamCurveStart.y)
+    points.logoAnchor = new Point(points.crossSeamCurveStart.x / 2, points.fork.y)
     snippets.logo = new Snippet('logo', points.logoAnchor)
     points.titleAnchor = points.logoAnchor.shift(-90, 60)
     macro('title', {
--- a/designs/titan/src/back.js
+++ b/designs/titan/src/back.js
@@ -271,7 +271,7 @@ export default (part) => {
 	}
 	macro('sprinkle', {
 	snippet: 'notch',
-	on: ['kneeInNotch', 'kneeOutNotch', 'seatOutNotch']
+	on: ['seatOutNotch', 'kneeInNotch', 'kneeOutNotch']
 	})
 	macro('sprinkle', {
 	snippet: 'bnotch',
--- a/designs/titan/src/back.js
+++ b/designs/titan/src/back.js
@@ -231,68 +231,68 @@ export default (part) => {
     })
 	//notches
   if (options.fitGuides) {
-	points.waistMid = points.waistOut.shiftFractionTowards(points.waistIn, 0.5)
-	points.seatMid = points.waistMid.shiftTowards(points.waistIn, measurements.waistToSeat).rotate(90, points.waistMid)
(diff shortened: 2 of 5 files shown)


bobgeorgethe3rd:
Ok turns out there was some confusion over hip and seat.
I will be adding seat notches to this so have converted to a draft for the minute.

bobgeorgethe3rd:
Added.

bobgeorgethe3rd:
Crap just checked on gitpod need to add an option to turn it off in extended patterns

bobgeorgethe3rd:
Had to do it manually because my permissions need a look into annoyingly.

designs/titan/src/back.js
```diff
@@ -221,15 +221,80 @@ export default (part) => {
...
+    points.logoAnchor = new Point(points.crossSeamCurveStart.x / 2, points.fork.y)
     snippets.logo = new Snippet('logo', points.logoAnchor)
     points.titleAnchor = points.logoAnchor.shift(-90, 60)
     macro('title', {
       nr: 1,
       title: 'back',
       at: points.titleAnchor,
     })
-
+	//notches
```
BenJamesBen, line 232: For this entire section of new code, indentation is missing. While not technically necessary, I recommend/request that indentation be added for readability and maintainability.

designs/titan/src/front.js
```diff
@@ -295,15 +295,78 @@ export default (part) => {
...
+	on: ['hipsIn', 'hipsOut',]
+	})
+	paths.hipline = new Path()
+	.move(points.hipsOut)
+	.line(points.hipsIn)
+	.attr('class', 'fabric help')
+	.attr('data-text', 'Hip Line')
+	.attr('data-text-class', 'center')
+	}
+}
```
BenJamesBen, line 369: Similarly, in this whole section I found it very difficult to read the non-indented code to determine what the code was supposed to do. Please add indentation.

designs/titan/src/back.js
```diff
@@ -221,15 +221,80 @@ export default (part) => {
...
+	points.kneeInNotch = utils.lineIntersectsCurve(points.kneeIn, points.kneeOut.rotate(180, points.kneeIn), points.fork, points.forkCp2, points.kneeInCp1, points.floorIn)
+	}
+	macro('sprinkle', {
+	snippet: 'notch',
+	on: ['seatOutNotch', 'kneeInNotch', 'kneeOutNotch']
+	})
+	macro('sprinkle', {
+	snippet: 'bnotch',
+	on: ['crossSeamCurveStart', 'seatIn']
+	})
```
BenJamesBen, line 280: On the Back, you are using bnotches for these two points. However, on the Front you are using regular notches for the same points. Is this intended?

BenJamesBen (changes requested):

BenJamesBen:
FYI, this is what the added help lines and notches look like on the upper sections of the parts.
Screen Shot 2022-08-12 at 12 30 50 PM

designs/titan/src/back.js
bobgeorgethe3rd: Yes it identifies the cross seam. This seam is connected back to back whereas the others are connected to the front. One of the requests was to add notch identifiers to the back. This fufills that request.

designs/titan/src/back.js
bobgeorgethe3rd: I have made this changed but personally I find the opposite to be true.

designs/titan/src/front.js
bobgeorgethe3rd: I have made this changed but personally I find the opposite to be true.

bobgeorgethe3rd:
> FYI, this is what the added help lines and notches look like on the upper sections of the parts. Screen Shot 2022-08-12 at 12 30 50 PM
> 
> I tested the changes in the lab, and the actual notch and helpline placements seem correct.

I did test in the lab and had realised I forgot to check Charlie and Paco. I was working on the local end when making the changes hence my misstep. This change would be very difficult to do without visual representation.

designs/titan/src/back.js
BenJamesBen: Okay. I still don't understand why crossSeamCurveStart and seatIn are bnotches instead of notches, and why so only on the back but not the front. However, if you say that this arrangement is intentional, then I am fine with it.

BenJamesBen (commented):
The indentation is still not quite correct. (All the nestled if-else statements are incorrectly at the same indentation level. And, you start your indentation at level 1 but incorrectly end up at level 0.) And, I suspect that some code lines are at the wrong indentation levels as a result. But, the code is now easier to read, where it is easier to see where the indentation errors are, so thank you for making the changes.

As I mentioned in a comment buried within the code review, I still don't quite understand why those two notches are bnotches instead of regular notches, and why so only on the back but not the front. However, since you said that this arrangement is intentional, I am okay with it.

designs/titan/config/index.js
```diff
@@ -71,6 +72,6 @@ export default {
     crotchSeamCurveAngle: { deg: 25, min: 0, max: 35 },
     waistBalance: { pct: 60, min: 30, max: 90 },
     grainlinePosition: { pct: 45, min: 30, max: 60 },
-    waistbandWidth: { pct: 3, min: 1, max: 6, snap: elastics, ...pctBasedOn('waistToFloor') },
+    waistbandWidth: { pct: 3, min: 0, max: 6, snap: elastics, ...pctBasedOn('waistToFloor') },
```
joostdecock, line 75: This seems out of scope for this PR

joostdecock (changes requested):
Please run `yarn prettier` in `designs/titan` then submit the changes.

joostdecock:
To be clear, @BenJamesBen is correct that the indentation is a bit all over the place, making it hard to read.

However, you don't have to worry about code formatting.  Just run `yarn prettier` and it will be done for you.

designs/titan/config/index.js
bobgeorgethe3rd: I thought it would be useful but I will change it back

bobgeorgethe3rd:
Figured out what my gitpod issue was :) Thank you for telling me about `yarn prettier`. I had no idea how to fix the indentation in the way needed.

BenJamesBen (approved):
The revised code changes look fine.

(I was wondering why you reverted the waistband minimum back to 1%, but for the purpose of this code review it doesn't matter.)

joostdecock (approved):

PR #2583, opened 2022-08-13
Wahid grainlines, issue 2564

OysteinHoiby:
Wahid grainlines, issue 2564

### DIFF
--- a/designs/wahid/src/back.js
+++ b/designs/wahid/src/back.js
@@ -108,9 +108,10 @@ export default (part) => {
 
   if (complete) {
     //Grainline
-    macro('grainline', {
-      from: points.cutonfoldVia1,
-      to: points.cutonfoldVia2,
+    macro('cutonfold', {
+      from: points.cbNeck,
+      to: points.cbHem,
+      grainline: true,
     })
     macro('scalebox', { at: new Point(points.logo.x, points.armholePitchCp2.y) })
     if (sa) {
--- a/designs/wahid/src/back.js
+++ b/designs/wahid/src/back.js
@@ -107,6 +107,11 @@ export default (part) => {
   paths.dart.render = false
 
   if (complete) {
+     //Grainline
+     macro('grainline', {
+      from: points.cutonfoldVia1,
+      to: points.cutonfoldVia2
+    })
     macro('scalebox', { at: new Point(points.logo.x, points.armholePitchCp2.y) })
     if (sa) {
       paths.sa = paths.saBase
--- a/designs/wahid/src/back.js
+++ b/designs/wahid/src/back.js
@@ -107,10 +107,10 @@ export default (part) => {
   paths.dart.render = false
 
   if (complete) {
-     //Grainline
-     macro('grainline', {
+    //Grainline
+    macro('grainline', {
       from: points.cutonfoldVia1,
-      to: points.cutonfoldVia2
+      to: points.cutonfoldVia2,
     })
     macro('scalebox', { at: new Point(points.logo.x, points.armholePitchCp2.y) })
     if (sa) {
--- a/designs/wahid/src/front.js
+++ b/designs/wahid/src/front.js
@@ -1,5 +1,5 @@
-import { constructMainDart, shapeSideSeam, dartPath } from './shared'
 
+import { constructMainDart, shapeSideSeam, dartPath } from './shared'
 export default (part) => {
   let {
     points,
@@ -17,11 +17,9 @@ export default (part) => {
     paperless,
     store,
   } = part.shorthand()
-
   // Cleanup from Brian
   for (let i of Object.keys(paths)) delete paths[i]
   delete snippets.armholePitchNotch
-
   // Neck cutout
   points.closureTop = new Point(
     measurements.chest * options.frontOverlap * -1,
@@ -41,24 +39,19 @@ export default (part) => {
   points.armholeHollow = points.armholeHollow.shift(180, frontInset / 2)
   points.armholeHollowCp2 = points.armholeHollowCp2.shift(180, frontInset / 2)
   points.armholeHollowCp1 = points.armholeHollowCp1.shift(180, frontInset / 2)
-
   // Shoulder inset
   points.shoulder = points.shoulder.shiftTowards(points.neck, shoulderLen * options.shoulderInset)
   points.shoulderCp1 = points.shoulderCp1.shift(
     points.shoulder.angle(points.neck),
     shoulderLen * options.shoulderInset
   )
-
   // Neck inset
   points.neck = points.neck.shiftTowards(points.shoulder, shoulderLen * options.neckInset)
   points.neckCp2 = points.neck.shift(points.shoulder.angle(points.neck) + 90, shoulderLen * 0.2)
-
   // Construct main dart
   constructMainDart(part)
-
   // Shape side seam
   shapeSideSeam(part)
-
   // Hem
   if (options.hemStyle === 'classic') {
     // Construct cutaway
@@ -120,16 +113,13 @@ export default (part) => {
   // (essentially a rounded hem with radius zero
   if (!points.roundStart) points.roundStart = points.closureBottom.clone()
   if (!points.roundEnd) points.roundEnd = points.closureBottom.clone()
-
   // Add dart start and end point regardless of style or front or back
   points.dartStart = options.hemStyle === 'classic' ? points.splitDartHemLeft : points.dartHemLeft
   points.dartEnd = options.hemStyle === 'classic' ? points.splitDartHemRight : points.dartHemRight
-
   // Pockets
   let pw = measurements.hips * options.pocketWidth // Pocket width
   let pwh = pw * options.weltHeight // Pocket welt height
   let pwvh = pwh / Math.cos(utils.deg2rad(options.pocketAngle)) // Pocket welt vertical height
-
   points.pocketTopMid = points.dartWaistCenter.shiftFractionTowards(
     points.dartHipCenter,
     options.pocketLocation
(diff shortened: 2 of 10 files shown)


OysteinHoiby:
Hm, not sure how I broke this. It worked fine in the lab 🙈

designs/wahid/src/front.js
```diff
@@ -273,22 +259,40 @@ export default (part) => {
...
+      from: points.grainLineFromFront,
+      to: points.grainLineToFront,
+    })
     if (sa) {
       paths.sa = paths.saBase
         .offset(sa)
         .join(paths.hemBase.offset(sa))
         .close()
-        .attr('class', 'fabric sa')
+        .attr('classF', 'fabric sa')
```
BenJamesBen, line 294: Is "classF" a typo?

designs/wahid/src/front.js
```diff
@@ -273,22 +259,40 @@ export default (part) => {
...
+    if (options.hemStyle === 'classic') {
+      points.grainLineFromFront = new Point(points.hemTip.x, points.closureTop.y)
+      points.grainLineToFront = new Point(points.hemTip.x, points.hemTip.y).shift(90, 20)
+    }
+    if (options.hemStyle === 'rounded') {
+      points.grainLineFromFront = new Point(points.roundEnd.x, points.roundEnd.y).shift(90, 200)
+      points.grainLineToFront = new Point(points.roundEnd.x, points.roundEnd.y)
+    }
+    if (options.hemStyle === 'square') {
+      points.grainLineFromFront = new Point(points.cutonfoldTo.x, points.cutonfoldTo.y).shift(
```
BenJamesBen, line 279: When using square hemStyle, the grainline can extend beyond the pattern. You might want to think about having the endpoint based on points.closuretop.y.

Screen Shot 2022-08-13 at 7 05 45 PM

designs/wahid/src/frontfacing.js
```diff
@@ -29,19 +27,42 @@ export default (part) => {
...
+      points.grainlineFromFrontFacing = new Point(points.roundEnd.x, points.roundEnd.y).shift(
+        90,
+        200
+      )
+      points.grainlineToFrontFacing = new Point(points.roundEnd.x, points.roundEnd.y)
+    }
+    if (options.hemStyle === 'square') {
+      points.grainlineFromFrontFacing = new Point(points.cutonfoldTo.x, points.cutonfoldTo.y).shift(
+        90,
+        200
```
BenJamesBen, line 46: Similarly, the front facing grainline extends past the pattern with the square hem.

Screen Shot 2022-08-13 at 7 05 45 PM

BenJamesBen (changes requested):

packages/i18n/src/next.mjs
```diff
@@ -7,20 +7,17 @@ import enNamespaces from "./next/en/index.mjs"
...
-export const uk = ukNamespaces
 
 export const languages = {
   de: "Deutsch",
   en: "English",
   es: "Español",
   fr: "Français",
-  nl: "Nederlands",
-  uk: "undefined"
+  nl: "Nederlands"
```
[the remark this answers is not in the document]
OysteinHoiby: That is correct ☺️

designs/wahid/src/back.js
```diff
@@ -107,6 +107,11 @@ export default (part) => {
   paths.dart.render = false
 
   if (complete) {
+    //Grainline
```
joostdecock, line 110: We need a combined cutonfold/grainline indicator here, as such:

```js
macro('cutonfold', {
  from: points.cbNeck,
  to: points.cbHem,
  grainline: true,
})
```

designs/wahid/src/front.js
```diff
@@ -273,22 +259,41 @@ export default (part) => {
...
-
     // Facing/Lining boundary (flb)
     paths.flbFacing = new Path()
       .move(points.dartTop)
       .curve(points.flbCp, points.flbCpTop, points.flbTop)
       .attr('class', 'fabric')
     paths.flbLining = paths.flbFacing.clone().attr('class', 'lining dashed')
 
+    // Grainline
+    let grainlineVariableLength = points.closureTop.dist(points.cfHips)
```
joostdecock, line 270: There's a lot going on here. Any reason this won't do?

```js
macro('grainline', {
  to: points.neck,
  from: new Point(points.neck.x, points.cfHem.y)
})
```

designs/wahid/src/frontfacing.js
```diff
@@ -1,11 +1,9 @@
 export default (part) => {
   let { points, Point, paths, Path, options, macro, snippets, Snippet, complete, sa } =
     part.shorthand()
-
   // Cleanup from front part
   for (let i of Object.keys(paths)) delete paths[i]
   for (let i of Object.keys(snippets)) delete snippets[i]
```
joostdecock, line 6: Change this line to:
```js
  for (let i of Object.keys(paths).filter(name => name !== 'grainline')) delete paths[i]
```

designs/wahid/src/frontfacing.js
```diff
@@ -29,19 +27,43 @@ export default (part) => {
...
+    }
+    if (options.hemStyle === 'square') {
+      points.grainlineFromFrontFacing = new Point(points.cutonfoldTo.x, points.cutonfoldTo.y).shift(
+        90,
+        grainlineVariableLength
+      )
+      points.grainlineToFrontFacing = new Point(points.cutonfoldTo.x, points.cutonfoldTo.y)
+    }
 
+    macro('grainline', {
```
joostdecock, line 52: We don't need all this, we can re-use the grainline from the front part. (if you change line 5 as indicated above)

packages/i18n/src/next.mjs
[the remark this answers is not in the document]
OysteinHoiby: And i do not know how to remove it :)

designs/wahid/src/frontlining.js
```diff
@@ -23,8 +22,21 @@ export default (part) => {
...
+    let grainlineVariableLength = points.armhole.dist(points.hem)
+    let grainlineVariableShift = points.dartHipRightCpTop.dist(points.hipsCp2) * 0.1
+    points.grainlineFromLining = new Point(points.dartEnd.x, points.dartEnd.y)
+      .shift(90, grainlineVariableLength)
+      .shift(0, grainlineVariableShift)
+    points.grainlineToLining = new Point(points.dartEnd.x, points.dartEnd.y).shift(
+      0,
+      grainlineVariableShift
+    )
+    macro('grainline', {
```
joostdecock, line 36: Here too, can't we just:
```js
macro('grainline', {
  from: points.flbTop,
  to: new Point(points.flbTop.x, points.cfHem.y)
})
```

(note that you'll need to add `Point` to the shorthand call to do this)

designs/wahid/src/pocketbag.js
```diff
@@ -45,7 +41,23 @@ export default (part) => {
...
+    grainlineVariableShift
+  )
+  points.grainlineToPocketbag = new Point(points.topLeft.x, points.topLeft.y)
+    .shift(0, grainlineVariableShift)
+    .shift(-90, ph)
+  points.grainlineToPocketbagRotated = points.grainlineToPocketbag.rotate(
+    options.pocketAngle,
+    points.grainlineFromPocketBag
+  )
+  macro('grainline', {
```
joostdecock, line 57: Here I would suggest:
```js
macro('grainline', {
  from: points.roundLeftEnd,
  to: new Point(points.roundLeftEnd.x, points.topLeft.y)
})
```

joostdecock (changes requested):
I did not go through all parts, because my remarks are sort of the same everywhere: Just put a grainline between 2 points (that are already there are easy to construct from what's already there).

You're going through a lot of effort to create a bunch  of points, when you can just use what's already there.

I gave some examples for how it can be done with just the macro. Let me know what you think.

designs/wahid/src/frontfacing.js
joostdecock: It's about line `5`, seems I misclicked

designs/wahid/src/back.js
OysteinHoiby: If we add a cutonfold, won't that give us one back piece instead of two? That is not in accordance with the sewing instructions for Wahid, so in such case we must update the instructions as well :)

designs/wahid/src/front.js
OysteinHoiby: Nope, that would work just fine (:

designs/wahid/src/back.js
BenJamesBen: Again, I believe that the original code and instructions were correct. There should not be a foldline on the Wahid back part. So, I believe that your code change should be reverted (but, I'm hoping that Joost will clarify the situation).

BenJamesBen (changes requested):
I tested the most recent code changes, and I believe that the new grainlines fit appropriately with all design options and sizes. The only thing I think needs to be changed is to remove the new cut-on-fold and re-add the grainline on the back part, but again, I am hoping that Joost will clarify whether this should be done.

designs/wahid/src/back.js
joostdecock: Apologies for the tardiness of my response, but @BenJamesBen is correct, the back is not cut on the fold, but joined with a center back seam.

This matters because there's an option to shape the back which does so on the center back seam, so it can't be cut on the fold: https://freesewing.org/docs/patterns/wahid/options/centerbackdart

So yeah, no cut on fold, only grainline.

joostdecock (approved):
Merging this, will deal with the back issue myself

joostdecock:
@all-contributors please add @OysteinHoiby for code

OysteinHoiby:
Sorry Joost :) got a bit caught up with school this week

joostdecock:
@OysteinHoiby good point, I've moved it to the side in e1d10aba628d7a6ed8b5e4675ea559a514be3a98

PR #2477, opened 2022-07-26
fix(simon): Remove superfluous notches. Fixes #2399

BenJamesBen:
This fix removes the superfluous notches in Simon's button and buttonhole plackets. 

It also removes the floating bnotch that appears with a split yoke. I came to the conclusion that the bnotch was probably a leftover debug tool or otherwise unintended.

(I am not sure whether I also need to update `designs/simon/CHANGELOG.md`, but I didn't see it updated in a similar previous PR.)

PR #2603, opened 2022-08-17
Add Bust-aligned buttons option to Simone. Resolves #2154

### LINKED ISSUE #2154: Add an option to Simone to ensure a button always falls at the bust apex
**What is it that you would like to see happen?**
In order to reduce gaping in shirts of people with breasts (particularly large breasts), it is common to move the buttons so that one button is at the height of the bust apex. This can often require a little shuffing of the buttons to ensure it looks well, so the buttons above the bust point may end up slightly closer together to accomodate the move. 

**Are you a FreeSewing patron?**

 - [ ] Yes, I am :hugs:
 - [ ] No, I am not :thinking:

**Additional context**

BenJamesBen:
This PR adds a Bust-aligned Buttons option to Simone, with 3 settings: Disabled (default), Even spacing, Split spacing. By default, the existing button spacing is used. With the setting set to Even or Split spacing, button spacing is arranged to ensure that a button is placed at the bustline to prevent possible gaping. This fulfills Enhancement Request #2154 .

The PR also:
- Adds a notch at the center front bustline to help sewists who might want to add a snap or other closure there, even if bust-aligned button spacing isn't used.
- Changes the location of points.bust after the FBA so it is the actual bust point. Additional comments are added to the code, documenting this change.
- Fixes the center front hem notch on separate buttonhole placket by moving it to the correct adjusted location.
- Re-adds the center front armhole, waist, and hem notches to the button placket. These notches were accidentally removed in PR #2477 .
- Removes a superfluous bust point notch from the buttonhole placket (overlooked in PR #2477 ).
- Updates changelog files.

I believe that I tested the changes to make sure they work with all designs and styles: with separate and attached button/buttonhole packets, with classic and seamless button/buttonhole plackets, and with all equivalent Simon variations to check for regressions.

### DIFF
--- a/designs/simone/config/index.js
+++ b/designs/simone/config/index.js
@@ -57,9 +57,10 @@ const config = {
     frontDartLength: { pct: 45, min: 30, max: 60 },
     contour: { pct: 50, min: 30, max: 75 },
     bustAlignedButtons: {
-      dflt: 'Disabled',
-      list: ['Even spacing', 'Split spacing', 'Disabled'], },
-  },
+      dflt: 'disabled',
+      list: ['even', 'split', 'disabled'],
+    },
+  }
 }
 
 export default config
--- a/designs/simone/config/index.js
+++ b/designs/simone/config/index.js
@@ -1,18 +1,31 @@
 import { version } from '../package.json'
 import { config as simonConfig } from '@freesewing/simon'
 
+// Add the Bust-aligned Buttons option into the "Style->Closure" submenu.
+// - Because the closure submenu is an object in the style array,
+//   we cannot simply replace/overwrite it by adding a new closure
+//   submenu. Doing so would just create a duplicate "Closure"
+//   submenu.
+// - Instead, we would need to remove the existing closure submenu
+//   object from simonConfig.optionGroups.style first, before adding
+//   the new closure submenu. (Doing it this way would also have the
+//   side effect of changing the position of the "Closure" submenu
+//   within the main"Style" menu.)
+// - Alternately, we can simply modify the existing closure submenu
+//   object in place, and that is what we are doing here: injecting the
+//   new 'bustAlignedButtons' option directly into the closure submenu
+//   array. This seems to be the least disruptive method.
+const closureindex = simonConfig.optionGroups.style.findIndex((item) =>
+    typeof item === 'object' && "closure" in item)
+simonConfig.optionGroups.style[closureindex].closure.push('bustAlignedButtons')
+
 const config = {
   version,
   ...simonConfig,
   name: 'simone',
   optionGroups: {
     ...simonConfig.optionGroups,
-    style: [
-        ...simonConfig.optionGroups.style,
-        'frontDarts',
-        'contour',
-        'bustAlignedButtons',
-    ],
+    style: [...simonConfig.optionGroups.style, 'frontDarts', 'contour'],
     advanced: [
       ...simonConfig.optionGroups.advanced,
       'bustDartAngle',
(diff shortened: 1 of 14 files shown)


joostdecock (changes requested):
Excellent PR 🙌 

If you move the option under *style* => *closure* in the option groups, I will merge it.
If you are unsure how to do so, let me know and I will merge it, and then add it.

Thanks for another awesome contribution 🙏

designs/simone/config/index.js
```diff
@@ -7,7 +7,12 @@ const config = {
   name: 'simone',
   optionGroups: {
     ...simonConfig.optionGroups,
-    style: [...simonConfig.optionGroups.style, 'frontDarts', 'contour'],
+    style: [
```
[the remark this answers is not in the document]
BenJamesBen: I moved the Bust-Aligned Buttons option into the Style->Closure submenu.

BenJamesBen:
If there was a more straightforward method I could/should have used to add the Bust-aligned Buttons option into the existing Closures submenu, please let me know.

designs/simone/config/index.js
```diff
@@ -1,6 +1,24 @@
...
+//   the new closure submenu. (Doing it this way would also have the
+//   side effect of changing the position of the "Closure" submenu
+//   within the main"Style" menu.)
+// - Alternately, we can simply modify the existing closure submenu
+//   object in place, and that is what we are doing here: injecting the
+//   new 'bustAlignedButtons' option directly into the closure submenu
+//   array. This seems to be the least disruptive method.
+const closureindex = simonConfig.optionGroups.style.findIndex((item) =>
+    typeof item === 'object' && "closure" in item)
+simonConfig.optionGroups.style[closureindex].closure.push('bustAlignedButtons')
```
joostdecock, line 20: Nicely done. FYI, we're going to change this in v3 to make it easier to handle this.
We'll only accept a top-level key and an array of options. For nested structure, you would use a `:` in they key.

So here you would have just added:
```js
"style:closure": [ "bustAlignedButtons"]
```
And that would then get merged into the closure group.

Still on the fence about using `:` or `.` as a separator, but this is the idea.

packages/i18n/src/locales/en/options/simone.yml
```diff
@@ -1,3 +1,7 @@
+bustAlignedButtons:
```
joostdecock, line 1: If we move this to `simon.yml` the tests should pass (and it will get inherited by simone anyway)

joostdecock (approved):

joostdecock:
> Also please let me know if I need to add anything to Simon

Since you've added the option, you should add the translation to simon too, that will fix the tests.
And as simone inherits its translations from simon, it should *just work*.

packages/i18n/src/locales/en/options/simone.yml
BenJamesBen: I moved the option translation from Simone to Simon.

BenJamesBen:
I removed the Bust-aligned Buttons option and translation from Simon and moved them entirely to Simone. Trying to put them into Simon was the wrong thing to do, to work around an error. I changed simon/src/shared.js to fix the error (to simply check whether the bustAlignedButtons option existed before trying to use it).

I also fixed the translation to have the list options translated as well, which was the way I should have done it in the first place.

I made changes to:
```
	modified:   designs/simon/config/index.js
	modified:   designs/simon/src/shared.js
	modified:   designs/simone/config/index.js
	modified:   packages/i18n/src/locales/en/options/simon.yml
	modified:   packages/i18n/src/locales/en/options/simone.yml
```

I am still getting an i18n test error in Simon, and I am investigating why this is occurring. Simon should no longer be aware of the bustAlignedButtons option, so it should not expect a translation for it:
`Error: No option translation found for bustAlignedButtons in simon`

joostdecock (approved):

PR #2629, opened 2022-08-22
Lucy documentation! 

SeaZeeZee:
In the right place this time, with the linedrawing for the front page added.


designs/lucy/src/pocket.js
```diff
@@ -51,10 +51,10 @@ export default function (part) {
       .line(points.middle)
       .attr('class', 'path fabric dashed')
 
-    points.logo = points.center.shiftOutwards(points.middle, pocketWidth / 5)
+    points.logo = points.centerLeft.shift(270, pocketWidth / 3)
```
BenJamesBen, line 54: I note that the new location can sometimes cause the logo to encroach upon the left seam, for example if the width is decreased. Your current change is an improvement from what exists now so I'm okay with the change as is. However, you might want to consider lowering it a bit more.
Screen Shot 2022-08-22 at 3 22 12 PM

BenJamesBen (changes requested):
You've accidentally included the `sites/lab/core.839` file as part of this PR, and it probably should be removed from the PR.

(I believe that if you delete the file from your GitHub SeaZeeZee:lucy-doc repository branch, it should remove itself from the PR automatically?)

Otherwise, the PR looks fine to me.

designs/lucy/src/pocket.js
SeaZeeZee: I will lower it! Skully hates me, but I want them on there!

joostdecock (changes requested):
Please export your SVG from inkscape as a plain SVG. We have no use for all that inkscape-specific stuff.

SeaZeeZee:
> Please export your SVG from inkscape as a plain SVG. We have no use for all that inkscape-specific stuff.

Done!

joostdecock (approved):

Issue #2403, opened 2022-07-12
Code QA for ursula/unice


joostdecock:
**What seems to be the problem?**

There's been a bunch of changes to Ursula/Unice lately that did not get much scrutiny.
We should review the code to make sure we're following best practices.

Mostly creating this so I don't forget it myself :)

bobgeorgethe3rd:
https://github.com/freesewing/freesewing/pull/2636
Reverts Ursula
Updates Unice

joostdecock:
Thank you @nataliasayang I'd love to have a chat to clear all this up because I feel people are tripping over each other's feet.

I've added it to [the agenda of the next contributor call](https://github.com/freesewing/freesewing/discussions/2613).

Meanwhile, I might merge @anna-puk 's PR as it is only about Unice. We'll figure out what to do with any changes made to Ursula that you aren't in favor of during the call.

I've also put an item on the agenda to remind people to be respectful of ownership/assignments when rolling up their sleeves, as working together also requires some restraint from time to time :wink:

BenJamesBen:
I submitted draft PR #2642 with some initial work done on the code review.

However, I'm wondering whether it might be preferable to simply fork/clone Unice from Ursula, and have Unice be its own separate pattern rather than trying to inherit from another. It seems as if it might make things easier in the future, especially if Unice would otherwise need to make such large changes to Ursula in order to work.

I'm talking about after v3, of course.

BenJamesBen:
Ursula has an "elastic" part that just has text with the amount of elastic needed. This part has no paths, and because of this the part does not appear in LayoutThing: ForPrinting (though the text does print when the PDF is exported). 

If Paperless is set to Yes, the part does appear in LayoutThing: ForPrinting. I believe this is because it has a dummy horizontal dimension. I assume that the dummy dimension was added as a workaround, to allow the part to be seen and moved around.

I note that the issue can be fixed without the dimension workaround by adding a box around the text:
```js
  paths.box = new Path()
     .move(new Point(-10,-10))
     .line(new Point(-10,15))
     .line(new Point(200,15))
     .line(new Point(200,-10))
     .close()
```

BenJamesBen:
[...]

PR #2636, opened 2022-08-23
Solve for #2403

bobgeorgethe3rd:
First off
- Reverted Ursula back to form before add markers update
I have done this so it returns to Natalia's original pattern and have left it as such.

- Reverted Unice to same point
Had to revert Unice to get to a point I could work on it.

- Add markers back to Unice
This was tricky and explains why Ursula got split into doubles. TLDR The legOpeningLength for notches requires all parts to be draft first so have created frontBase and backBase to allow this.


joostdecock:
@bobgeorgethe3rd I'm afraid I'll have to close this.

- 2.22 is as good as released, so more fundamental changes would have to be part of a port to v3
- I don't think you discussed these changes with the original designers, which risks alienating them
- #2403 is already assigned to two people

PR #2638, opened 2022-08-23
Unice: update gusset and add curve options (pr version)

anna-puk:
Fixes:
- bug which caused Unice's version of back and gusset parts to be used (accidental 'return part' at the start of the code)
- missing path lengths which prevented the notches / 'markers for elastic' from being drawn

Improvements:
- updated gusset: always curves inward, less wide by default, based on a measurements that's less sensitive to girth
- new option 'front curve': as an alternative (or in combination with) front dip, allow the front to curve around the belly (top of front pattern piece will be straight on paper but curved when worn)
- new option 'back curve': analogous to front curve, but for the back part


designs/unice/src/back.js
```diff
@@ -1,4 +1,321 @@
...
+  // recalculate backWaistBandRight  
+  points.backWaistBandRight = points.backWaistBandLeft.flipX(points.backWaistMid)
+  
+
+  // if back is both curved and dipped, apply the dip
+  // backDipOrCurve2 is zero if this isn't needed, so apply regardless of value
+  points.backWaistBandMid = points.backWaistBandMid.shift(270, measurements.waistToUpperLeg * backDipOrCurve2 * store.get('yScaleReduced'))        
+  
+  // report the expected 'drop' in waist band from side to center (when worn) 
+  raise.info(['expectedDropFromWaistToCenterBack', units(measurements.waistToUpperLeg * (options.backCurve + options.backDip))])
```
BenJamesBen, line 85: This currently prints an events log entry like: `expectedDropFromWaistToCenterBack0.88cm`

If you wanted to make it look prettier, you could change it to something like:
```js
  const measurementString = units(measurements.waistToUpperLeg * 
    (options.backCurve + options.backDip))
  raise.info("The expected drop for the back center waist is " + measurementString + ".")
``` 

to produce `The expected drop for the back center waist is 0.88cm.`

designs/unice/config/index.js
```diff
@@ -32,6 +32,16 @@ const config = {
       'gussetWidth',
       'gussetLength'
     ],
+    style: [
+      'rise',
+      'legOpening',
+      'frontDip',
+      'frontCurve',
+      'backDip',
+      'backCurve',
```
BenJamesBen, line 41: You'll also want to make sure you add the new `frontCurve` and `backCurve` settings to i18n strings and the documentation markdown pages.

BenJamesBen:
I tested your changes, and increasing frontCurve or backCurve seems to cause the bottom of the piece to decrease in size and the the piece to become shorter. Is that the expected effect?

What fitting situations are these settings supposed to address? Is it, "my bottom is actually smaller than my waist measurement might suggest; please adjust the pattern accordingly"?

BenJamesBen (changes requested):
You'll also want to update the `config/changelog.yaml` file.

joostdecock:
Thanks @anna-puk When @BenJamesBen 's remarks have been addressed, I'm happy to merge this.
Especially the remarks of using `Point.clone()` are important, because you'll get some weird bugs if you re-use a point since both points will (in memory) be the same object.

anna-puk:
> I tested your changes, and increasing frontCurve or backCurve seems to cause the bottom of the piece to decrease in size and the the piece to become shorter. Is that the expected effect?
> 
> What fitting situations are these settings supposed to address? Is it, "my bottom is actually smaller than my waist measurement might suggest; please adjust the pattern accordingly"?

As mentioned in my description of the pull request, the front curve option 'allows the front to curve around the belly (top of front pattern piece will be straight on paper but curved when worn)'. The pattern piece will indeed be shorter in the vertical direction (which saves fabric), but when worn, the effect is similar to the existing 'front dip' option: height stays the same at the sides but is reduced at the waist.

(I've tested this both via the TestDesign in the lab and by creating test garments and can confirm that it works as intended.)

designs/unice/config/index.js
anna-puk: Where do I find those?

designs/unice/src/back.js
anna-puk: I thought I read somewhere that the strings are supposed to be defined elsewhere, to make translation easier... Am I mistaken?

designs/unice/src/front.js
```diff
@@ -163,14 +182,57 @@ export default function (part) {
   points.frontGussetLeftCp1 = points.frontGussetLeft
     .shift(270, Math.max(Math.max(points.frontGussetLeft.dy(points.frontWaistMid) * options.taperToGusset / 2,points.frontGussetLeft.dy(points.frontLegOpeningLeft) * 2),points.frontGussetLeft.dy(points.frontWaistBandMid)))
 
+  // straighten the part (if needed)
+  let curveAsAngle
+  if (options.frontCurve > 0) {
+    // create copies of the original points
+    points.frontWaistBandLeftPreCurve = points.frontWaistBandLeft
+    points.frontLegOpeningLeftPreCurve = points.frontLegOpeningLeft
+    points.frontLegOpeningLeftCp1PreCurve = points.frontLegOpeningLeftCp1
```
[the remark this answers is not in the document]
anna-puk: Ah, thanks, I did not know that!

Is it okay to keep these lines of code (after adding the .clone()) even though the points are not used anywhere? I think I might need them later, but I'm not sure whether that's enough of a reason to keep them.

anna-puk:
> You'll also want to update the `config/changelog.yaml` file.

What does this file do and what changes do I need to make?

BenJamesBen:
> > You'll also want to update the `config/changelog.yaml` file.
> 
> What does this file do and what changes do I need to make?

The changelog file contains the "release notes" for FreeSewing and its designs. By looking at it, people will be able to see that Unice was added in version 2.21.0 on June 27th and that you changed it in version X.XX on such-and-such date, once it gets released. I would add something like:
```diff
--- a/config/changelog.yaml
+++ b/config/changelog.yaml
@@ -16,6 +16,9 @@ Unreleased:
       - Fix hem allowance. Closes [#2350](https://github.com/freesewing/freesewing/issues/2350)
     theo:
       - migrated `waistbandWidth` option from `mm` to snapped `pct` option
+    unice:
+      - Added new Front Curve and Back Curve style options.
+      - Updated gusset to always curve inward.
 
   Fixed:
     breanna:
```

designs/unice/src/front.js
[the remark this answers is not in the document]
BenJamesBen: Yes, you can keep the lines of code if you believe having the points will be useful in the future, even if the design doesn't use them now. Doing this will also help other pattern designers understand what your design is doing when they read your code.

designs/unice/src/back.js
joostdecock: Correct, but we'll add that in v3.

That being said, it's a best practice to split translatable strings and units.

See: https://freesewing.dev/reference/api/hooks/inserttext

designs/unice/config/index.js
BenJamesBen: The i18n strings are in:
`packages/i18n/src/locales/en/options/unice.yml`

And, the docs are in this directory:
`markdown/org/docs/patterns/unice/options`
You would create `frontcurve` and `backcurve` subdirectories and an `en.md` file within each new subdirectory. (I believe that all the other language files will automatically get created.)

You can look around the options directory to see how the other options are documented. (I'm not sure how the superimposed images are created, so you might want to omit the "Effect of this option on the pattern" sections for now.)

designs/unice/src/back.js
BenJamesBen: Given that information, it is your choice to leave your code as is (to make it easier to have translated strings in the future) or to change it (to make the message text prettier now). Either way is fine.

anna-puk: I'll keep it the way it is, then.

anna-puk:
I've addressed all remarks (see recent commits). I see that the changes to changelog.yaml resulted in a conflict and the automatic merge seems to be putting the new entries under an existing release. @BenJamesBen, would you be willing to resolve the conflict?

BenJamesBen:
> I've addressed all remarks (see recent commits). I see that the changes to changelog.yaml resulted in a conflict and the automatic merge seems to be putting the new entries under an existing release. @BenJamesBen, would you be willing to resolve the conflict?

Hmm... I am not able to resolve the conflict myself. (The merge conflict occurred because the changelog.yaml that was in your repository was an older version, but the file had since changed to https://github.com/freesewing/freesewing/blob/develop/config/changelog.yaml .) 

Are you able to edit the merged file in your repository to put the Unice changes into the right places in the new file? Something like:

```
  Changed:
    plugin-title:
      - Use localized date format
    unice:
      - Updated gusset to always curve inward.

  Added:
    unice:
      - Added new Front Curve and Back Curve style options.
```

BenJamesBen (approved):
The updated changes appear fine to me.

anna-puk:
Thanks for the suggestions, @BenJamesBen! I ended up simply grabbing the latest version of the changelog off of GitHub and adding the unice-related lines there. I still got a conflict, but that one was straight-forward to resolve.

I think everything is in a good shape now.

joostdecock:
@anna-puk Am I correct to assume this is superseded by #2857

anna-puk:
> @anna-puk Am I correct to assume this is superseded by #2857

No, they are independent. (I felt that would be easier since these changes were to be backported to v2.)

This PR was originally for v2, so I will check whether everything is still in order.

anna-puk:
Now updated for v3. TestDesign shows that everything is working as intended.

designs/unice/src/gusset.mjs
```diff
@@ -37,22 +37,42 @@ export const gusset = {
...
+        .shift(270, 0.3 * actualGussetLength) // shift to same height as for high back exposure
+        .shift(0, horShift)
+      points.gussetCp1 = points.gussetCp1.shiftFractionTowards(
+        points.frontGussetLeft,
+        Math.max(0, options.backExposure) - backExposureThreshold
+      )
+    }
 
     // Flip points to right side
     points.gussetCp2 = points.gussetCp1.flipX(store.get('frontGussetMid'))
```
BenJamesBen, line 78: I note that this change, later down below in the code, would put the dimension inside the part, so it doesn't intersect the leg seam:
```diff
       macro('vd', {
         from: points.frontGussetRight,
         to: points.backGussetRight,
-        x: points.frontGussetRight.x + sa + 15,
+        x: points.frontGussetRight.x - sa - 15,
       })
     }
```

And, this would put the dimension outside the part entirely:
```diff
       macro('vd', {
         from: points.frontGussetRight,
         to: points.backGussetRight,
-        x: points.frontGussetRight.x + sa + 15,
+        x: points.backGussetRight.x + sa + 15,
       })
     }
```

(Just suggestions, please feel free to ignore.)

config/changelog.yaml
```diff
@@ -2466,3 +2474,20 @@ Unreleased:
...
+    Status
+    Docs
+    Contact GitHub
+    Pricing
+    API
+    Training
+    Blog
+    About
+
+You have no unread notifications
```
BenJamesBen, line 2493: I'm not sure what all this text is about. Was it accidentally added? Or, did one of the lint commands add this text automatically?

BenJamesBen (changes requested):
The only change I'm requesting is removing the `changelog.yaml` text, if you accidentally added it. (If it was one of the lint programs that added it for you, then please feel free to leave it in.). The rest of my comments are just optional suggestions.

I tested this PR in my localhost lab, and it seems to produce patterns just fine.

config/changelog.yaml
anna-puk: I have no idea were it came from, so I removed it.

anna-puk:
> The only change I'm requesting is removing the `changelog.yaml` text, if you accidentally added it. (If it was one of the lint programs that added it for you, then please feel free to leave it in.). The rest of my comments are just optional suggestions.
> 
> I tested this PR in my localhost lab, and it seems to produce patterns just fine.

I removed the superfluous section in the changelog.

joostdecock (approved):

PR #2639, opened 2022-08-23
[Fix] Mix inline and class styling for buttons and buttonholes

### LINKED ISSUE #2631: [bug]: Buttons and Buttonholes render/display incorrectly on latest release.

- Buttons appear black and filled in.

- Buttonholes appear purple and are hard to distinquish from notches.

_No response_

No, I am not 😞

### Additional context

image

eriese:
Fixes #2631 by leaving the inline styling in for svg rendering and using classes for pdf rendering. The svg to pdf parser won't use css vars, unfortunately. 

I think it's not ideal for maintenance to have to use both to accomplish the same outcomes in different places, but looking at our css, I'm kind of itching to use nesting to make it easier to read, navigate, and group, and I think that in that effort it'd be easier to put styling into the shadow dom in ways that don't require a ton of additional overhead.

In the mean time, this is backwards compatible for v2 and org, and it allows the snippets to render properly in pdf exports.

### DIFF
--- a/plugins/plugin-buttons/src/button.js
+++ b/plugins/plugin-buttons/src/button.js
@@ -2,10 +2,11 @@ export default `
 <g id="button">
   <circle
     cx="0" cy="0" r="3.4"
+    style="stroke:var(--pattern-mark);fill:none;stroke-width:var(--pattern-stroke);"
     class="mark"
   />
-  <circle cx="-1" cy="-1" r="0.5" class="no-stroke fill-mark"/>
-  <circle cx="1"  cy="-1" r="0.5" class="no-stroke fill-mark" />
-  <circle cx="1"  cy="1"  r="0.5" class="no-stroke fill-mark" />
-  <circle cx="-1" cy="1"  r="0.5" class="no-stroke fill-mark" />
+  <circle cx="-1" cy="-1" r="0.5" class="no-stroke fill-mark" style="stroke:none;fill:var(--pattern-mark)"/>
+  <circle cx="1"  cy="-1" r="0.5" class="no-stroke fill-mark" style="stroke:none;fill:var(--pattern-mark)" />
+  <circle cx="1"  cy="1"  r="0.5" class="no-stroke fill-mark" style="stroke:none;fill:var(--pattern-mark)" />
+  <circle cx="-1" cy="1"  r="0.5" class="no-stroke fill-mark" style="stroke:none;fill:var(--pattern-mark)" />
 </g>`
--- a/plugins/plugin-buttons/src/buttonhole.js
+++ b/plugins/plugin-buttons/src/buttonhole.js
@@ -3,17 +3,20 @@ export default `
   <path
     class="mark"
     d="M -1,-5 L 1,-5 L 1,5 L -1,5 z"
+    style="fill:none;stroke:var(--pattern-mark);stroke-width:var(--pattern-stroke);"
   />
 </g>
 <g id="buttonhole-start">
   <path
     class="mark"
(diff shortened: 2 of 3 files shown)


BenJamesBen (commented):
I tested the changes and verified that the buttons and buttonholes still appear correct in Chrome desktop browser for the localhost lab.

joostdecock:
I haven't had coffee yet, so perhaps this is a dumb question, but why do we need a fix for something that is already fixed?

#2631 is fixed in the `v2` while your initial changes that broke v2 remain in `develop`.

Seems to me everything is fine, so why generate extra work? Am I missing something @eriese & @BenJamesBen ?

eriese:
The issue is that what's in develop now still doesn't work in Firefox. 

It was also my understanding that we want to include print layout and export to pdf in a final v2 release, so I thought it was necessary to have it not break anything where the plugin is used in org while also making sure it works where the plugin is used in lab, but I may have misunderstood the necessity of that

joostdecock:
> *The issue is that what's in develop now still doesn't work in Firefox.*

Didn't realize that. But I still don't think this is the right fix.
When you change the classes on the button or buttonhole snippet, you're essentially causing a breaking change for people who theme their own patterns.

> It's perhaps something that I have not been clear enough about, but I'm waaaay more easily persuaded to merge some stuff in our lab or other code that is part of the push for v3.
>
> But if you make changes to existing plugins, or god forbid core, there's a much higher bar to pass.
> Are these changes needed? Do they make sense? Is it the right approach? And so on.
> **When in doubt, reach out and ask.**

In this case, I remain unconvinced that this is the right approach.
Case in point: I've fixed it with 4 lines of CSS in the lab: 67de3b260ac3131d61b035f7ce9f25d3f892d8c7
Screenshot from 2022-08-24 18-51-54

PDF export seems fine to me too. [I've attached a PDF so you can see](https://github.com/freesewing/freesewing/files/9418030/freesewing-legend-3.pdf)
You can verify for yourself in the lab: https://lab.freesewing.dev/v/next/legend

> *It was also my understanding that we want to include print layout and export to pdf in a final v2 release, so I thought it was necessary to have it not break anything where the plugin is used in org while also making sure it works where the plugin is used in lab, but I may have misunderstood the necessity of that*

The plugin(s) just add snippets to the SVG. They apply generic classes to them so that users can theme FreeSewing.
Then, there are two ways to render those patterns:

- Render in React: In this case style comes from `sites/shared/styles/svg-freesewing-draft.css` while colours are set in the theme(s)
- Render in core: Spits out SVG, and you should use the `@freesewing/plugin-theme` (or your own theme plugin) which will include the styles in the SVG 

What I don't want to keep on mucking about in v2 is:
- `@freesewing/plugin-buttons`: This does a simple thing and does it well
- `@freesewing/plugin-theme`: Has served us well for years, let's keep it as is

What I am happy to consider is changes to the CSS in a NextJS environment (`sites/shared/style`).
That being said, the way things work together and the unique challenges of styling the shadow-dom in SVG often make styling more tricky than it seems on the surface.

I would suggest to reach out and discuss things, but I also know that time zone differences don't always make that so easy.
So I find myself trying to explain why I think this should not be merged, but I can only imagine that this is very frustrating for contributors.

So I won't close this at this time, for I don't want to underestimate @eriese 's capacity to convince me.
But as things stand now, I don't see why we need to make changes to plugins, nor why we need anything special to style exported PDFs.

joostdecock:
@eriese Checked the deploy of your latest export changes, all snippets look fine both on-screen and in the PDF: https://freesewing-dnw5n2h0b-freesewing.vercel.app/v/next/legend

Can this be closed?

PR #2654, opened 2022-08-24
[feature] PDF export part 2

eriese:
Now Implemented
[ x] cover page is optional
[ x] cover page is styled  
[ x] disable export button if there are 0 pages
[ x] use a web worker to handle the pdf generation
[ x] show loading indicator while generating the pdf
[ x] " instead of inch
[ x] add TODO note about migrating plugin to new parts implementation once available
[ x] check for points that have text or circles, all paths don't render in `Part.prototype.isEmpty`


BenJamesBen:
I'm seeing a behavior in DraftDesign that occurs after exporting a pattern as PDF or SVG.
1. For example, Simon Without Breasts Size 40.
2. In DraftDesign, the pattern appears as expected.
3. Go to ExportThing and export the pattern as SVG or one of the PDF options.
4. Go back to DraftDesign.
5. The web page is now much longer than expected, and the pattern appears in the middle of the page with lots of whitespace above and below. You need to scroll down to view the pattern.

The same thing occurs when exporting as PDF from the LayoutThing: ForPrinting page.

(Also, I note that changing the part layout in LayoutThing: ForPrinting doesn't change the layout in the DraftDesign page. I am assuming this is the correct and expected behavior, that the printing layout is supposed to be different than the DraftDesign layout.)

BenJamesBen:
Bug: A blank cover page is printed when coverPage is Disabled and page A1 is blank.

1. With Bob, go to LayoutThing: ForPrinting and set coverPage to Disabled.
3. Select a page size, and Export a PDF. Note that the cover page is not printed, as expected.
4. Move the bib part to A2 or B1 (or anywhere else), leaving page A1 blank, and Export a PDF.
5. Note that a blank cover page is unexpectedly and incorrectly printed.
6. Move the bib part back to A1, and Export a PDF.
7. Note that the cover page is again correctly no longer printed.

BenJamesBen:
For layouts that are taller than wide, the layout image can encroach the header on the cover page.
Screen Shot 2022-08-24 at 1 15 24 PM

eriese:
@joostdecock I somehow seem to have messed up the node version in the builds here, but I can't tell how. Any thoughts?

eriese:
> I'm seeing a behavior in DraftDesign that occurs after exporting a pattern as PDF or SVG.
> 
> 1. For example, Simon Without Breasts Size 40.
> 2. In DraftDesign, the pattern appears as expected.
> 3. Go to ExportThing and export the pattern as SVG or one of the PDF options.
> 4. Go back to DraftDesign.
> 5. The web page is now much longer than expected, and the pattern appears in the middle of the page with lots of whitespace above and below. You need to scroll down to view the pattern.
> 
> The same thing occurs when exporting as PDF from the LayoutThing: ForPrinting page.

This is a weird, long-standing issue with the SizeMe component, and I haven't investigated what triggers it, but I don't think exports are the only thing. We should log it, but it's not been introduced by this PR.

> (Also, I note that changing the part layout in LayoutThing: ForPrinting doesn't change the layout in the DraftDesign page. I am assuming this is the correct and expected behavior, that the printing layout is supposed to be different than the DraftDesign layout.)

yes, this is expected behavior

Looking into the other two. Thanks for catching them!

BenJamesBen:
> This is a weird, long-standing issue with the SizeMe component, and I haven't investigated what triggers it, but I don't think exports are the only thing. We should log it, but it's not been introduced by this PR. Edit: jk, I found it

This seems to fix the issue, though I don't know if it might have any side effects:
```diff
--- a/sites/shared/components/workbench/draft/svg-wrapper.js
+++ b/sites/shared/components/workbench/draft/svg-wrapper.js
@@ -28,14 +28,14 @@ import Part from './part'
 const SvgWrapper = props => {
   const { patternProps, gist, app, updateGist, unsetGist, showInfo } = props
 
-  return <SizeMe>{({ size }) => (
+  return <SizeMe monitorHeight>{({ size }) => (
     <TransformWrapper
       minScale={0.1}
       centerZoomedOut={true}
       wheel={{ activationKeys: ['Control'] }}
     >
       <TransformComponent>
-        <div style={{ width: size.width+'px', }}>
+        <div style={{ width: size.width+'px', height: size.height+'px', }}>
           <Svg {...patternProps} embed={gist.embed}>
             <Defs {...patternProps} />
             <style>{`:root { --pattern-scale: ${gist.scale || 1}} ${patternProp:
```

eriese:
> > This is a weird, long-standing issue with the SizeMe component, and I haven't investigated what triggers it, but I don't think exports are the only thing. We should log it, but it's not been introduced by this PR. Edit: jk, I found it
> 
> This seems to fix the issue, though I don't know if it might have any side effects:
> 
> ```diff
> --- a/sites/shared/components/workbench/draft/svg-wrapper.js
> +++ b/sites/shared/components/workbench/draft/svg-wrapper.js
> @@ -28,14 +28,14 @@ import Part from './part'
>  const SvgWrapper = props => {
>    const { patternProps, gist, app, updateGist, unsetGist, showInfo } = props
>  
> -  return <SizeMe>{({ size }) => (
> +  return <SizeMe monitorHeight>{({ size }) => (
>      <TransformWrapper
>        minScale={0.1}
>        centerZoomedOut={true}
>        wheel={{ activationKeys: ['Control'] }}
>      >
>        <TransformComponent>
> -        <div style={{ width: size.width+'px', }}>
> +        <div style={{ width: size.width+'px', height: size.height+'px', }}>
>            <Svg {...patternProps} embed={gist.embed}>
>              <Defs {...patternProps} />
>              <style>{`:root { --pattern-scale: ${gist.scale || 1}} ${patternProp:
> ```

This seems like probably a good solution in general; I'm also not sure about side effects. It turned out that this was happening because `gist.embed` was getting set to `false` in the export, and it needs to be `true` on the draft page, so I've just fixed that. 

Everything else should be fixed too

BenJamesBen:
I re-tested, and all the issues I previously found are now fixed.

eriese:
> I re-tested, and all the issues I previously found are now fixed.

Thank you so much!

BenJamesBen:
Not a bug, but just a note that cover sheet images can get ridiculously small if the parts are moved far down/across the layout.
Screen Shot 2022-08-24 at 3 00 13 PM

BenJamesBen:
For the column identifiers, I notice that special characters and non-ascii characters are used. Is this intentional? I mean "[\\]^_`" (which appear after A-Z) and "{|}~" (which appear after a-z) and non-ASCII characters like the Euro, dagger, double-dagger, and ellipses symbols (which appear afterward).

Some of the non-ASCII characters are non-printing, so just the row number "1" appears in the circle, and the alignment diamond to the right is blank. These characters include the one after "~" and before the Euro character. Another is the one after the Euro character and before the subscript single-quote character. Another is the one after that Latin character that looks like the letters "OE" pushed together.
Screen Shot 2022-08-24 at 3 17 29 PM

eriese:
> For the column identifiers, I notice that special characters and non-ascii characters are used. Is this intentional? I mean "[\]^_`" (which appear after A-Z) and "{|}~" (which appear after a-z) and non-ASCII characters like the Euro, dagger, double-dagger, and ellipses symbols (which appear afterward).
> 
> Some of the non-ASCII characters are non-printing, so just the row number "1" appears in the circle, and the alignment diamond to the right is blank. These characters include the one after "~" and before the Euro character. Another is the one after the Euro character and before the subscript single-quote character. Another is the one after that Latin character that looks like the letters "OE" pushed together. Screen Shot 2022-08-24 at 3 17 29 PM

I foolishly imagined 26 columns would be enough and didn't bother with wrapping around. It's fixed now

BenJamesBen:
I updated and verified that the columns now go A-Z then AA-AZ, BA-... .

(Not a bug in your PR; it is also seen in existing lab.freesewing.dev. The logo doesn't resize on the LayoutThing: ForPrinting screen when Scale is changed. If I move the Scale slider up and down, the logo stays the same size. However, when the PDF is exported, the logo does end up being the correct Scale size.)

However, I note that your "A1", "A2" center page numbers, "2 cm" text, and the "A", "1" diamond text, they all scale up/down with Scale, and I do not know whether that was intended. At the smallest scale the text cannot be read, and at the largest scale the text is awkwardly big.

joostdecock:
> @joostdecock I somehow seem to have messed up the node version in the builds here, but I can't tell how. Any thoughts?

@eriese we've moved on to NodeJs >= 16.

I thought you'd be happy 😉

I'm working on ESM only, that should be for later this week.

packages/core/src/part.js
```diff
@@ -416,7 +416,20 @@ Part.prototype.setCutOnFold = function (p1, p2) {
 }
 
 Part.prototype.isEmpty = function() {
```
joostdecock, line 418: I'm still on the fence about this. I remember you said that you can't trust designs to have accurately set `render=false` but I feel that we should fix the underlying issue, rather than paper over it with this check.

joostdecock (approved):
So I'm still on the fence about `part.isEmpty()` but not enough to block this ;)

PR #2655, opened 2022-08-24
Remove Workbench and Next gist re-creation links.

BenJamesBen:
The [design].freesewing.dev Workbench and next.freesewing.org Next gist re-creation links no longer work, and my understanding is that they will not be restored in the future. This PR removes the links from the bug reports generated on GitHub.


joostdecock (changes requested):
We should add a link to the lab here:

```
https://lab.freesewing.dev/v/next/[design]?from=github&preload=[gist]
```

BenJamesBen:
Hmm... I botched my branch and PR trying to resolve a merge in the one file. There are now too many changed files.  I will close this PR and open a new one.

PR #2673, opened 2022-08-27
Add json type assertions for import statements.

BenJamesBen:
This PR appends `assert { type: 'json' }` to every "import from .json" statement that did not already have it.

### DIFF
--- a/config/templates/design/config.js
+++ b/config/templates/design/config.js
@@ -1,4 +1,4 @@
-import { version } from './package.json'
+import { version } from './package.json' assert { type: 'json' }
 
 export default {
   name: '{{name}}',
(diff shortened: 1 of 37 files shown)


joostdecock:
Happy to merge this, but I am working on this and more in the `v3-phase1` branch. So be mindful that things are going to be rather volatile until we land on a more or less stable v3 setup.

I've mentioned that the `v2` and `main` branches contain the more stable code, but I'm not aure everyone got the memo.

To be clear: I don't want to discourage you from making changes. But I want to protect you from being disappointed when over the next couple of weeks more sweeping changes undo some of this work.

That being said, I appreciate the effort 👍

joostdecock (approved):

PR #2696, opened 2022-09-01
Move aaron plugin config to part-level.

BenJamesBen:
- This is a PR into the v3 branch.
- This PR removes the plugin info from `index.mjs` for aaron. 
- However, it does _not_ move the plugin info into any other file. 
 
This is because aaron's `front.mjs` imports `base` from brian, and that part already imports pluginBundle. As I understand it, plugins are configured only once in the first part that uses them, and they should not be re-configured again in other parts. If this is incorrect, please let me know.

### DIFF
--- a/designs/aaron/src/index.mjs
+++ b/designs/aaron/src/index.mjs
@@ -1,6 +1,4 @@
 import { Design } from '@freesewing/core'
-import { pluginBundle } from '@freesewing/plugin-bundle'
-import { withCondition as bustPlugin } from '@freesewing/plugin-bust'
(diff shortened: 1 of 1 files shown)


joostdecock:
> This is because aaron's front.mjs imports base from brian, and that part already imports pluginBundle. As I understand it, plugins are configured only once in the first part that uses them, and they should not be re-configured again in other parts. If this is incorrect, please let me know.

That is correct. Or rather, that is how I intent it to work. We still might run into some bugs, but this is certainly the right way.

joostdecock (approved):

PR #2699, opened 2022-09-01
Move bee and bella plugin config to part-level. Use new data property.

BenJamesBen:
PR into v3 branch.

There is an issue with bee where the imported bella front and back parts are incorrectly not hidden, and these two parts are appearing by default in the lab.


joostdecock (approved):

joostdecock:
Hiding dependencies is now implemented: https://github.com/freesewing/freesewing/commit/bf7926b98d95b30b1a8b6dc5cf3fd14c53d316df

Example for bee: https://github.com/freesewing/freesewing/commit/89b0f3f4c1a3a0c315883994c3b2d509de8b988d

PR #2700, opened 2022-09-01
Complete porting albert to v3.

BenJamesBen:
I was not able to get albert's `crossBox` local plugin to not get `ReferenceError: Cannot access 'crossBox' before initialization` errors. I even tried moving the plugin to a separate file outside of `index.mjs`, but that did not work. For now, the `crossBox` plugin is commented out in `front.mjs`.


joostdecock:
Thanks @BenJamesBen I'll look into the plugin issue

Issue #2695, opened 2022-08-31
[bug]: [v3] Parts uninvolved in dependencies do not get added


BenJamesBen:
### What seems to be the problem? 🤔

In v3, if a Part is not involved in any "from" or "after" dependencies, then it does not get added to the pattern.

- Parts with a from/after dependency get added.
- Parts referred to in a from/after dependency get added.
- All other parts do not get added.
- Designs with only a single Part will end up creating zero-part Patterns.

### Design / Plugin / Package 🧐

packages/core

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

I saw this problem when porting Albert and Bob to v3. Both designs do not have any part dependencies, so the patterns end up partless and blank. With Benjamin, the bow parts are involved in dependencies and show up. However, the ribbon part is uninvolved in dependencies and does not appear.

Bob is a design with a single part so there are no from/after dependencies. In the v3 lab, Bob produces a pattern with zero parts. The bib part does not appear in the lab, neither in the Events log nor under Contents.

As a workaround, when I add a dummy base part by creating `base.mjs`...
```
export const base = {
  name: 'bob.base',
  draft: part => {
    return part
  }
}
```

... and add the dependency to `bib.mjs`...
```
import { base } from './base.mjs'
export const bib = {
  name: 'bob.bib',
  after: base,
```

... then the bib part appears in the pattern.

BenJamesBen:
This issue was fixed via commit e5f398bcab7e9619cab4f91b29c2c129032a3cc7. Closing.

PR #2702, opened 2022-09-01
Port benjamin to v3.

BenJamesBen:
I believe this PR preserves history. If there still is a history problem with this PR, please let me know.

(In my previous PR for bob, I did actually use `git mv` to rename the files. However, I did not `git commit` before editing and `git add`ing the updated files before the commit, and I think that subsequently caused the rename to be changed to delete/add by git. This time I immediately `git commit` after the renames.)

I verified in the lab that v3 benjamin works. However, currently the ribbon part does not appear due to #2695 .


joostdecock:
If you look at the changes for this PR, it still removes the old files and added the new files (rather than renaming them).

I think we should fix this. While I don't mind for the bib pattern, I feel merging would be unfair to previous contributors to this code.

BenJamesBen:
@joostdecock  What am I doing wrong in the process? Specifically, I am:
1. `git mv oldfile newfile` for all the files.
2. `git commit`
3. Then, edit the new files.
4. `git commit` again.

joostdecock:
I'm not entirely certain, I feel that *should* do the trick.

I'll put up a PR myself to see what happens, bit it'll be tomorrow because I'm AFK tonight.

We'll figure it out somehow 😉

PR #2704, opened 2022-09-01
[wip]: Draft changes for Breanna, Bruce, and Titan

BenJamesBen:
This is a draft PR with changes that I had been working on for Breanna, Bruce, and Titan. I do not wish to put back the changes or do further work until after the switchover to v3 is complete.

- For Breanna, the change prevents spurious Warning messages from being logged when there is no secondary bust dart.
- For Bruce and Titan, we are adding "runaway rotation/shift" protection. Iterative rotations/shifts theoretically make changes that keep getting closer and closer to the target state. However, sometimes they can stop getting closer to the target and instead start to get further and further away. This code change monitors and stops the process if it sees that it has stopped producing better results.
- For Bruce and Titan, new Warning messages have been added to alert the customer when it is known that a non-optimal pattern has been produced, telling them that _"Manual fitting/alteration of the Front pattern piece may be needed."_ and to _"Please recheck your measurements, make a test garment to check fit, and alter if necessary."_
- For Titan, code has been added to produce Warning messages if certain points fall on the other side of knee or hip points. Previously, these unexpected scenarios would cause the pattern to fail with an Error.

### DIFF
--- a/designs/breanna/src/dart-utils.js
+++ b/designs/breanna/src/dart-utils.js
@@ -84,27 +84,37 @@ export const getDartLocationsAsNumbers = (options) => {
   else return [loc1, loc2]
 }
 
-export const getDartPaths = (Path, points) => [
-  new Path()
-    .line(points.primaryBustDart1)
-    .line(points.primaryBustDartTip)
-    .line(points.primaryBustDart2),
-  new Path()
-    .line(points.secondaryBustDart1)
-    .line(points.secondaryBustDartTip)
-    .line(points.secondaryBustDart2),
-]
+export const getDartPaths = (Path, points) => {
+  let dart_paths = [
+    new Path()
+      .line(points.primaryBustDart1)
+      .line(points.primaryBustDartTip)
+      .line(points.primaryBustDart2),
+  ]
+  if ('secondaryBustDart1' in points)
+    dart_paths.push(new Path()
+      .line(points.secondaryBustDart1)
+      .line(points.secondaryBustDartTip)
+      .line(points.secondaryBustDart2)
+    )
+  return dart_paths
+}
 
-export const getSaDartPaths = (Path, points) => [
-  new Path()
-    .line(points.primaryBustDart1)
-    .line(points.primaryBustDartEdge)
-    .line(points.primaryBustDart2),
-  new Path()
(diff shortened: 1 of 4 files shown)


BenJamesBen:
- For Titan, the code change also moves `points.crotchSeamCurveStart` down to `points.styleWaistIn`, for situations where the customer has requested that the waist be lower than where the crotch seam curve would normally start. This change was made for Charlie bug #2595 , with the hope that it either fixes the bug entirely or that it would be the start of work on the fix. (I have not yet tested Charlie with this Titan code change.)

bobgeorgethe3rd:
> * For Titan, the code change also moves `points.crotchSeamCurveStart` down to `points.styleWaistIn`, for situations where the customer has requested that the waist be lower than where the crotch seam curve would normally start. This change was made for Charlie bug [Error while drafting charlie #2595](https://github.com/freesewing/freesewing/issues/2595) , with the hope that it either fixes the bug entirely or that it would be the start of work on the fix. (I have not yet tested Charlie with this Titan code change.)

I'm just checking. Did you check with the user about their crossSeamFront measure like I suggested? Because I don't believe the changes to Titan are warranted if they are because of #2595

BenJamesBen:
No, I did not ask the specific customer who filed the bug. I don't think I have access to that customer's contact information?

In any case, there will be other instances where the customer is unavailable to ask or where they might re-iterate that the measurement is correct. The code change is intended to address those situations, instead of having the pattern just crash with an Error.

(The code change will print a _"Forced to shorten front crotch length to accommodate waist height."_ Warning message to the customer, along with the _"Manual fitting/alteration of the Front pattern piece may be needed."_ and _"Please recheck your measurements, make a test garment to check fit, and alter if necessary."_ Warning messages. So, the customer will have been advised to recheck their measurements. If needed, we can add an additional Warning message advising them to specifically recheck crossSeamFront.)

BenJamesBen:
Closing this draft/WIP PR. 

- The Breanna change was filed as a separate PR. 
- Titan guideilnes were removed, making those code changes no longer needed. 
- Titan and Bruce runaway rotation protection will be addressed in separate bugs/PRs if needed.
- Titan crotch seam curve adjustment will be addressed in a separate bug/PR if needed.

PR #2712, opened 2022-09-02
Fix v3 ForPrinting and ForCutting layouts to use design shortname.

BenJamesBen:
Before the fix:
Screen Shot 2022-09-01 at 10 46 56 PM

After the fix:
Screen Shot 2022-09-01 at 10 56 33 PM


BenJamesBen:
Added ForCutting to this PR. (I don't usually look at the ForCutting page, so I had not realized that the same fix was needed there.)

joostdecock (approved):

PR #2721, opened 2022-09-03
Port Benjamin to v3.

BenJamesBen:
Retry of Benjamin port to v3.

(I also have Bent, Breanna, and Titan ports prepared, but I want to make sure that this PR goes smoothly before continuing with the others.)


joostdecock:
@BenJamesBen ok, I think I've figured out why we seem unable to preserve history for all but a few lines.

It's because we are nesting the draft method under the part object, thereby changing the indentation. So we are, in effect, changing all lines.
The few lines that are preserved seem to be dumb luck as they are lines that were further indented and happen to be identical to another line in the file.

So, I have side-stepped this problem in thus PR for lucy: https://github.com/freesewing/freesewing/pull/2722

Essentially, I've kept the draft method body as-is, only changing it so it's no longer the default export. Then add the part definition at the bottom of the file and referencing the draft method defined at the top if the file, rather than include it inline.

You can confirm with the lucy PR linked above.

I suggest we use this approach to migrate existing patterns. I might attempt to cherry-check together a new branch that avoids sone of the earlier commits that caused a loss if commit history. But for those looking to port patterns, this approach should give you peace of mind.

BenJamesBen:
I moved the draft code back into the draft function, committed, and re-pushed.

If you would prefer that I start over with a clean branch to avoid the large changes between the 2 commits that caused and fixed the problem, please close this PR. I will be happy to re-submit the PR with a clean branch and a single, minimal commit.

BenJamesBen:
Closing this PR. I will resubmit a new PR with a cleaner commit history.

PR #2732, opened 2022-09-04
Port Carlton to v3.

BenJamesBen:


BenJamesBen:
I am unsure about the location of measurements and options for this Carlton port, which parts they should go in. There also seemed to be a number of redundancies where inherited Bent/Brian measurements/options might not need to be specified in Carlton.

BenJamesBen:
I am closing this PR temporarily. I believe I might have had a fundamental misunderstanding about how dependencies work (assuming that `from:` automatically implies `after:`) If so, then I might have incorrectly omitted some `after:`s in my ports. I will re-open the PR once things are settled.

BenJamesBen:
Re-opening this PR, now that I know that I have been doing things correctly with my `after:` properties.

(For all the designs that I ported to v3, including this Carlton, before submitting the PRs I have been testing them in the localhost lab to verify that they work, comparing them to lab.freesewing.dev to make sure the same patterns are being produced, etc.)

PR #2733, opened 2022-09-04
[wip]: Issues porting Carlita to v3.

BenJamesBen:
This is a draft PR. I encountered 2 issues when trying to port Carlita to v3:

1. I am unable to remove the inherited DraftForHighBust option from the menu. 
   - The design still works properly because we load the bustPlugin directly. However, the (now no-op) option remains in the options menu.
   - Also, loading the bustPlugin directly causes the following rude message in the Events log: `Plugin @freesewing/plugin-bust was requested conditionally, but is already loaded explicitly. Not loading bitch.`
3. If I try to inherit topSleeve and underSleeve directly from Carlton in index.mjs, the brian.front part is no longer hidden. 
   - I am not sure whether this is a bug. In carlton.topsleeve and carlton.undersleeve, `hideDependencies: true,` is set, so any inherited parts should not appear, as I understand it.
   - When the issue occurs in the lab, if you view Carlton and then switch back to Carlita, the brian.front part disappears. So, something is happening dynamically to cause the `hideDependencies` to be obeyed.
   - A workaround is to explicitly define carlita.topsleeve and carlita.undersleeve parts, just to re-iterate `hideDependencies: true,`. I've created `topsleeve.mjs` and 'undersleeve.mjs` files and commented code in `index.mjs` to show the workaround.


BenJamesBen:
I was unable to workaround either issue by setting the `draftForHighBust` option in `front.mjs`. The conditional bustPlugin did not activate (producing an error), and the draftForHighBust option did not hide itself in the menu.

```
  options: {
     draftForHighBust: { bool: true, hide: () => true },
   },
```

BenJamesBen:
I believe I might have had a fundamental misunderstanding about how dependencies work (assuming that `from:` automatically implies `after:`) If so, then I might have incorrectly omitted some `after:`s in my ports. 

I will close and re-submit this PR once things are settled.

joostdecock:
Oops 🙈😂 that rud comment is on me. I got a little frustrated while working on some plugin loading edge-case and I wanted to ensure it was using my updated code so I changed the message temporarily and forgot to change it back, I'll fix that.

As for the rest, there's still some edge cases where we need to see what is the best way to handle inheritance.

I wouldn't worry about too much at this stage. We need to port everything so we can run tests on it. Then I will use those edge cases that give us grief to write new unit tests that replicate the problem and address it.

So if it's showing an option that shouldn't be there or a part, make a note of it and move on. We'll catch it on the next pass

joostdecock:
Removed the debug statement

BenJamesBen:
I am now closing this draft PR.

I filed #2742 to track the options inheritance issues. 

Because switching from one design to another fixes the brian.front-not-hidden issue, I will assume that this is another instance of the patterns-change-state issue tracked by #2736 .

 I will start with a clean branch and re-submit this PR without all the extra workaround files and code.

PR #2734, opened 2022-09-05
Port Cathrin to v3.

BenJamesBen:
This v3 port of Cathrin imports `pluginBundle` 6 times, in the 6 panel files. All 6 files inherit from `panels.mjs`, so I could have imported the plugin just once, which seems more elegant. However, `panels.mjs` doesn't actually use or need `pluginBundle`, which is why I put the imports in the 6 other files.

If you would prefer having just the 1 import, please let me know and I'll make the change.


BenJamesBen:
I am closing this PR temporarily. I believe I might have had a fundamental misunderstanding about how dependencies work (assuming that `from:` automatically implies `after:`) If so, then I might have incorrectly omitted some `after:`s in my ports. I will re-open the PR once things are settled.

BenJamesBen:
Re-opening now that I know there is no `after:` problem.

PR #2735, opened 2022-09-05
Charlie port to v3

BenJamesBen:
I am seeing strange issues with Charlie, that I thought was related to dependency issues. However, now I am not so sure.

Sometimes when i load a fresh Charlie pattern, it incorrectly looks like this:
Screen Shot 2022-09-04 at 7 52 04 PM

But, then I refresh the web page, the pattern changes to something that looks more correct:
Screen Shot 2022-09-04 at 7 52 22 PM

And, when I look in the Events log, I see measurements like this:
Screen Shot 2022-09-04 at 7 57 07 PM

But, when I refresh the web page, the numbers change to:
Screen Shot 2022-09-04 at 7 57 17 PM

This is all using the same measurements. 

When the bug state occurs and I refresh the page to get the correct pattern, when I switch to the Events page and then back to the DraftDesign page, the pattern reverts back to the initial bad pattern.


joostdecock:
These are bugs in the lab. I suspect they were introduced when Enoch was working on the layout.

Essentially, it doesn't reset the state when you switch to a new design. Reloading the page will force the reset.

BenJamesBen:
Changing from draft to actual PR. I assume that because one of the pattern states is the correct pattern, that the Charlie port is correct.

I removed the redundant `after:`s that I initially added in `back.mjs` for investigation. (Specifically, I removed titanBack and titanFront. The back already has `from: titanBack` and `after: front` and the front is already `from: titanFront`. So, titanBack and titanFront aren't needed in back's `after:`.)

I filed #2736 to track the lab pattern state issue.

PR #2759, opened 2022-09-06
Move snapseries to cibuild_step1.

BenJamesBen:
This PR changes the `snapseries` package from cibuild_step6 to cibuild_step1.

We are doing this because Titan depends on `snapseries` during step 3, and the 'yarn cleanall; yarn build` sequence currently fails due to the dependency.
```
lerna ERR! yarn run cibuild_step3 exited 1 in '@freesewing/titan'
...
 ✘ [ERROR] Could not resolve "@freesewing/snapseries"
```


joostdecock:
@BenJamesBen you need to change `config/build-order.mjs` because the entire `package.json` is auto-generated 🤷

BenJamesBen:
Closing. I will resubmit from a clean branch.

PR #2761, opened 2022-09-07
[wip] Initial v3 pattern test changes.

BenJamesBen:
Some initial work on pattern tests.

(I'm not sure whether functions are acceptable for option menus. The ones for Octoplushy currently aren't working properly in the lab.)

The other currently-failing designs are Noble, Penelope, Simon, and Simone.

### DIFF
--- a/config/software/designs.json
+++ b/config/software/designs.json
@@ -282,7 +282,7 @@
       "code": "Rika Tamaike",
       "design": "Rika Tamaike",
       "difficulty": 1,
-      "tags": [ "tops", "historic" ]
+      "tags": [ "tops", "historical" ]
     },
     "ursula": {
       "description":  "A FreeSewing pattern for a basic, highly-customizable underwear pattern",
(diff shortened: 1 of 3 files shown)


BenJamesBen:
I need to revert my "If it has a 'menu' property, it should be a string or function" change. Instead, the fix is to change the design code, moving the function into the option `hide:` property, the way Simon does it.

```js
    menu: 'style',
    hide: ({ options }) => (options.type == 'octoplushy'),
```

The two designs that need this change are Noble and Octoplushy.

joostdecock:
We currently have the `hide` property on options. It's not used very often, but it can hold a method that will receive the draft settings and should return a truthy/falsy value to indicate whether the option should be shown in the UI (true) or not (false).

Which overlaps with the new `menu` property on options that, when set to false also indicates that this option should be hidden.

So, fir those designs that use a `hide()` method I've used that as the value for `menu`. But when it should be shown, rather than returning `true` it should return where it should be shown (eg: `style`).

I haven't implemented this in the UI yet, but that's the idea

joostdecock:
@BenJamesBen Note that I've fixed the issue that was causing the tests to fail on Simon. And because of that, Simone is ok too.

joostdecock:
v3 is merged into develop now. This PR was auto-closed since I removed the `v3` branch to avoid confusion.

PR #2778, opened 2022-09-09
feat(lunetius): add linedrawing

### LINKED ISSUE #2724: [docs]: Add Lunetius line drawing to the site

https://freesewing.org/designs/lunetius/

Yes, I am a tier-4 patron ❤️  💙

### Suggested content

Since Lunetius doesn't currently have a line drawing, I have created the attached .svg as a suggested version.

lunetius

### Additional context

_No response_

starfetch:
Add linedrawing for Lunetius by the fantastic @nataliasayang. Closes #2724 .
I'm not sure if some other file somewhere needs to be pointed to the `.svg`, looking at https://github.com/freesewing/freesewing/commit/cde3fb1709815b01f0f1938fef3a3fd2d4b05083 for reference but the file in question doesn't exist anymore.

### DIFF
--- a/artwork/linedrawings/lunetius.svg
+++ b/artwork/linedrawings/lunetius.svg
@@ -1,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="375.99097mm"
+   height="754.86261mm"
+   viewBox="0 0 375.99096 754.86261"
+   version="1.1"
+   id="svg882"
+   inkscape:version="1.1 (c68e22c387, 2021-05-23)"
+   sodipodi:docname="lunetius.svg"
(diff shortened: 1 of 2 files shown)


eriese (approved):

joostdecock:
FYI: This won't actually resolve #2724 
The way the linedrawings are handles in v3 is a bit convoluted. It's efficient in the way it's rendered, but a bit of a pain in the way it's maintained.

To backport this to v2, this is the file to update: https://github.com/freesewing/freesewing/blob/v2/packages/components/src/LineDrawing/patterns.js
To apply this in v3, well we don't have linedrawings yet in v3 :)

This is also why this issue/PR was left open for so long. It's not relevant to v3 yet, and I'm not too bothered to backport this to v2.

That being said, this merge doesn't do any harm either

eriese:
Well, given that I think this doesn't meet our criteria for urgency for backporting to v2, I think this does actually resolve #2724 because now the drawing will be in the expected place to be handled properly when we do add linedrawings to v3

PR #2781, opened 2022-09-10
fix(noble): Add missing "raise" from shorthand.

BenJamesBen:
These two files use the `raise()` method later in the code, but they never pulled it out of shorthand.


BenJamesBen:
The code changes were made via 7ffba2f81d242248c1fae6dc6ed861743e824f79 , so this PR is no longer needed. Closing.

PR #2786, opened 2022-09-10
Clarify the meaning of 'binding'

jdhoek:
The section on breast binding confused me at first due to the generic nature of the word 'binding'. This commit attempts to clarify this point to make it more readily understandable for a broader audience — hopefully without introducing exclusive or unsuitable language.

### DIFF
--- a/markdown/org/docs/faq/breasts/en.md
+++ b/markdown/org/docs/faq/breasts/en.md
@@ -27,7 +27,7 @@ Any user can draft any pattern on our site, as long as they provide the
 required measurements. Which for some patterns includes measurements
 that are only relevant if you have breasts.
 
-##### How it applies to binding
+##### How it applies to chest binding
 
-If you are binding, use the **without breasts** patterns, and use the cirumference
-of your binded chest as [chest measurement](/docs/measurements/chest/)
+If you are chest binding — effectively flattening your chest — use the **without breasts** patterns,
+and use the cirumference of your bound chest as [chest measurement](/docs/measurements/chest/)


joostdecock (approved):
LGTM but going to run this by our trans ambassador

SeaZeeZee (changes requested):
I think that adding 'breast' to the word binding is slightly redundant. I think however, I am not the kind of person to be confused. I think perhaps changing the word breasts to chest would read better though, since that is how it is generally used. 'Sally uses a binder to bind their chest.'

joostdecock:
I have pinged our trans ambassador for a review of this PR. I think we can put a pin in this until that happens.

jdhoek:
> I think that adding 'breast' to the word binding is slightly redundant. I think however, I am not the kind of person to be confused. I think perhaps changing the word breasts to chest would read better though, since that is how it is generally used. 'Sally uses a binder to bind their chest.'

Amended.

The confusion rises partly from the similar header above ('How it applies to patterns'); I was wondering what a 'binding' was in terms of sewing terminology primed by the 'patterns' above.

KaerMorhan:
I do agree that using chest binding instead of breast binding might be a better fit. While breast binding is a phrase used by Wikipedia, most sources made by and aimed at transmasculine people use **chest binding**. Alternatively, both terms can be used as well - _if you are chest/breast binding_ - as cis girls/women also bind for different reasons (cosplay, comfort and aesthetics being some of them) and might use the phrase **breast binding** instead.

I am also not sure about _bound chest_ as I didn't encounter this phrase being used (usually it's _flattened_ or _compressed chest_, the former being more common) - might be useful to ask a native English speaker about this one.

Other than that, changes look good.

SeaZeeZee:
Im also a non native speaker but have heard bound chest a lot. Language is fun! I can check with my US partner when they're awake.

joostdecock (approved):

joostdecock:
Tx @KaerMorhan , merging

joostdecock:
@all-contributors please add @jdhoek for docs

PR #2790, opened 2022-09-10
fix(ursula): Add box around elastic text, and remove dimensions.

BenJamesBen:
This PR adds a Path box around the elastic text and removes the placeholder dimensions for that part.

- Before the PR, the elastic text is cut off in the DraftDesign view. After the PR, the box causes the entire text to show up.
- Before the PR, the placeholder dimension allows the cut-off elastic text to be seen in the LayoutThing: ForPrinting view, if Paperless is enabled. If Paperless is disabled, the part doesn't show up at all. After the PR, the box causes the full text to always show up in LayoutThing: ForPrinting without the need for the placeholder dimension.

(The below screenshots also include the Ursula plugin fix PR.)

Before the PR:
Screen Shot 2022-09-10 at 9 51 59 AM
Screen Shot 2022-09-10 at 9 52 06 AM
Screen Shot 2022-09-10 at 9 52 28 AM
Screen Shot 2022-09-10 at 9 52 21 AM

After the PR:
Screen Shot 2022-09-10 at 9 49 40 AM
Screen Shot 2022-09-10 at 9 49 30 AM
Screen Shot 2022-09-10 at 9 49 50 AM
Screen Shot 2022-09-10 at 9 49 58 AM


designs/ursula/src/elastic.mjs
```diff
@@ -32,14 +34,13 @@ function ursulaElastic({
...
-      y: points.elasticInfo.y + sa + 15,
-    })
-  }
+  // Draw a box around the text, so the part shows up correctly.
+  paths.box = new Path()
+    .move(new Point(-10,-10))
+    .line(new Point(-10,15))
+    .line(new Point(200,15))
+    .line(new Point(200,-10))
+    .close()
```
joostdecock, line 43: Would it make sense to add a `.setClass('hidden')` here to avoid confusing the user who might assume these lines are meaningful?

It will have the same effect (taken on board for calculating the bounding box of the part) but will be hidden from the output.

joostdecock (changes requested):

designs/ursula/src/elastic.mjs
BenJamesBen: I added the `.setClass('hidden')` as suggested.

joostdecock (approved):

PR #2800, opened 2022-09-11
chore(tutorial): Port to v3 stage 2

BenJamesBen:
A PR back into the v3-stage2 branch.

There is a `step8.mjs` file that doesn't appear to be used anywhere in the design. I made v3 stage 2 porting changes to this file and included it in the PR, but I am wondering whether it should be deleted?


joostdecock (approved):

joostdecock:
> There is a step8.mjs file that doesn't appear to be used anywhere in the design. I made v3 stage 2 porting changes to this file and included it in the PR, but I am wondering whether it should be deleted?

Well spotted, I've removed it.

PR #2813, opened 2022-09-13
Add support for windows-specific build and prebuild scripts

eriese:
I know we technically don't support windows, but I've added the capability to at least smooth the experience. most of our node scripts don't require anything different for windows vs. unix, but currently `build` in `packages/new-design` is blocking windows users.

This PR allows us to add windows-specific `build` and `prebuild` scripts to `config/scripts/yaml` where needed, as `wbuild` and `prewbuild` respectively. 

When `reconfigure` writes the individual `package.json` files, it checks for these scripts and if there aren't any it defaults to the regular `build` and `prebuild`s. It then writes a `wcibuild_stepx` script accordingly, and adds `kickstart:windows` and `wbuildall` to the root `package.json`, so that windows users can run those command instead of having to overwrite the default scripts and then remember to keep their personal hacks out of the working tree.

My thinking is that when a script is changed and it's easy to tell what needs to happen differently for windows, we change it for windows as well. Where it's not easy, it'll break, which has always been the windows user experience. But at least after troubleshooting it, we can fix it for all windows users in one place, thereby fielding fewer support requests for windows.


eriese:
@anna-puk best I can do is tag you here

anna-puk (changes requested):
I found two small fixes, and with that, it runs!

I'm hoping we can get this approved; it would make it much easier for me to create pull requests. (Now I need to separate my own changes from the windows-workaround ones, and then I cannot test anymore.)

We should probably also update the instructions, so Windows users know that their command is 'yarn kickstart:windows'.

eriese:
> We should probably also update the instructions, so Windows users know that their command is 'yarn kickstart:windows'.

Good point. Are those in strapi?

anna-puk:
I pulled the new changes and can confirm that it works for me!

anna-puk (approved):
Based on running the code (rather than reading it): it works.

joostdecock:
> I know we technically don't support windows, but ...

:trollface:

Issue #2657, opened 2022-08-24
Add linters for JS and YAML files


nchilada:
Ideally run this in both [the pre-commit hook](https://github.com/freesewing/freesewing/blob/bae39f714b9f7cc8f9b209a3e438a14ae45de2b0/package.json#L49) (for convenience) and in CI (as a backstop).

joostdecock:
Started doing some work on this in the `eslint` branch: https://github.com/freesewing/freesewing/tree/eslint
For now, focus is on the core library, but I plan to extend it beyond that.

joostdecock:
I made an effort to add it initially as I was doing dome major refactoring of our core library and wanted to include the linting step.

But yeah, it needs some fine-tuning as you point out.
The reason things are at a design/olugin/package level is because it's all auto-generated from `config/templates/eslintrc.yml`

Happy to get some help with this as I'm currently writing v3 docs

nchilada:
Okay, thanks for the background on that! I'll push the centralized `.eslintrc` to my PR so you can judge how it looks.

bobgeorgethe3rd:
^ Has this been solved by the above merged PR request? If so can it be closed?

nchilada:
> ^ Has this been solved by the above merged PR request? If so can it be closed?

Mostly done. I think we can close this issue.

But I should note that there is a gap in coverage, IIRC: eslint configs in lerna workspaces end up shadowing the root eslint config, which ultimately makes the linter excessively lenient in those workspaces. IMO it would be  less confusing if we had a _single_ eslint config (the one at the root). I was working on this but I couldn't keep up with @joostdecock and @eriese's rate of development in the codebase…maybe the correct approach is to get that config change into mainline even if it (temporarily) causes the lint.all.yml action to fail again.

joostdecock:
> Mostly done. I think we can close this issue.

Let's :)

PR #2839, opened 2022-09-21
chore(eslint): fine-tune eslint and fix most errors

nchilada:
In service of #2657, the monorepo now has a pre-commit hook that runs eslint on staged files, but there's still a bit more work to be done.

In this PR:

* Special handling many common files that would currently fail the lint check if/when they are changed and staged for commit:
  * JSON files
  * JS files containing JSON module imports (`import ... assert {type: 'json'}`
  * Markdown files, with JS linting for the embedded code examples
* Make a distinction between frontend files (which may use Web APIs, React/Next globals, JSX syntax, etc.) and backend files (which may use Node APIs)
* Ignore some files: `dist`, build scripts, etc.
* Add a linting job to CI, for all files changed in a given pull request. (We already have a job to assess lint on all files, but it's currently failing and seems likely to fail at times in the future too.) 
* Remove the legacy `.eslintrc` files that had been specified in a subset of the codebase.

Future work?

* Get all files to pass eslint and prettier, to satisfy the all-files check in CI and to reduce churn in future pull requests


.eslintrc.cjs
```diff
@@ -0,0 +1,114 @@
...
+  'sites/*/themes/**',
+]
+const frontendFiles = [
+  '**/components/**',
+  '**/hooks/**',
+  '**/pages/**',
+  '**/page-templates/**',
+  'packages/i18n/**/*.md/*.js',
+]
+const isomorphicFiles = ['packages/core/**', 'plugins/**', 'tests/designs/**', 'tests/plugins/**']
```
nchilada, line 20: Based on the [distinction](https://freesewing.dev/guides/plugins/types-of-plugins) between build-time and run-time plugins, I gathered that plugin code should be able to run in either Node.js or a web browser, and therefore ought to restrict itself to the intersection of Node builtins and browser builtins. (And I'm guessing the same is true for some of the core code.) Ironically, however, the only plugin that [seems to use](https://github.com/freesewing/freesewing/blob/a44c23c270546a2229da2a583ea33405e7bcaaab/plugins/plugin-timing/src/index.mjs#L4-L11) any such builtins actually uses items from the _union_ of Node builtins and browser builtins, and will therefore need an `eslint-ignore` directive in order to comply with the lint rules. So maybe I'm over-engineering here.

nchilada, line 20: Based on the [distinction](https://freesewing.dev/guides/plugins/types-of-plugins) between build-time and run-time plugins, I gathered that plugin code should be able to run in either Node.js or a web browser and therefore ought to restrict itself to the _intersection_ of Node builtins and browser builtins. (I'm guessing the same is true for some of the core code, but I'm even less sure about that.)

Ironically, however, the only plugin that [seems to use](https://github.com/freesewing/freesewing/blob/a44c23c270546a2229da2a583ea33405e7bcaaab/plugins/plugin-timing/src/index.mjs#L4-L11) any such builtins actually uses items from the _union_ of Node builtins and browser builtins, and will therefore need an `eslint-ignore` directive in order to comply with the lint rules.

So maybe I'm over-engineering here.

.eslintrc.cjs
```diff
@@ -0,0 +1,114 @@
...
+        'no-console': 'off',
+      },
+    },
+    {
+      files: frontendFiles,
+      excludedFiles: nodeFiles,
+      extends: ['next/core-web-vitals'],
+      env: {
+        // We can be stricter than 'next/core-web-vitals' is
+        node: false,
```
nchilada, line 56: The `next/core-web-vitals` eslint config, which is recommended for use throughout a Next.js project, specifies `node: true, browser: true` and therefore technically allows JavaScript to access Node and browser builtins, even where that might not make sense. I'm trying to be stricter than that, here, but I may be over-engineering again.

.eslintrc.cjs
```diff
@@ -0,0 +1,114 @@
...
+    {
+      files: isomorphicFiles,
+      excludedFiles: [].concat(nodeFiles, frontendFiles),
+      env: {
+        'shared-node-browser': true,
+      },
+    },
+    // Additional globals for JavaScript files that happen to be CommonJS.
+    // Only allowed in *.cjs files, not *.js files, because we probably want to move towards
+    // a `"type": "module"` future where any CommonJS files would have to have .cjs extensions.
```
nchilada, line 68: More background [here](https://nodejs.org/api/packages.html#determining-module-system), though it doesn't directly explain why one might prefer ESM.

.eslintrc.cjs
```diff
@@ -0,0 +1,114 @@
...
+      },
+    },
+    {
+      files: isomorphicFiles,
+      excludedFiles: [].concat(nodeFiles, frontendFiles),
+      env: {
+        'shared-node-browser': true,
+      },
+    },
+    // Additional globals for JavaScript files that happen to be CommonJS.
```
nchilada, line 66: `require`,` `module`, `exports`

.eslintrc.cjs
```diff
@@ -0,0 +1,114 @@
...
+  },
+  // Required when using experimental EcmaScript features
+  parser: '@babel/eslint-parser',
+  parserOptions: {
+    ecmaVersion: 'latest',
+    sourceType: 'module',
+    // Options specific to the Babel parser
+    requireConfigFile: false,
+    babelOptions: {
+      plugins: ['@babel/plugin-syntax-import-assertions'],
```
nchilada, line 35: This fixes lint for `import ... assert { type: 'json' }`

.eslintrc.cjs
nchilada: The bottom line, for developers in this repo: whenever we touch a `*.js`, we'll either need to convert it to ESM in order to lint successfully (and rename it as `*.mjs` if it's in a non-module package, in order to satisfy Node.js), or we'll need to rename it as `*.cjs` (e.g. the `tailwind.config.cjs` for the recalcitrant Tailwind config that @joostdecock mentioned somewhere).

.eslintrc.cjs
```diff
@@ -0,0 +1,114 @@
...
+    // Additional globals for JavaScript files that happen to be CommonJS.
+    // Only allowed in *.cjs files, not *.js files, because we probably want to move towards
+    // a `"type": "module"` future where any CommonJS files would have to have .cjs extensions.
+    {
+      files: ['**/*.cjs'],
+      env: {
+        commonjs: true,
+      },
+    },
+    // Additional globals for JavaScript files that happen to contain Mocha tests
```
nchilada, line 75: `context`, `describe`, `it`

.eslintrc.cjs
```diff
@@ -0,0 +1,114 @@
...
+        node: true,
+      },
+      rules: {
+        'no-console': 'off',
+      },
+    },
+    {
+      files: frontendFiles,
+      excludedFiles: nodeFiles,
+      extends: ['next/core-web-vitals'],
```
nchilada, line 53: This adds support for JSX syntax and also adds a bunch of React- and Next-specific rules

.eslintrc.cjs
```diff
@@ -0,0 +1,114 @@
+const nodeFiles = [
+  '**/build.dflt.{js,mjs,cjs}',
+  '**/build.{js,mjs,cjs}',
+  '**/config/**',
+  '**/prebuild.{js,mjs,cjs}',
+  '**/prebuild/**',
+  '**/scripts/**',
+  'packages/new-design/lib/**',
+  'sites/backend/**',
```
nchilada, line 9: Note that when using a single centralized `.eslintrc`, we sometimes have to list specific directories where we want customized lint rules. An alternative would be to add new files, e.g. `sites/backend/.eslintrc`, that override parts (or all) of `./.eslintrc`.

joostdecock (approved):
:crossed_fingers:

nchilada:
@joostdecock thanks for merging this! I think I was playing with a couple more improvements offline, so I will refresh my memory tomorrow and create a followup PR if applicable

PR #2847, opened 2022-09-22
v2 - fix(holmes) Holmes adds stores.

bobgeorgethe3rd:
Problem:
gore Plugin reduces the circumference of the hat when below 50% meaning that the headCircumference used for the ear flaps and  visor are inaccuarate to the circumference of the crown. This also applies to the length of the ear flaps.

Solution:
Store the measures needed for the ear flaps and visor.

Additions
- added `store.set('circumference')` to holmes/gores.js
- added `store.set('goreLength')` to holmes/gores.js

Changes
- changed 'headCircumference' to 'circumference' and `store.get('circumference')` for holmes/ears.js
- changed 'earFlapLength' value to use `store.get('goreLength')` for holmes/ears.js
- changed 'buttonholeDistance' value to use `store.get('goreLength')` for holmes/ears.js
- changed 'headCircumference' to 'circumference' and `store.get('circumference')` for holmes/visor.js
- changed 'headRadius' to 'radius' for holmes/visor.js

I am sorry this is in v2 and not in v3. I am still having trouble grasping v3 atm but I thought If I could at least get the v2 fix on paper someone else can port this fix to v3.

Edit: Had to rollback a change and in doing so went one commit too far back, fixed now.

### DIFF
--- a/designs/holmes/config/index.js
+++ b/designs/holmes/config/index.js
@@ -24,10 +24,13 @@ export default {
     advanced: ['visorLength'],
   },
   measurements: ['head'],
-  dependencies: {},
+  dependencies: {
+    ear: 'gore',
+    visor: 'gore',
+  },
   inject: {},
   hide: [],
-  parts: ['gore', 'visor', 'ear'],
+  parts: ['gore'],
   options: {
     headEase: {
       pct: 3,
--- a/designs/holmes/src/ear.js
+++ b/designs/holmes/src/ear.js
@@ -17,9 +17,9 @@ export default function (part) {
   } = part.shorthand()
 
   // Design pattern here
-  let headCircumference = store.get('headCircumference')
+  let circumference = store.get('circumference')
   let earFlapLength = store.get('goreLength') * options.earLength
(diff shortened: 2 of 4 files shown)


joostdecock:
Wouldn't this be better handled with the `extraLength` option in the gore plugin?

bobgeorgethe3rd:
> Wouldn't this be better handled with the `extraLength` option in the gore plugin?

The extra length only adds additional or takes away length it is not the whole length of the crown. This is adressing the fact that the gore plugin reduces the crownCircumference when the extra length takes away from the overall crown length which is expected as you go up, the head the circumference would reduce but the earFlaps and visor were not connected to this reduce in circumference which means they would not draft the way intended which in default form is no overlap between the visor and earFlaps (This is width) and no overlap of the earFlaps (This is length). 

It's not an issue with the plugin.

If I misunderstood what you are asking please ask again.

bobgeorgethe3rd:
Oh wait I see what you mean. Yes that can be used I just didn't want to alter the code to much.

PR #2854, opened 2022-09-23
[Fix] [Bug] Measurement Inputs Not Showing

eriese:
The issue: 

1. Measurement inputs don't show
2. The title for Optional Measurements always shows even if the pattern doesn't required any

Steps to reproduce:

1. Go to the measurements view on any pattern
2. If the measurement inputs are showing, refresh the page
3. measurement inputs will not show

The fix:
Because measurements are now in the part configs, `design.config` only has the default config values, so we need to run `new design(gist).__init()` to properly populate the config on the design, even if we don't intend to use the instantiated pattern for anything else

### DIFF
--- a/sites/shared/components/workbench/measurements/index.js
+++ b/sites/shared/components/workbench/measurements/index.js
@@ -1,13 +1,12 @@
-import React, {useMemo, useEffect, useState} from 'react'
+import React, { useMemo, useEffect, useState } from 'react'
 import MeasurementInput from '../inputs/measurement.js'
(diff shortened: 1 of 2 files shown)


joostdecock:
Thanks for the PR.
This was fixed in one of my local branches, which has since been merged.

In general, we should never call `__` prefixed methods. Those are *interenal* and won't be documented.
So whenever you have a situation that requires such a method, please create an issue so we can implement a proper solution.

For this particular issue, the design object now has a `patternConfig` properties that hold what you need.

PR #2855, opened 2022-09-25
[Fix] remove unused vars blocking lab and dev builds

eriese:
Stricter linting was meaning that unused vars were failing the ci builds for lab and dev. I have removed all the unused vars that were pointed out by the linter.

All other apparent changes are caused by prettier having been run on the files that vars were removed from

### DIFF
--- a/sites/dev/components/footer.js
+++ b/sites/dev/components/footer.js
@@ -14,81 +14,88 @@ import RedditIcon from 'shared/components/icons/reddit.js'
 import TwitterIcon from 'shared/components/icons/twitter.js'
 
 // Classes
-const link = "text-secondary font-bold hover:pointer hover:underline px-1"
-const accent = "text-accent font-bold text-lg px-1 block sm:inline"
-const freesewing = "px-1 text-lg font-bold block sm:inline"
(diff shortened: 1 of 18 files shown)


joostdecock:
Thanks for this @eriese :pray: 
I've merged some of my more recent work, so it's probably a good idea to rebase this. But obviously it's something that needs to go in.

joostdecock (commented):
Could you look into why you are (I assume automatically) converting single quotes to double quotes?

Because the code style enforced by prettier is single quotes. So we should look into what on your side is converting this to double quotes because we're going to end up with this perpetual warfare of the quotes if you keep on changing everything to single quotes, and then prettier keeps on changing everything back.

eriese:
@joostdecock It's a prettier config issue. Single quotes are being properly enforced in regular js, but by default prettier enforces double quotes in jsx. So if we leave it, there won't be a war because prettier is responsible for both types of changes. But if we want to have single quotes in jsx as well, we'll need to add that rule to our prettier config

joostdecock:
Oh ik, my bad, I did not know that 🤷
Merging

PR #2866, opened 2022-09-26
Remove grow CSS from tabs to prevent them from being obscured.

BenJamesBen:
This change removes the "grow" CSS class from tabs to prevent them from expanding. For the tabs in the Logs view, this prevents the "Set 0" tab from being obscured by the View menu.

(I simply removed the " grow" text. The rest of the code formatting was done by eslint.)

Before:
Screen Shot 2022-09-26 at 9 17 55 AM

After:
Screen Shot 2022-09-26 at 9 18 20 AM

### DIFF
--- a/sites/shared/components/mdx/tabs.js
+++ b/sites/shared/components/mdx/tabs.js
@@ -19,7 +19,9 @@ export const Tabs = ({ tabs = '', active = 0, children }) => {
         {tablist.map((title, tabId) => (
           <button
             key={tabId}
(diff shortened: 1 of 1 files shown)


joostdecock:
I appreciate the effort, but I'm not going to merge this because:

- It's a workaround that does keep the tab titles visible, but would not do anything to contain the log lines within the available space
- It would also impact the Tabs component in MDX and I think I prefer the grrown tabs style
- It is already fixed in the log view

Issue #2844, opened 2022-09-21
[bug]: [v3] round macro paths are no longer hidden by default, affecting existing Designs


BenJamesBen:
### What seems to be the problem? 🤔

In localhost lab, the Bob design is now unexpectedly and incorrectly showing a new path:
![Screen Shot 2022-09-20 at 11 42 56 PM](https://user-images.githubusercontent.com/109869956/191435117-d48e6146-bf31-42c8-8b86-2198ef09102f.png)

This path is created via two `round` macros in `bib.mjs` (the ones named 'tipRightTop' and 'tipRightBottom').

The cause is a change in the default behavior of the `round` macro. Via commit 5dc740130894546ffbed98fbd313215e8cc4ee0d, the `round` macro was changed such that the prefix+'Rounded' paths created by the macro are not hidden by default. Previously these paths were hidden (not rendered) by default.
```js
--- a/plugins/plugin-round/src/index.mjs
+++ b/plugins/plugin-round/src/index.mjs
@@ -31,9 +31,8 @@ export const plugin = {
           this.points[prefix + 'End']
         )
         .attr('class', so.class ? so.class : '')
-      if (typeof so.render !== 'undefined' && so.render)
-        this.paths[prefix + 'Rounded'].render = true
-      else this.paths[prefix + 'Rounded'].render = false
+      if (so?.hidden) this.paths[prefix + 'Rounded'].hide()
+      else this.paths[prefix + 'Rounded'].unhide()
     },
   },
 }
```

### Design / Plugin / Package 🧐

plugins/plugin-round

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

It might be that this will be deemed a 'Will Not Fix' bug, with the remedy being to change existing Designs to conform to the new API (by adding `hidden: true` to the macro calls):
```js
    macro('round', {
      from: points.edgeTop,
      to: points.tipRight,
      via: points.tipRightTop,
      prefix: 'tipRightTop',
      hidden: true,
    })
```

joostdecock:
You are correct, we should not change the default behaviour. Fixed in https://github.com/freesewing/freesewing/commit/71904f240c188c91af39b954ef8a23af91869aa8

PR #2871, opened 2022-09-27
fix(benjamin): Remove unused, inherited elements; hide round macro paths; fix ribbon dependency.

BenJamesBen:
1. https://github.com/freesewing/freesewing/issues/2844 causes the `round` macro paths to be visible by default. This PR adds the explicit `hidden: true` to them.
2. I missed adding the `after: base` dependency in `ribbon.mjs` when I ported the Benjamin to v3. This PR fixes it.
3. The new v3 changes somehow change how parts are injected/inherited such that paths and snippets from `base.mjs` are displayed for `bow2` and `bow3`, even if those parts are hidden. This PR deletes inherited Points, Paths, and Snippets when the part is hidden. 
Screen Shot 2022-09-27 at 9 08 01 AM
Screen Shot 2022-09-27 at 9 08 26 AM
Screen Shot 2022-09-27 at 9 08 36 AM

### DIFF
--- a/designs/benjamin/src/base.mjs
+++ b/designs/benjamin/src/base.mjs
@@ -67,9 +67,9 @@ function draftBenjaminBase({
   // Paths
   paths.cap = new Path().hide().move(points.tip2Bottom)
   if (options.endStyle === 'straight') {
-    paths.cap = new Path().move(points.tip2Bottom).line(points.tip2Top)
+    paths.cap = new Path().move(points.tip2Bottom).line(points.tip2Top).hide()
   } else if (options.endStyle === 'pointed') {
-    paths.cap = new Path().move(points.tip2Bottom).line(points.tip).line(points.tip2Top)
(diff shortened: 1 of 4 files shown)


BenJamesBen:
This Draft PR contains some fixes for Benjamin. However, I'm not sure whether the inject/inherit and round macro issues might be fixed in core, making the fixes unnecessary.

joostdecock:
Could you double-check with the latest changes in the `develop` branch? 
Benjamin looks fine to me, but it's not a design I'm very familiar with.

BenJamesBen:
All of the Benjamin issues in this bug are no longer reproducible, fixed by the code changes that were committed yesterday/today. Closing.

BenJamesBen:
Actually, the `ribbon.mjs` dependency issue is still present. I'll submit a fresh PR for it.

PR #2895, opened 2022-09-30
chore(plugintest): Convert Title to use v3 "for" store data.

BenJamesBen:


designs/plugintest/src/plugin-title.mjs
```diff
@@ -1,9 +1,10 @@
 import { titlePlugin } from '@freesewing/plugin-title'
 import { base } from './base.mjs'
 
-const pluginTitle = ({ points, Point, paths, Path, macro, options, part }) => {
+const pluginTitle = ({ points, Point, paths, Path, macro, options, store, part }) => {
   if (['title', 'all'].indexOf(options.plugin) !== -1) {
-    if (options.titleMeta) part.context.settings.metadata = { for: 'Some user' }
+    if (options.titleMeta) store.data.for = 'Some user'
```
joostdecock, line 6: Can we use `store.set('data.for', 'Some user')` here?
It's the same under the hood today, but using the method is more future-proof

designs/plugintest/src/plugin-title.mjs
```diff
@@ -1,9 +1,10 @@
...
+    else delete store.data.for
```
joostdecock, line 7: Same remark here, better to use `store.remove('data.for')`

joostdecock (changes requested):

designs/plugintest/src/plugin-title.mjs
BenJamesBen: I made the change to `store.set('data.for', 'Some user')`.

BenJamesBen:
Hmm... my attempts to update the branch brought in too many other files. Closing and re-submitting.

PR #2896, opened 2022-09-30
fix(plugintest): Cast options.dimensionD to integer.

BenJamesBen:
Currently (in both lab.freesewing.dev and `develop` localhost lab) changing the dimensionD option causes an error.


BenJamesBen:
I am not sure whether this is actually a bug in the design code. I am now thinking that instead this is a bug in the core infrastructure. 

By adding some debug code, I see that the `typeof()` of `count` options is "string", not "number". Were they intended to be numbers?

BenJamesBen:
Closing this PR. I will file a bug instead.

PR #2925, opened 2022-10-05
fix(Titan) fitGuides points.hipsOut

bobgeorgethe3rd:
Fix for #2924

Problem:
points.hipsOut would sometimes not be able to be set because the line used with utils.lineIntersectsCurve would not past through the outseam.

Solution:
points.hipsOutTarget is now extended out from where it originally drafted using .shiftOutwards and points.hipsInTarget.
points.hipsIn is no longer rotated around points.hipsOutTarget in the utils.lineIntersectsCurve as the line created between it and the new points.hipsOutTarget will cross the outseam curve.

Additional Changes:
I have also changed fitGuides to an optional option in advanced so if another bug is found do to with the fitGuides they can be turned off as a workaround whilst a fixed is made. (Rather than the current which would be using the advanced contents for Paco or Charlie) This has presented an interesting an intersting bug in the v3 lab detailed here https://discord.com/channels/698854858052075530/698862765053575188/1027282797313937408

Easy to backport to v2 if needed,

### DIFF
--- a/designs/titan/src/back.mjs
+++ b/designs/titan/src/back.mjs
@@ -242,9 +242,12 @@ function titanBack({
       points.hipsInTarget = points.waistIn
         .shiftTowards(points.waistOut, measurements.waistToHips)
         .rotate(-90, points.waistIn)
-      points.hipsOutTarget = points.waistOut
-        .shiftTowards(points.waistIn, measurements.waistToHips)
-        .rotate(90, points.waistOut)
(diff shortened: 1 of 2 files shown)


joostdecock:
Please consult with the designer first.
This was addressed in https://github.com/freesewing/freesewing/commit/f2170288751362ed9efd617acba4d553978abe90

PR #2940, opened 2022-10-08
fix(shared): Correct error/warning counts for sets.

BenJamesBen:
This PR fixes an error in 295d3481f9faa813b8162a8cf8e7cd73629b00ae that caused draft/set errors and warnings to be incorrectly counted against patterns:
```
There are 2 errors in the pattern logs
There are 8 warnings in the pattern logs
```

It also normalizes the variable name from `set` to `sets` (assuming that there will be multiple sets in the future, not just Set 0).


joostdecock (approved):
Good catch

PR #2941, opened 2022-10-09
chore(tutorial): Port Pattern Design tutorial text to v3.

BenJamesBen:
A first step at porting the Pattern Design tutorial to v3.

These are just the markdown changes. The image files will also need to be updated later.

Also, I noticed that there is a new `structure/en.md` page, but I did not touch it. It will need to be edited and integrated into the tutorial at some point.

### DIFF
--- a/markdown/dev/tutorials/pattern-design/adding-measurements/en.md
+++ b/markdown/dev/tutorials/pattern-design/adding-measurements/en.md
@@ -41,4 +41,6 @@ For example `38` as 38cm is a realistic head circumference measurement for a bab
 Enter `38` in the box, and click on **Draft Design** in the sidebar under the **View** heading.
 This brings you back to our work in progress:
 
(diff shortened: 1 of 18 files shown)


BenJamesBen:
I converted the old \<Example> tag to new \<Examples>. (I left the `pattern="tutorial"` attribute within the tags, in case Examples is extended in the future to accept different patterns via that attribute.)

Some of the examples (sampling, paperless) do not work as intended, due to #2943 .

joostdecock:
I'm sorry @BenJamesBen I've been traveling have only had intermittent internet access.
During those offline hours though, I've been working on the documentation, including the tutorial.
You can review it here: https://freesewing-psjblphza-freesewing.vercel.app/tutorials/pattern-design

Please let me know what you thing. But I'll have to close this as it's stale.

PR #2944, opened 2022-10-09
chore(docs): Port examples to v3.

BenJamesBen:
Starting work on porting the docs examples to v3. I'm basically just copying the code that was used for the v2 examples.


joostdecock:
Please note that (for example) the pattern config no longer exists. So I'm worried this might be a lot of work not all of it relevant :grimacing:

joostdecock:
Perhaps we should instead bring back the examples design, but remove the api specific stuff that's now handled inline.

joostdecock:
@BenJamesBen I've reverted the removal of the examples desigh: https://github.com/freesewing/freesewing/commit/ed576747d8cc8a2f1bf363863730c19212494445

What I suggest we do is:

- Change the `Example` custom component to `Examples`
- Bring the legend stuff into the examples design so all examples are in 1 design
- Remove those parts from examples that are/can be replaced with an inline Example

As a reminder: The new Example component can handle anything that can be done within a draft method.
What it can't do is handle imports. So no extra plugins, no extra dependencies and so on.

What do you think?

BenJamesBen:
I am fine with whatever way you want to implement this. I was just trying to follow your intent, by guessing what that intent was from looking at the code.

(However, for complicated examples like the diagram showing how patterns work, I did start wondering whether using a static PNG might be easier.)

joostdecock:
> (However, for complicated examples like the diagram showing how patterns work, I did start wondering whether using a static PNG might be easier.)

Easier, sure. But [dogfooding](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) is not not important.

I'm closing this. As all examples have been ported.

PR #2953, opened 2022-10-10
fix(Bee) Options changes restored.

bobgeorgethe3rd:
Have started to work in v3 and noticed that Bee had lost the changes it made to Bella's options when it was ported.
This PR simply restores the changes Bee makes.

Changes:
- Bella options that are constant Bee have been restored
- Bella options that sit in a differenet catergory for Bee have now been moved back to locations.

Note: If there is an easier way to just change the catergory rather than having to rewrite out the option please let me know and I will update the PR.

### DIFF
--- a/designs/bee/src/cup.mjs
+++ b/designs/bee/src/cup.mjs
@@ -9,12 +9,29 @@ export const cup = {
   },
   after: neckTie,
   options: {
+    //Bee's
     topDepth: { pct: 54, min: 50, max: 80, menu: 'fit' },
     bottomCupDepth: { pct: 8, min: 0, max: 20, menu: 'fit' },
     sideDepth: { pct: 20.6, min: 0, max: 30, menu: 'fit' },
     sideCurve: { pct: 0, min: -50, max: 50, menu: 'fit' },
     frontCurve: { pct: 0, min: -50, max: 50, menu: 'fit' },
     bellaGuide: { bool: false, menu: 'fit' },
+    //changed from Bella
+    backArmholeCurvature: 0.63,
+    backArmholePitchDepth: 0.35,
+    backArmholeSlant: 5,
+    frontArmholeCurvature: 0.63,
+    bustDartCurve: 1,
+    bustDartLength: 1,
+    waistDartLength: 1,
+    backHemSlope: 2.5,
+    backNeckCutout: 0.06,
+    //catergory changed from Bella
+    armholeDepth: { pct: 44, min: 38, max: 46, menu: 'advanced.bellaArmhole' },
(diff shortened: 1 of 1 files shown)


designs/bee/src/cup.mjs
```diff
@@ -7,12 +7,29 @@ export const cup = {
...
   options: {
+    //Bee's
     topDepth: { pct: 54, min: 50, max: 80, menu: 'fit' },
     bottomCupDepth: { pct: 8, min: 0, max: 20, menu: 'fit' },
     sideDepth: { pct: 20.6, min: 0, max: 30, menu: 'fit' },
     sideCurve: { pct: 0, min: -50, max: 50, menu: 'fit' },
     frontCurve: { pct: 0, min: -50, max: 50, menu: 'fit' },
     bellaGuide: { bool: false, menu: 'fit' },
+    //changed from Bella
+    backArmholeCurvature: 0.63,
```
joostdecock, line 18: Why does a bikini pattern concern itself with so many options about the back of the block it's based on?

designs/bee/src/cup.mjs
```diff
@@ -7,12 +7,29 @@ export const cup = {
...
+    backArmholePitchDepth: 0.35,
+    backArmholeSlant: 5,
+    frontArmholeCurvature: 0.63,
+    bustDartCurve: 1,
+    bustDartLength: 1,
+    waistDartLength: 1,
+    backHemSlope: 2.5,
+    backNeckCutout: 0.06,
+    //catergory changed from Bella
+    armholeDepth: { pct: 44, min: 38, max: 46, menu: 'advanced.bellaArmhole' },
```
joostdecock, line 28: Options don't belong to a design. They belong to a part. So if you want to have the menu reference that, please use the design name + part name.

designs/bee/src/cup.mjs
bobgeorgethe3rd: It doesn't that's why I locked the ones that do not affect Bee. This was present on v2 but was lost in the port. This is simply relocking them because they do not affect Bee.

designs/bee/src/cup.mjs
bobgeorgethe3rd: This how they were labeled before the port. It's a prefix to the catergories they were in under Bella.

joostdecock:
Can you please fix the linting errors so we can merge this? Tx

bobgeorgethe3rd:
Rebased to fix lint issues

SeaZeeZee:
@joostdecock I think Bob did what you requested so this can be merged now!

SeaZeeZee (approved):
Since this restores things to how they were before I think this is good!

PR #2970, opened 2022-10-13
chore(eslint): Clean up the eslint config, round 2

nchilada:
I'll make all the actual lint fixes in a follow-up PR.

### DIFF
--- a/packages/new-design/templates/from-bella/package.json.mustache
+++ b/packages/new-design/templates/from-bella/package.json.mustache
@@ -70,7 +70,6 @@
     "mocha": "^9.1.1",
     "chai": "^4.2.0",
     "autoprefixer": "^10.4.0",
-    "eslint-config-next": "12.1.6",
     "highlight.js": "^11.5.1",
     "js-yaml": "^4.1.0",
     "pdfkit": "^0.13.0",
--- a/packages/new-design/templates/from-bella/package.json.mustache
+++ b/packages/new-design/templates/from-bella/package.json.mustache
@@ -70,6 +70,7 @@
     "mocha": "^9.1.1",
     "chai": "^4.2.0",
     "autoprefixer": "^10.4.0",
+    "eslint-config-next": "12.1.6",
     "highlight.js": "^11.5.1",
     "js-yaml": "^4.1.0",
     "pdfkit": "^0.13.0",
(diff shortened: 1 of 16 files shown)


packages/new-design/templates/from-bella/package.json.mustache
```diff
@@ -70,7 +70,6 @@
     "mocha": "^9.1.1",
     "chai": "^4.2.0",
     "autoprefixer": "^10.4.0",
-    "eslint-config-next": "12.1.6",
```
joostdecock, line 73: This would remove eslint-config-next from he stand-alone development environment (outside the monorepo.
Is that what we want? Won't it be missed?

(same for the other templates)

joostdecock (approved):

packages/new-design/templates/from-bella/package.json.mustache
nchilada: Thanks for catching this - you know better than me! I remember that standalone is a thing, but I'd forgotten about it and didn't consider it at all in commit c18f05f643f202530026e1c551056bf6b749a549. I'll revert these lines, and I'll also check whether I need to add any other `devDependencies` that I'd only added in the workspace root.

nchilada: On further thought, I'm going to merge this PR now, then reintroduce eslint to standalone environments later. Several reasons for this:

* It looks like `.eslintrc` and, as suggested earlier, associated `eslint-plugin-...` dependencies are required for eslint to work in a truly standalone environment (not a descendent of a monorepo worktree).
* There's still a lot of lint to clean up - I think a beginner would be put off by all the output.
* I predict there will be a lot of churn in `packages/new-design` in the near future. And in part due to things I'm seeing and would like to improve, some of which might rely on conventions introduced in this PR

nchilada: >     * It looks like `.eslintrc` and, as suggested earlier, associated `eslint-plugin-...` dependencies are required for eslint to work in a truly standalone environment (not a descendent of a monorepo worktree).

I was wrong about this! Per the [Next docs](https://nextjs.org/docs/basic-features/eslint), it's possible to use `npx next lint` even when `eslint` isn't a direct (or hoisted) dependency. Though I still think it's important to clean up any lint errors for people who want to use this.

joostdecock (approved):

joostdecock:
> On further thought, I'd like to merge this PR now

As you wish

PR #2971, opened 2022-10-14
refactor(new-design): miscellaneous cleanups

nchilada:
...otherwise known as, rabbit holes that I've entered while looking at the new pattern design tutorial and contemplating where to add `.eslintrc` for standalone environments.


packages/new-design/lib/utils.mjs
```diff
@@ -253,29 +264,44 @@ const showTips = (config, choices) => {
...
   try {
-    await oraPromise(copyTemplate(config, choices), {
+    const templateVars = {
+      Name: capitalize(choices.name),
+      name: choices.name,
+      tag: config.tag,
+    }
+
+    try {
+      await stat(join(monorepoDesignsDir, choices.template))
```
eriese, line 296: @nchilada can you say more about the purpose of this? Is it a check to make sure that the intended inherited design exists so that the block argument can be set?

nchilada:
Amazing to come back from vacation and find that one's old PRs are already rebased and incorporated elsewhere! Thank you @eriese, I'll close this and review the changes in #3118.

It looks like I had marked this as a draft PR shortly before I left…sometimes that means there were other changes I wanted to mark or problems I meant to fix, but probably not in this case since the branches on my dev machine are an exact mirror of what's here on GitHub.

PR #2981, opened 2022-10-18
Fix more broken links & typos, add link to translation guide

starfetch:
There are a few broken links where I could not find where they supposed to go (nothing in the search):

On https://freesewing.dev/reference/api:
`/reference/api/utils/generateStackTransform`
`/reference/api/utils/curveedge`

On https://freesewing.dev/guides/v3:
`/reference/api/attributes/addclass`
`/reference/api/store/remove`
I pointed the `part` links to the `/part/draft` pages, I hope that's correct.

Also, there is no quick start guide, right, @joostdecock? I linked to the tutorials instead.

Additionally, I added a link to the translation guide in the part where we talk about the `en.md` files.

### DIFF
--- a/markdown/dev/guides/v3/en.md
+++ b/markdown/dev/guides/v3/en.md
@@ -180,16 +180,16 @@ The following methods have been added to the core API in v3:
 #### On the `Attributes` object
 
 - [Attributes.addClass](/reference/api/attributes/addclass)
-- [Attributes.asPropIfPrefixIs](/reference/api/attributes/aspropifprefixis)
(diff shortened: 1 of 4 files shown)


joostdecock (approved):

joostdecock:
> There are a few broken links where I could not find where they supposed to go (nothing in the search):

Those methods are no longer used/documented, so those pages are gone. But I forgot to update the list.

> I pointed the part links to the /part/draft pages, I hope that's correct.

That is the best you could with the given situation. But it's a valid point, the way the URL is structured is inconsistent with the rest of the docs. So I've updated this part of the docs.

PR #3019, opened 2022-10-27
Remove duplicated `Added:` in changelog

timorl:
It resulted in `npm run reconfigure` failing, which in turn made `yarn new design` fail in a spectacular and confusing way.


joostdecock (approved):

joostdecock:
Thanks for your contribution @timorl nice catch

@all-contributors please add @timorl for code

PR #3053, opened 2022-11-11
introducing the lily leggings (based on titan)

anna-puk:
The `lily` leggings design is just about ready for pattern testing! 🎉 

There's a manual work-around for now to get all the options at their correct values, but when that's done, it does seem to generate reasonable patterns.


designs/lily/src/back.mjs
```diff
@@ -0,0 +1,567 @@
...
+    'ankle',
+    'heel', // secondary measurement, used instead of ankle
+  ],
+  options: {
+    fitGuides: {bool: true, menu: undefined},
+    fitKnee: {bool: true, menu: undefined},
+    legBalance: 0.5, // between back and front parts
+    waistBalance: 0.5,
+    waistHeight: { ...titanBack.options.waistHeight, pct: 50 }, // halfway between waist and hips
+    fabricStretch: { pct: 40, min: 0, max: 50, menu: 'fit' },
```
BenJamesBen, line 556: fabricStretch seems to affect only the width of the lower leg. The waistband and seat/upper leg stay the same regardless of the amount of fabricStretch. This seems to be intentional, but I am not sure.

designs/lily/src/back.mjs
```diff
@@ -0,0 +1,567 @@
...
+    waistEase: {pct: -4,  menu: undefined}, // -fabricStretch/10,
+    seatEase: {pct: -4, menu: undefined}, // -fabricStretch/10,
+    kneeEase: {pct: -4, menu: undefined}, // -fabricStretch/10,
+    test: {pct: ({ options }) => (options.fabricStretch/2)},
+    lengthBonus: 0,
+    lengthReduction: { pct: 0, min: 0, max: 100, menu: 'style'},
```
BenJamesBen, line 562: 100% seems like a large maximum value for lengthReduction. 

I note that the pattern crashes with an error if Doll cisFemale 10% is used and lengthReduction is 42.4-43.3%. It crashes again with lengthReduction >= 71.5%.

With cisFemale size 28, it crashes with lengthReduction 45.3-45.6% and 73.6%.

designs/lily/src/back.mjs
```diff
@@ -0,0 +1,567 @@
...
+  points.grainlineTop = points.upperLegY.shiftFractionTowards(
+    points.fork,
+    options.grainlinePosition
+  )
+  points.knee = new Point(points.grainlineTop.x, measurements.waistToKnee)
+  points.floor = new Point(
+    points.grainlineTop.x,
+    measurements.waistToFloor * (1 + options.lengthBonus)
+  )
+  points.grainlineBottom = points.floor
```
BenJamesBen, line 129: The grainline extends outside the pattern piece if lengthReduction is increased. You might want to set it later in the code after the lengthReduction adjustments and base it in relation to the reduced length.

BenJamesBen (dismissed):
I am not sure how complete designs should be before they are put back into the repository. However, think that the code could be okay to put back. The design does cause errors under certain circumstances, but it should be stable enough to use to test fit.

designs/lily/src/back.mjs
```diff
@@ -0,0 +1,567 @@
...
+    points.titleAnchor = points.logoAnchor.shift(-90, 60)
+    macro('title', {
+      nr: 1,
+      title: 'back',
+      at: points.titleAnchor,
+    })
+    
+    //notches
+    //if (options.fitGuides) {
+    if (true) {
```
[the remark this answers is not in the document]
anna-puk: I know they were removed from Titan, but I found them rather useful and kept them in intentionally.

It's not very pretty, but having the knee notches even for large length reduction values does make it easy to see where the hem is expected to hit.

designs/lily/src/back.mjs
anna-puk: Yes, the other widths are controlled via their respective ease values. Ideally, fabricStretch would control those values, but I haven't yet found a way to do so.

designs/lily/src/back.mjs
anna-puk: This seems to occur consistently when the leggings are cut off _at_ the knee (rather than above or below). It's a bit tricky to fix, so it'll have to wait until a newer version.

designs/lily/src/back.mjs
anna-puk: I've made a note, to be fixed in a later version.

designs/lily/src/back.mjs
```diff
@@ -552,15 +557,16 @@ export const back = {
...
+    crotchDrop: { pct: 0, min: 0, max: 15, menu: 'advanced' }, // 'downgrade' to advanced menu
     waistHeight: { ...titanBack.options.waistHeight, pct: 50 }, // halfway between waist and hips
     fabricStretch: { pct: 40, min: 0, max: 50, menu: 'fit' },
-    waistEase: {pct: -4,  menu: undefined}, // -fabricStretch/10,
-    seatEase: {pct: -4, menu: undefined}, // -fabricStretch/10,
-    kneeEase: {pct: -4, menu: undefined}, // -fabricStretch/10,
-    test: {pct: ({ options }) => (options.fabricStretch/2)},
+    waistEase: {pct: -4, min: 0, max: 50, menu: 'fit' }, // -fabricStretch/10,
+    seatEase: {pct: -4, min: 0, max: 50, menu: 'fit' }, // -fabricStretch/10,
+    kneeEase: {pct: -4, min: 0, max: 50, menu: 'fit' }, // -fabricStretch/10,
```
BenJamesBen, line 565: You have set the default value of `waistEase`, `seatEase`, and `kneeEase` to -4%. However, the minimum value is set to 0%. This seems as if something is incorrect.

BenJamesBen (approved):
I tested updated Liiy in the localhost lab, including the updated options. There are still issues with the design, but I believe it is in a good enough state for the PR to be merged, for further testing/development.

designs/lily/src/back.mjs
```diff
@@ -0,0 +1,573 @@
...
+    fabricStretch: { pct: 40, min: 0, max: 50, menu: 'fit' },
+    waistEase: {pct: -4, min: 0, max: 50, menu: 'fit' }, // -fabricStretch/10,
+    seatEase: {pct: -4, min: 0, max: 50, menu: 'fit' }, // -fabricStretch/10,
+    kneeEase: {pct: -4, min: 0, max: 50, menu: 'fit' }, // -fabricStretch/10,
+    //test: {pct: back.options.fabricStretch/2, min: 0, max: 50, menu: 'fit'},
+    lengthBonus: 0,
+    lengthReduction: { pct: 0, min: 0, max: 100, menu: 'style'},
+    waistbandWidth: { ...titanBack.options.waistbandWidth, menu: 'style' },
+    },
+  hideDependencies: true,
```
BenJamesBen, line 571: The parts config API has since changed for hiding parts and dependencies. Please see https://freesewing.dev/reference/api/part/config/hide .
I think the updated config would be: `hide: 'HIDE_TREE',`

You'll probably want to change it here and in other files.

designs/lily/src/back.mjs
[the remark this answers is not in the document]
BenJamesBen: Also, you'll probably want to remove the surrounding `if (true) { }` block, to avoid the lint error.

BenJamesBen (changes requested):
I retested the PR in the lab. I note that all the model size patterns crash due to the fitguides issue when the Waist Ease option is increased beyond a certain point.

The recent part-hiding core changes also result in the Titan parts appearing in the pattern.

There are some llnt errors that should be fixed before the PR is merged. Otherwise, the errors will appear in other people's PRs.

joostdecock:
If we can address Ben's outstanding remarks we could merge this.

designs/lily/src/back.mjs
anna-puk: Indeed, I expect negative values here... now adjusted.

designs/lily/src/back.mjs
anna-puk: As for the lengthReduction maximum value: it actually caps when the leggings end at the fork, with the following log message: "length reduction capped; cutting off at fork".

The percentage this ends up being depends on the measurements, so I just let it go all the way to 100%.

designs/lily/src/back.mjs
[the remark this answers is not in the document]
anna-puk: I made the fit guides optional and implemented the fix you suggested; I've seen no issues since.

designs/lily/src/back.mjs
```diff
@@ -0,0 +1,573 @@
...
+        paths.hipline = new Path()
+          .move(points.hipsIn)
+          .line(points.hipsOut)
+          .attr('class', 'fabric help')
+          .attr('data-text', 'Hip Line')
+          .attr('data-text-class', 'center')
+      }
+    }
+   
+    if (sa) {
```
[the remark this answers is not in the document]
anna-puk: This is now fixed.

designs/lily/src/front.mjs
```diff
@@ -0,0 +1,676 @@
...
+        // points.seatIn = utils.lineIntersectsCurve(
+          // points.seatMid,
+          // points.seatInTarget,
+          // points.fork,
+          // points.forkCp2,
+          // points.points.KneeInCp1,
+          // points.kneeIn
+        // )   
+          // TODO: fix this
+          points.seatIn = points.fork.clone()
```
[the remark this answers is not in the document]
anna-puk: fit guides are disabled by default now, so I hope it's okay to fix this later

anna-puk:
I've completed all the changes which are feasible for now - the design appears to work fine for the majority of cases now.

BenJamesBen (approved):
I retested the updated PR in the lab, and I believe the PR is fine to merge for wider testing. There are still some minor issues, but they can be addressed later.

I noticed these issues:
- Grainline/scalebox/title fall outside the front and back parts when `lengthReduction` is increased.
- There is still a crash when `lengthReduction` is specific values (for cisFemale Size 28, it occurs from 45.4% to 45.6%).
- You might want to consider changing the `scale` and position of the Waistband title, to make it fit better and not get cut off.

I also noticed:
- With cisFemale Size 28, something odd happens to the shape of the Front seat when I use keyboard arrow keys to change `seatEase` from -3.1% to -3.0% to -2.9% to -2.8%. The shape jumps from straight to curved to straight and then back to curved. With the Back part, the shape jumps from curved to straight when `seatEase` is changed from -7.5% to -7.6%.
- I see similar shape jumps in Front when adjusting `crotchSeamCurveAngle` around 4.3-5.7%, and when adjusting `crotchSeamCurveBend` around 97.0-98.5%, and also with `crotchSeamCurveStart`.
- The Front seat `fitGuide` slope jumps between 82.6% and 82.7%, but I am not sure if this might be correct or not.
- I note that the outline of Front and Back seem to get bolder/thicker when `lengthReduction` is set to anything other than 0%. (I'm not sure what might be going on, but this is easier to see if you zoom out in DraftDesign view.)

BenJamesBen:
Another minor issue, the bottom leg opening dimensions do not adjust values or positions when `lengthReduction` is changed.
Screenshot 2023-04-18 at 4 48 35 PM

BenJamesBen:
I started to review the latest version of this PR, but the Lily design doesn't appear in the lab. I see that the source files still exist, but I don't see Lily in `config/software/designs.json`. 

Were some files accidentally reverted during one of the upstream merges?

anna-puk:
> I started to review the latest version of this PR, but the Lily design doesn't appear in the lab. I see that the source files still exist, but I don't see Lily in `config/software/designs.json`.
> 
> Were some files accidentally reverted during one of the upstream merges?

It was actually a consequence of this PR originating from v2 and never having been ported to v3 properly. I added `lily` to config/software/design.json manually and now it appears to work.

There are still some issues, including the 'flag' approach not working for the settings I'm trying to change.

designs/lily/src/back.mjs
anna-puk: (I've reverted this back to a 'draft' pull request since the `flag` thing doesn't work yet.)

anna-puk:
(I've reverted this back to a 'draft' pull request since the flag thing doesn't work yet.)

BenJamesBen:
I tested the updated PR in the lab, though I realize that it is in Draft status and is not yet complete.

Things I need to test or look at the next time the PR is updated:
- Increasing lengthReduction to a certain point (bottom of leg moves above the fork, I believe) caused an error to occur. I think it failed when trying to set `points.bottomOut`.
- The length dimension does not adjust based on lengthReduction.
- Some of the dimensions inherited from Titan don't seem to be applicable to Lily. I think they include dimensions related to Titan's fly area.
- The extra files included in the PR, like the test5 design and `box.mjs`.

BenJamesBen:
There is still a crash that occurs in `front.mjs` when Length Reduction is set near the knee points. 

To reproduce:
1. Generate Lily using the left-most curated set
2. Move the Length Reduction to between 41.2% and 41.6%.

What is happening is that `utils.lineIntersectsCurve()` returns `false`, so setting `points.bottomOut` or `points.bottomIn` fails. I am not sure if it is because `utils.lineIntersectsCurveI()` is failing due to Bug #3367 or if it is because there might be a logic error in the line and curve you are using. (It might be that the line and curve actually don't intersect, due to a logic error in where you've set the horizontal line. But, I'm not sure. More debug would need to be done to figure out exactly what is going on.)

I came up with this workaround that pegs the points at `points.kneeOut` and `points.kneeIn` whenever `utils.lineIntersectsCurve()` fails. It's not the best solution, but it prevents the crashes enough to allow my testing to continue.

```diff
--- a/designs/lily/src/front.mjs
+++ b/designs/lily/src/front.mjs
@@ -294,38 +294,46 @@ function draftLilyFront({
         upperPoint = points.seatOut
         upperCp = points.seatOutCp2
       }
-       points.bottomOut = utils.lineIntersectsCurve(
+       let testpoint = utils.lineIntersectsCurve(
           points.bottom.shift(0,999),
           points.bottom.shift(180,999),
           points.kneeOut,
           points.kneeOutCp1,
           upperCp,
           upperPoint)
+      if (testpoint) points.bottomOut = testpoint
+      else points.bottomOut = points.kneeOut
           
-        points.bottomIn = utils.lineIntersectsCurve(
+        testpoint = utils.lineIntersectsCurve(
           points.bottom.shift(0,999),
           points.bottom.shift(180,999),
           points.kneeIn,
           points.kneeInCp2,
           points.forkCp1,
           points.fork)     
+      if (testpoint) points.bottomIn = testpoint
+      else points.bottomIn = points.kneeIn
     } else {
       // 'cut' between knee and 'floor'      
-      points.bottomOut = utils.lineIntersectsCurve(
+      let testpoint = utils.lineIntersectsCurve(
         points.bottom.shift(0,999),
         points.bottom.shift(180,999),
         points.kneeOut,
         points.kneeOutCp2,
         points.floorOutCp2,
         points.floorOut)
+      if (testpoint) points.bottomOut = testpoint
+      else points.bottomOut = points.kneeOut
         
-      points.bottomIn = utils.lineIntersectsCurve(
+      testpoint = utils.lineIntersectsCurve(
         points.bottom.shift(0,999),
         points.bottom.shift(180,999),
         points.kneeIn,
         points.kneeInCp1,
         points.floorInCp2,
         points.floorIn)     
+      if (testpoint) points.bottomIn = testpoint
+      else points.bottomIn = points.kneeIn
     }
     
     // define the three parts of the path, then combine 
```

anna-puk:
> There is still a crash that occurs in `front.mjs` when Length Reduction is set near the knee points.
> 
> To reproduce:
> 
>     1. Generate Lily using the left-most curated set
> 
>     2. Move the Length Reduction to between 41.2% and 41.6%.
> 
> 
> What is happening is that `utils.lineIntersectsCurve()` returns `false`, so setting `points.bottomOut` or `points.bottomIn` fails. I am not sure if it is because `utils.lineIntersectsCurveI()` is failing due to Bug #3367 or if it is because there might be a logic error in the line and curve you are using. (It might be that the line and curve actually don't intersect, due to a logic error in where you've set the horizontal line. But, I'm not sure. More debug would need to be done to figure out exactly what is going on.)

Thanks for testing this again, and for the suggestion of a logic error!

I found the problem and have now fixed it.

(For the curious: the issue was caused by an incorrect assumption that the Y location of points.knee was equal to measurements.waistToKnee. This is true for `back` but not for `front`. I had checked only the former... I finally noticed that the error was originating from `front` and found that `back` could be drawn without errors for any length reduction.)

anna-puk:
I'm seeing a message about 'conflicting files' which will prevent the pull request from being merged. How do I get rid of these? I never intended to change them.

BenJamesBen:
To be honest, I think that the best thing to do might be to make sure your `develop` branch is up-to-date, create a new branch off of it, and add your new Lily files/changes to the new branch. Basically, abandon your old branch and use a new one. That would eliminate the conflicts.

(I think that this might mean that you would also need to abandon this PR and file a new one, but I'm not sure about this. Or, perhaps by renaming the old branch and re-using the `leggings-(based-on-titan)` name for the new branch, this might allow you to continue using this PR?)

BenJamesBen:
Screenshot 2024-02-11 at 7 54 31 PM

Steps to Reproduce
1. In the localhost lab, create a new Lily using the left-most curated set.
2. Design Options -> Style -> Length Reduction, adjust slider to 68.5%.
3. The crash occurs.

In `src/back.mjs`:
```js
    let halves = paths.seam.split(points.bottomIn)
     paths.upperInseam = halves[0]
```
The `paths.seam.split()` is not returning a Path in `halves[0]`. I suspect that `points.bottomIn` is either not on the path or it is at the very end of the path.

PR #3056, opened 2022-11-12
fix(core): Drop noop Bezier curves from offset curves. Add debug messages. Fixes #3038.

### LINKED ISSUE #3038: [bug]: Offset curves can include spurious lines

Offset curves can sometimes include spurious lines. For example, the two diagonal lines seen in the seam allowance and the path distance dimensions in this screenshot:
Screenshot_20221102-171342

The issue is an edge case occuring with some curves but not others, depending on the geometry of the curve.

Steps to reproduce:
1. In lab.freesewing.dev, generate a Hugo pattern Without Breasts Size 50.
2. In the DraftDesign view, enable Paperless and Include Seam aallowance.
3. Observe that the diagonal lines are seen in Part 6 HoodSide.
4. Generate the same Hugo pattern Without Breasts Size 32.
5. Observe that the diagonal lines are no longer seen. (The Size 32 curve happens to not result in the issue.)

packages/core

No, I am not 😞

### Additional context

The root cause is a combination of a behavior in the `bezier-js` package and how our `Path.offset()` code interacts with the behavior.

The `Bezier.offset()` method returns an array of Bezier curves that approximate a curve that is offset from an input Bezier curve. However, an unexpected behavior is that this array can sometimes contain spurious Bezier curves which are noops. That is, all 4 points are identical when the x and y values are rounded to the nearest integer.
https://pomax.github.io/bezierjs/#offset

For example, these 8 Bezier curves were returned for one offset() operation. Note the 3 noop Bezier curves at #<span/>129, #<span/>135, and #<span/>136.
```
129 | 9,312 : 9,312 : 9,312 : 9,312
130 | 9,312 : 8,309 : -3,285 : -14,250
131 | -14,250 : -23,222 : -32,188 : -36,154
132 | -36,154 : -37,142 : -38,131 : -38,120
133 | -38,120 : -38,67 : -21,20 : 23,-6
134 | 23,-6 : 43,-18 : 70,-25 : 104,-25
135 | 79,0 : 79,0 : 79,0 : 79,0
136 | 79,0 : 79,0 : 79,0 : 79,0
```

Then, in our `path.mjs` the `__asPath()` method converts the noop Bezier curves to spurious Paths with spurious `move()` and `curve()` operations.

And finally, when we `__joinPaths()` these paths together, a spurious `move()` operation gets turned into a spurious `line()`:
```js
} else if (op.type !== 'close') {
  // We're using sitsRoughlyOn here to avoid miniscule line segments
  if (current && !op.to.sitsRoughlyOn(current)) joint.line(op.to)
 ```

In the above example, a spurious line gets drawn from 104,-25 to 79,0 when paths/curves #<span/>134 and #<span/>135 are joined.

A possible fix, in `packages/core/src/path.mjs`:
```diff
@@ -1279,7 +1279,22 @@ function __pathOffset(path, distance, log) {
         { x: cp2.x, y: cp2.y },
         { x: op.to.x, y: op.to.y }
       )
-      for (let bezier of b.offset(distance)) offset.push(__asPath(bezier, path.log))
+      for (let bezier of b.offset(distance)) {
+        // Bezier.offset() sometimes includes noop Bezier curves (consisting
+        // of 4 effectively-identical points) in its returned array. This
+        // can cause problems if we later try to join these noop curves to
+        // a Path. The workaround is to ignore these noop curves.
+        if (
+          (Math.round(bezier.points[0].x) === Math.round(bezier.points[1].x)) &&
+          (Math.round(bezier.points[0].x) === Math.round(bezier.points[2].x)) &&
+          (Math.round(bezier.points[0].x) === Math.round(bezier.points[3].x)) &&
+          (Math.round(bezier.points[0].y) === Math.round(bezier.points[1].y)) &&
+          (Math.round(bezier.points[0].y) === Math.round(bezier.points[2].y)) &&
+          (Math.round(bezier.points[0].y) === Math.round(bezier.points[3].y))) {
+          continue
+        }
+        offset.push(__asPath(bezier, path.log))
+      }
     } else if (op.type === 'close') closed = true
     if (op.to) current = op.to
     if (!start) start = current
```

BenJamesBen:
This PR fixes https://github.com/freesewing/freesewing/issues/3038 by dropping noop Bezier curves returned by the `Bezier.offset()` method. 

Basically, if all 4 effective (rounded to the nearest millimeter) points of a Bezier curve are identical, it is considered to be a noop. (It's not an actual curve. It is just a single point.)

This PR also adds debug messages that can be toggled on for future debugging purposes. The debug messages are disabled by default. When enabled, the debug log messages look like:
```
162 | Array of Beziers returned for an offset curve:
163 | (23,310, 23,310, 23,310, 23,310) (dropped)
164 | (23,310, 22,307, 11,284, 1,250)
165 | (1,250, -8,222, -16,188, -20,154)
166 | (-20,154, -21,144, -21,134, -21,124)
167 | (-21,124, -21,76, -7,31, 33,8)
168 | (33,8, 52,-3, 77,-10, 109,-10)
169 | (99,0, 99,0, 99,0, 99,0) (dropped)
170 | (99,0, 99,0, 99,0, 99,0) (dropped)
171 | (99,0, 99,0, 99,0, 99,0) (dropped)
```

### DIFF
--- a/packages/core/src/path.mjs
+++ b/packages/core/src/path.mjs
@@ -1256,6 +1256,7 @@ function __pathOffset(path, distance, log) {
   let current
   let start = false
   let closed = false
+  const BEZIER_OFFSET_DEBUG = false
   for (let i in path.ops) {
     let op = path.ops[i]
     if (op.type === 'line') {
@@ -1279,7 +1280,46 @@ function __pathOffset(path, distance, log) {
         { x: cp2.x, y: cp2.y },
         { x: op.to.x, y: op.to.y }
       )
-      for (let bezier of b.offset(distance)) offset.push(__asPath(bezier, path.log))
+      if (BEZIER_OFFSET_DEBUG) log.debug('Array of Beziers returned for an offset curve:')
+      for (let bezier of b.offset(distance)) {
+        // The array of Bezier curves returned by Bezier.offset() sometimes
+        // includes noop curves (consisting of 4 effectively-identical
+        // points) in its returned array. This can cause spurious lines
+        // if we later try to join them to a Path. The workaround is to
+        // ignore these noop curves.
+        let debug_message =
+          '(' +
+          Math.round(bezier.points[0].x) +
+          ',' +
+          Math.round(bezier.points[0].y) +
+          ', ' +
+          Math.round(bezier.points[1].x) +
+          ',' +
+          Math.round(bezier.points[1].y) +
+          ', ' +
+          Math.round(bezier.points[2].x) +
+          ',' +
+          Math.round(bezier.points[2].y) +
+          ', ' +
+          Math.round(bezier.points[3].x) +
+          ',' +
+          Math.round(bezier.points[3].y) +
+          ')'
+        if (
+          Math.round(bezier.points[0].x) === Math.round(bezier.points[1].x) &&
+          Math.round(bezier.points[0].x) === Math.round(bezier.points[2].x) &&
+          Math.round(bezier.points[0].x) === Math.round(bezier.points[3].x) &&
+          Math.round(bezier.points[0].y) === Math.round(bezier.points[1].y) &&
+          Math.round(bezier.points[0].y) === Math.round(bezier.points[2].y) &&
+          Math.round(bezier.points[0].y) === Math.round(bezier.points[3].y)
+        ) {
+          debug_message += ' (dropped)'
+          if (BEZIER_OFFSET_DEBUG) log.debug(debug_message)
+          continue
+        }
+        if (BEZIER_OFFSET_DEBUG) log.debug(debug_message)
+        offset.push(__asPath(bezier, path.log))
+      }
     } else if (op.type === 'close') closed = true
     if (op.to) current = op.to
     if (!start) start = current


joostdecock (commented):
Would we be better of with a method on the path object that filters out these `noop` path operations?
As in, this guards against this scenario in the specific case of using `Path.offset()` but people could (in principle at least) end up with these type of `noop` drawing operations in other ways.

So we could extract this logic in some sort of `Path.clean()` method. Then keep this method as it is, but just return `this.clean()` at the end.

Thoughts?

BenJamesBen:
I am fine with whatever way you think the fix should be implemented. I am not sure how the `Path.clean()` method would be done. 

If the op `type === 'curve'` and `cp1 == cp2 == cp3`, then drop the op? The ops array doesn't seem to preserve the value of starting point `cp0`; instead it seems to assume that starting point was already set in a previous op. Bug #3038 occurs when the starting point of the curve is actually _not_ the ending point of the previous op.

joostdecock (approved):

joostdecock:
As discussed above, a more generic fix was implemented in https://github.com/freesewing/freesewing/commit/fab1e2f8777e75ef950016b594230435752ccf30

Still mad props to @BenJamesBen for getting to the bottom of this :pray:

BenJamesBen:
The previous commit did not fix the problem, so I am re-opening this PR. More details can be found in Bug https://github.com/freesewing/freesewing/issues/3038.

joostdecock:
Thanks for your efforts on this @BenJamesBen
I've fixed this in https://github.com/freesewing/freesewing/commit/5ba6e124690cea6b9ab9b91934268ce558a669f2 rather than implement your suggestion in #3038 

My reason for doing is is two-fold:

- Your fix handled the check for problematic segments in the offset code. However, we previously said that it would be better to have a generic method for this (`Path.clean()`) so we impemented that. `Path.clean()` actually did what it was supposed to do and would strip out the problematic drawing operations. Problem was that it would then return a path with only a move operation, which caused the same problem. So I've fixed that in the `Path.clean()` method by adding an extra check
- You where using Math comparisons to compare the various points to each other, but we need to convert them to Point instances anyway prior to returning them in `Path.__asPath()` so I just used the `Point.sitsRoughlyOn()` API method instead.

The end result in the lab:

image

PR #3058, opened 2022-11-12
fix(core): Calculate option distances correctly. Fixes #2995.

### LINKED ISSUE #2995: [bug]: unable to disable options inherited from other parts

When I inherit from `titan` (using `from: titanFront` and `from:titanBack`, respectively), options which I try to override are still shown in the menu. Also, the logs show that these options are being taken from `titan.back`.

The values with which I want to override are in `lily.back`. Draft order is: `titan.back, lily.back, titan.front, lily.front`.

_No response_

No, I am not 😞

### Additional context

code which triggers the bug is available here: https://github.com/anna-puk/freesewing/tree/leggings-(based-on-titan)/designs/lily/src

BenJamesBen:
This PR:
1. Fixes the logic for when options are overwritten. It does so by flipping a "<" to ">" to make lower values take precedence.
2. Changes how option distance is calculated for from:/after: parts. For these dependent parts, the distance is set to the current part distance + 1. The code ensures that the distance for a part can only increase.
3. Adds extensive debug messaging controlled by a `DISTANCE_DEBUG` flag (disabled by default).
4. Adds additional debug log messages (not controlled by `DISTANCE_DEBUG`) listing the final distances for each part. The messages use the wording "options priority" to try to make it easier to understand for those not familiar with the "distance" concept:
```
0 | ⚪️  paco.back options priority is 3
1 | ⚪️  paco.front options priority is 2
2 | ⚪️  paco.cuff options priority is 1
3 | ⚪️  paco.frontPocketBag options priority is 1
4 | ⚪️  paco.backPocketBag options priority is 3
5 | ⚪️  paco.backPocketWelt options priority is 2
6 | ⚪️  paco.backPocketWeltInterfacing options priority is 1
7 | ⚪️  paco.waistband options priority is 1
8 | ⚪️  titan.back options priority is 4
9 | ⚪️  titan.front options priority is 3
```

This PR fixes https://github.com/freesewing/freesewing/issues/2995 .

### DIFF
--- a/packages/core/src/pattern.mjs
+++ b/packages/core/src/pattern.mjs
@@ -12,6 +12,8 @@ import { Hooks } from './hooks.mjs'
 import { version } from '../data.mjs'
 import { __loadPatternDefaults } from './config.mjs'
 
+const DISTANCE_DEBUG = false
+
 //////////////////////////////////////////////
 //               CONSTRUCTOR                //
 //////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


joostdecock (commented):
Is the intent here to merge & keep all this debug code?
Because I feel that adds a lot of noise to the signal :thinking:

joostdecock (approved):

joostdecock:
@all-contributors please add @BenJamesBen for bugs

Issue #2997, opened 2022-10-23
[bug]: PDF export currently broken in the lab


anna-puk:
### What seems to be the problem? 🤔

When using the lab, I cannot export any patterns. (I've tried several.) I get the same error in my own branch and develop (both on Windows) and in Gitpod as well. On Discord, other users report the same issue.

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

To reproduce, start the lab, got to e.g. `titan`, go to `ExportThing` and click the 'SVG' button.

error message:

```
Unhandled Runtime Error

TypeError: plugin.plugin is undefined
Source

../../packages/core/src/pattern.mjs (494:8) @ getPluginName

  492 | } else {
  493 |   if (plugin.name) return plugin.name
494 |   if (plugin.plugin.name) return plugin.plugin.name
      |      ^
  495 | }
  496 | 
  497 | return false
```

joostdecock:
Note that SVG export is fixed.
And while PDF export is still broken, you can export to SVG and then use our on-demand tiler: https://tiler.freesewing.org/

I'm going to assign this to @eriese as they wrote the PDF exporter in the frontend. No pressure :)

eriese:
fixed by #3068

Issue #3008, opened 2022-10-24
[bug]: Exporting Paperless pattern to PDF or SVG results in error or undefined file


BenJamesBen:
### What seems to be the problem? 🤔

In the v3 `develop` localhost lab, when a pattern with Paperless enabled is exported as SVG, the resulting file is a 9-byte file consisting of the text characters "undefined" (with no EOL at the end of the line).

### Are you a FreeSewing patron? 😃

No, I am not 😞

BenJamesBen:
Exporting a Paperless pattern to PDF (using the updated exporting functionality in #3068) results in an error message printed to the web page:
```
WARNING | ERROR: somethingWentWrong
```

eriese:
For PDF, do we want it to automatically turn paperless off, or do we want it to export with all the paperless markup? @joostdecock thoughts?

Issue #2803, opened 2022-09-11
[bug]: [v3] [lab] LayoutThing: ForPrinting is blank


BenJamesBen:
### What seems to be the problem? 🤔

In the localhost lab, the LayoutThing: ForPrinting view is blank:

![Screen Shot 2022-09-11 at 3 25 10 PM](https://user-images.githubusercontent.com/109869956/189551432-e5123db7-cf78-4f4e-9240-4f47abd82f65.png)

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

#2802 ("[bug]: [v3] [lab] Runtime error when exporting to PDF") is another, possibly-related issue introduced around the same time.

eriese:
fixed by #3608

PR #3068, opened 2022-11-14
[Feature] V3 printing layout and pdf export

eriese:
Printing functionality reinstated for v3. doesn't solve the open question in #3008, but I think it's worth merging as a start


packages/core/src/pattern.mjs
```diff
@@ -55,7 +56,7 @@ export function Pattern(designConfig) {
   __addNonEnumProp(this, '__hide', {})
 
   // Enumerable properties
-  this.designConfig = designConfig // The design configuration (unresolved)
+  this.designConfig = cloneDeep(designConfig) // The design configuration (unresolved)
```
eriese, line 59: necessary because the design config was being shallow copied from the gist, so runtime changes that were supposed to be ephemeral were also mutating the gist, and then if a gist save was triggered those changes were polluting the gist

BenJamesBen:
I tested the changes in the localhost lab, and the print layout and exporting to PDF functions seem to work correctly.

eriese:
fixes #2997  #2803

packages/core/src/pattern.mjs
```diff
@@ -1132,6 +1143,7 @@ Pattern.prototype.__optionSets = function (optionName) {
  * @return {Pattern} this - The Pattern instance
  */
 Pattern.prototype.__pack = function () {
+  this.__runHooks('preLayout')
```
eriese, line 1146: I wanted hooks that got called internally by `__pack` because as it was `preRender` was being called *before* `__pack` in `getRenderProps` and *after* `__pack` in `render`, and I needed a consistent order of operations

sites/shared/components/workbench/draft/circle/index.js
```diff
@@ -1,20 +1,16 @@
...
-  return (
-    <circle
-      cx={props.point.x}
-      cy={props.point.y}
-      r={r}
-      {...extraProps}
-    />
-  )
-})
+    return <circle key={r} cx={props.point.x} cy={props.point.y} r={r} {...extraProps} />
```
eriese, line 13: The real change in this file is that I added a key because this returns an array

eriese:
@joostdecock this is ready to go

joostdecock (approved):

joostdecock:
Pretty great stuff @eriese 👍

PR #3076, opened 2022-11-15
[Fix] Get tests passing

eriese:
- Updated github actions versions in workflows
- changed lerna configuration in testall script for faster test running and better output
- refactored path tests that were relying on `pattern.draft()` to unit test path functions


joostdecock:
Merged this on the CLI. Tx for the work @eriese 
There's a test in Jaeger that's not passing (sampling the `centerFrontHemDrop` option), but I don't have time to look into that now. 

Still, this is obvs good stuff, so merging it coz YOLO

eriese:
Weird... It was passing before. I'll look into that

PR #3080, opened 2022-11-16
[Fix] Codecov report uploads

eriese:
- the correct file is uploaded to codecov after tests pass
- codecov has a known rate limiter issue where it sometimes rejects the upload, so codecov upload failure doesn't fail the check
- lint checks prebuild instead of building everything. I'm not sure this is faster any, but it makes me feel better
- terser reporter now includes test-failure-collector's functionality, which moves that out of the `after` hook and into a listener that only runs on failures, which helps speed things up in the tests


eriese:
@joostdecock , I'm done playing around with this and it can be merged

joostdecock (approved):

joostdecock:
I'm trying 🤷

eriese:
@joostdecock you might have to come to actual github to merge it?

joostdecock:
Yeah well I don't seem to come around here too often 😂

PR #3118, opened 2022-11-29
Refactor and fix new-design

eriese:
This started as a little fix to add a missing file that was blocking compilation, but then I got into the code a bit and decided to consolidate the templates to make maintenance easier.

The way it works now is that we have a single template for `package.json`, a single template for `design/src/index.mjs` and a single template for part files. Each possible base has a configuration file which tells the compiler what parts to make files for and how. At the simplest end where it's inheriting from an existing design, that config file just lists part names and dependencies. At the most complex end is the `from-scratch`, which has individual configuration for parts


eriese:
All remaining diff lint errors seem to be due to the linter interpreting mustache files as js files. the `lint` command in `new-design` skips the templates folder, so these errors won't crop up again

eriese:
Converting to draft to work on combining with #2971

eriese:
> > Converting to draft to work on combining with #2971
> 
> So no review for now?

@joostdecock  It's ready now.

@nchilada, I merged your PR branch in and kept what still made sense after I consolidated the templates. 
The one thing I didn't keep that could have stayed was templating `shared/pages/design.mjs` to import the design by name. I did keep the named export of the design in `design/src/index.mjs` because I agree that we should encourage the habits we want to see in the monorepo, but I felt it was better to keep the design name only in files that we actually want standalone users to interact with, rather than in the boilerplate. I have been known to choose a random name for my pattern and then later want to rename it, so with `import {design} from 'design/src/index.mjs` we avoid people with that use case having to leave the design folder to muck around in code that's not covered in design documentation.

packages/new-design/lib/cli.mjs
```diff
@@ -1,15 +1,9 @@
...
-} from './utils.mjs'
-
+// import { banner } from './banner.mjs'
+import { checkNodeVersion, getChoices, createEnvironment } from './utils.mjs'
 
 export const cli = async () => {
-
   // Make it pretty
-  console.log(banner+"\n")
+  // console.log(banner+"\n")
```
joostdecock, line 6: You don't like the banner?

joostdecock (approved):
yolo

packages/new-design/lib/utils.mjs
```diff
@@ -114,41 +117,123 @@ export const getChoices = async () => {
...
+  await ensureDir(to)
+
+  if (templateVars) {
+    const rendered = mustache.render(fromRootOrTemplate, templateVars)
+    await writeFile(to, rendered)
+  } else {
+    const from = join(fromRootOrTemplate, relativeDest)
+    await copyFile(from, to)
+  }
+}
```
nchilada, line 142: IMO this function is a little confusing now. `fromRootOrTemplate` may be either:

* A directory name (i.e. a `fromRoot`), in which case `relativeDest` doubles as a `relativeSource`
* The serialized _contents_ of a template file

Might be better for this argument to act as the same kind of thing in both cases, or two split this function into two function, `copyFile` and e.g. `renderTemplate`

packages/new-design/templates/shared/index.mjs.mustache
```diff
@@ -1,13 +1,17 @@
+{{!
```
nchilada, line 1:
```suggestion
//{{!
```
as in #2971, and ditto for all the files below

nchilada (commented):
Thanks for resuscitating my branch! And great timing, as I just returned home a couple days ago. Let me know what you'd like me to do with my PR. I imagine it's easiest for me to close my PR, since you've already rebased and probably won't want to rebase again, but let me know if you prefer for me to merge separately now that I'm back.

In any case, I'm looking forward to seeing what you all have done while I've been out! As soon as I finish getting my life back in order 😄

> @nchilada, I merged your PR branch in and kept what still made sense after I consolidated the templates. The one thing I didn't keep that could have stayed was templating `shared/pages/design.mjs` to import the design by name. I did keep the named export of the design in `design/src/index.mjs` because I agree that we should encourage the habits we want to see in the monorepo, but I felt it was better to keep the design name only in files that we actually want standalone users to interact with, rather than in the boilerplate. I have been known to choose a random name for my pattern and then later want to rename it, so with `import {design} from 'design/src/index.mjs` we avoid people with that use case having to leave the design folder to muck around in code that's not covered in design documentation.

Makes sense to me! I agree, while it may be okay to expect developers to update the boilerplate in the `design` subdirectory, it's probably unreasonable to ask them to care about the `pages` subdirectory. Kudos for identifying such paper cuts ahead of time.

Edit: hwhoops, too slow! Oh well, consider the inline comments as suggestions for the future. I imagine we'll need to fix the Mustache templates, at the very least

packages/new-design/lib/utils.mjs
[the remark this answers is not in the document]
eriese: agreed. I liked that you had combined writing and copying but it made more sense when every file that needed to be written had its own file that needed to be read from. Now that parts are all written from the same template, probably templating needs to move back out to its own function

packages/new-design/templates/shared/index.mjs.mustache
nchilada: > I found that the old one just made the first line of the file be `//`.

Ah, yes, that's right. Per the code comment itself, all of this came out of my efforts to [get our Mustache templates to pass lint](https://github.com/freesewing/freesewing/pull/2970/commits/93788aa0b41f63810b48793b7db0701269827799). In files like this, the best solution I could find was to [change the delimiter](https://github.com/janl/mustache.js#custom-delimiters) with a comment in front of the delimiter change. This does unfortunately produce an empty comment in the Mustache output.

(I think I'd previously tried to fix all our Mustache linting problems by writing an [eslint preprocessor](https://eslint.org/docs/latest/developer-guide/working-with-plugins#processors-in-plugins) that could render Mustache templates, but IIRC it's nontrivial to find dummy values for the various template variables. In any case I guess it's probably a good thing for our JS files to be syntactically correct, for the sake of prettier and editors and other static analysis tools.)

> I imagine we'll need to fix the Mustache templates, at the very least

Oops, disregard this, it's lower-priority than I'd feared. Because you're right, syntax issues are cleaned up by Mustache before the file ever reaches a JS interpreter. I can simply follow up on this in my next linting PR.

eriese: Probably the thing for us to do is  add the mustache files to the lint ignore, then make a `prelint` script that generates the files for from scratch, tutorial, and from bent. Then those will be linted and a `postlint` can clean then up

PR #3124, opened 2022-12-03
[Fix] Everything passes lint

eriese:
What it says on the tin. I fixed all remaining errors, and any warning that I saw that weren't about using `next/Image`


joostdecock:
Nice. I probably broke a bunch of stuff again by upgrading to Next 13/React 18 but it had to happen at some point :shrug:

PR #3139, opened 2022-12-05
fix(breanna): Eliminate warning messages when there is no secondary dart

BenJamesBen:
This PR fixes the issue of when there is no secondary bust dart (set to none or set to same as primary bust dart), 6 warning messages are logged.
Screenshot 2022-12-05 at 6 48 25 AM


joostdecock (approved):

joostdecock:
Nice work 👍

PR #3169, opened 2022-12-08
fix(docs): Update and correct Part Options docs

BenJamesBen:


joostdecock (commented):
Based on your edits, we might have different opinions on this, but I feel it's important to clarify when something is not -- strictly speaking -- relevant for the core API docs.

Both in **Specifying the menu** and **Extending and overriding inherited options** we are overloading options with things that are specific to *how we manage the frontend integration*.  
But people looking for reference documentation on our core API may very well use a different frontend, and so this section is not relevant for them.

So since in the menu section you added, this is not mentioned. And in the translation section (where it was mentioned) you removed it. I guess you feel different :)

Perhaps we should discuss so we can make sure that everything is documented without confusing people about what's core fuctionality and what is merely consensus.

joostdecock:
Added this to the call agenda: https://github.com/freesewing/freesewing/discussions/3108

BenJamesBen:
Thanks, I'll be on the call.

There were 2 duplicate "Supress Translations" sections on that documentation page. I left 1 intact and changed the other to cover extending inherited options. Because the page was named "Extending Options", that was what I assumed the page was intended for... with the Supressing Translations info was added to it as an afterthought. Or that the Supressing Translations page was intended to be converted to cover extending options.

It might be that documentation specific to the FreeSewing frontend UI belongs somewhere else. However, it isn't clear to me (with current documentation structure) where that would be. From my perspective, the first place a pattern designer/coder wanting to learn how to use `menu:` in their code would look would be in this API documentation.

BenJamesBen:
Closing this PR. Per today's Contributors Call, documentation structure will be reworked for docs that might cross between Core API and frontend UI/implementation.

For those edits not related to this issue, I will update the files and submit a new PR for them.

PR #3181, opened 2022-12-08
chore(docs): Add Utils.curveedge() documentation. Closes #3143.

### LINKED ISSUE #3143: [docs]: utils.curveEdge() is not documented

_No response_

No, I am not 😞

### Suggested content

The utils.curveEdge() method is not in the API reference.

Missing from:
- https://freesewing.dev/reference/api
- https://freesewing.dev/reference/api/utils

### Additional context

_No response_

BenJamesBen:
This PR might need modification, but I will need guidance about how to proceed.

I ran into issues when creating the Example for `curveEdge()`. I was not able to import BezierJS to use in the live Example, so I was not able to create the required `Bezier` object. 

As a workaround, I created a new `Utils.bezier()` helper function that creates and returns a `Bezier`. If this is not the best way to resolve the issue, the code change will need to be removed from this PR.

In the documentation Example, I commented out the original code I was trying to use. If we do keep `Utils.bezier()`, then we might want to remove this commented code from this PR. (But, we could also just keep it in to demonstrate how to create a `Bezier` directly.)

If we keep `Utils.bezier()`, then further work will need to be done to create documentation for it.

It could be that we might want to simply omit the live Example for `curveEdge()`. Instead, we could just list the code that should be used.

### DIFF
--- a/markdown/dev/reference/api/en.md
+++ b/markdown/dev/reference/api/en.md
@@ -65,6 +65,7 @@ The following named exports are **utility methods**:
 | `beamsIntersect`          | See the [beamsIntersect](/reference/api/utils/beamsintersect) documentation |
 | `capitalize`              | See the [capitalize](/reference/api/utils/capitalize) documentation |
 | `circlesIntersect`        | See the [circlesIntersect](/reference/api/utils/circlesintersect) documentation |
+| `curveEdge`               | See the [curveEdge](/reference/api/utils/curveedge) documentation |
 | `curveIntersectsX`        | See the [curveIntersectsX](/reference/api/utils/curveintersectsx) documentation |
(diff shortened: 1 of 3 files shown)


joostdecock:
I've put up a new PR ( #3213 ) that has your documentation commits cherry-picked, but changes the way we make `Bezier` available.

Rather than create a dedicated method for creating a `Bezier` instance, I've just added the Bezier library to the things that are passed to the part's draft method. This gives the user/developer maximum flexibility to use it as they see fit.
It also means that the entire Bezier library can be used in examples.

I've asked @BenJamesBen 's review of #3213. If it's ok for them, I think we can close this and merge #3213 instead

BenJamesBen:
Closing in favor of #3213 .

PR #3188, opened 2022-12-09
fix(core): Update Pattern.sampleOptions() and documentation

BenJamesBen:
This is a more complicated PR, with a number of changes to `packages/core/src/pattern.mjs` to allow `Pattern.sampleOptions()` to work with all option types:
- Return an empty array of sets if the `optionName` does not exist (instead of throwing an error).
- Code refactoring to allow a different numbers of runs/steps (needed for `count` and `mm` options where there might be fewer than 10 values to step through).
- For `count` and `mm` options, adjust the number of runs/steps if needed and `round()` the values to integers.
- Extend `__listOptionSets()` to also work with Boolean options. Rename the method to `__listBoolOptionSets()`.

The PR also includes an update to the `Pattern.sampleOptions()` documentation for the new functionality.

I tested the code changes with every type of option, except `mm`.

### DIFF
--- a/markdown/dev/reference/api/pattern/sampleoption/en.md
+++ b/markdown/dev/reference/api/pattern/sampleoption/en.md
@@ -7,13 +7,13 @@ to draft multiple variants of the same pattern, and stack them on
 top of each other.
 
 In this particular case, the variants it drafts depend
- on [the type of option](/config/options/):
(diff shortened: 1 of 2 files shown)


joostdecock (approved):
Nice work Ben 👍

PR #3191, opened 2022-12-09
chore(deps) fix dependency issues

eriese:
- add necessary i18next peer dependencies to lab to get rid of errors
- pin next version to 13.0.5 because 13.0.6 seems to be missing the bin script
- do a little cleanup in the lab's dependencies while I'm in there

### DIFF
--- a/sites/dev/package.json
+++ b/sites/dev/package.json
@@ -37,7 +37,7 @@
     "lodash.get": "^4.4.2",
     "lodash.orderby": "^4.6.0",
     "lodash.set": "^4.3.2",
-    "next": "13.0.5",
+    "next": "^13.0.6",
     "react": "^18.2.0",
     "react-copy-to-clipboard": "^5.0.4",
     "react-dom": "^18.2.0",
--- a/sites/dev/package.json
+++ b/sites/dev/package.json
@@ -37,7 +37,7 @@
     "lodash.get": "^4.4.2",
     "lodash.orderby": "^4.6.0",
     "lodash.set": "^4.3.2",
-    "next": "^13.0.6",
+    "next": "^13.0.0",
     "react": "^18.2.0",
     "react-copy-to-clipboard": "^5.0.4",
     "react-dom": "^18.2.0",
--- a/sites/dev/package.json
+++ b/sites/dev/package.json
@@ -37,7 +37,7 @@
     "lodash.get": "^4.4.2",
     "lodash.orderby": "^4.6.0",
     "lodash.set": "^4.3.2",
-    "next": "^13",
+    "next": "13.0.5",
     "react": "^18.2.0",
     "react-copy-to-clipboard": "^5.0.4",
     "react-dom": "^18.2.0",
--- a/sites/lab/package.json
+++ b/sites/lab/package.json
@@ -31,7 +31,7 @@
   "dependencies": {
     "@headlessui/react": "^1.6.6",
     "i18next": "^22.1.4",
-    "next": "13.0.5",
+    "next": "^13.0.6",
     "next-i18next": "^13.0.0",
     "react": "^18.2.0",
     "react-i18next": "^12.1.1",
--- a/sites/lab/package.json
+++ b/sites/lab/package.json
@@ -31,7 +31,7 @@
   "dependencies": {
     "@headlessui/react": "^1.6.6",
     "i18next": "^22.1.4",
-    "next": "latest",
+    "next": "13.0.5",
     "next-i18next": "^13.0.0",
     "react": "^18.2.0",
     "react-i18next": "^12.1.1",
--- a/sites/lab/package.json
+++ b/sites/lab/package.json
@@ -30,41 +30,14 @@
   },
   "dependencies": {
     "@headlessui/react": "^1.6.6",
-    "@heroicons/react": "^2.0.1",
-    "@mdx-js/loader": "^2.0.0-rc.2",
-    "@mdx-js/mdx": "^2.0.0-rc.2",
-    "@mdx-js/react": "^2.0.0-rc.2",
-    "@mdx-js/runtime": "next",
-    "@tailwindcss/typography": "^0.5.0",
-    "algoliasearch": "^4.11.0",
-    "d3-dispatch": "^3.0.1",
-    "d3-drag": "^3.0.0",
-    "d3-selection": "^3.0.0",
-    "daisyui": "^2.0.6",
-    "lodash.get": "^4.4.2",
-    "lodash.orderby": "^4.6.0",
-    "lodash.set": "^4.3.2",
+    "i18next": "^22.1.4",
     "next": "latest",
     "next-i18next": "^13.0.0",
-    "react-copy-to-clipboard": "^5.0.4",
-    "react-hotkeys-hook": "^4.0.4",
-    "react-instantsearch-dom": "^6.18.0",
-    "react-markdown": "^8.0.0",
-    "react-swipeable": "^7.0.0",
-    "react-timeago": "^7.1.0",
-    "rehype-highlight": "^6.0.0",
-    "rehype-sanitize": "^5.0.1",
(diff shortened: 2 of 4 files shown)


joostdecock (commented):
We might need to think about how we manage these dependencies between `shared` and the actual sites. But you've removed at least some stuff here that's actually used by the lab  (tailwind & daisyui for example).

I suspect things will continue to work because those dependencies are *also* in shared. But it feels counter-intuitive to me to have dependencies *hidden* this way.

Thoughts?

sites/dev/package.json
```diff
@@ -37,7 +37,7 @@
     "lodash.get": "^4.4.2",
     "lodash.orderby": "^4.6.0",
     "lodash.set": "^4.3.2",
-    "next": "^13",
+    "next": "13.0.5",
```
nchilada, line 40:
```suggestion
    "next": "^13.0.6",
```

sites/lab/package.json
```diff
@@ -30,41 +30,14 @@
...
-    "d3-dispatch": "^3.0.1",
-    "d3-drag": "^3.0.0",
-    "d3-selection": "^3.0.0",
-    "daisyui": "^2.0.6",
-    "lodash.get": "^4.4.2",
-    "lodash.orderby": "^4.6.0",
-    "lodash.set": "^4.3.2",
-    "next": "latest",
+    "i18next": "^22.1.4",
+    "next": "13.0.5",
```
nchilada, line 34:
```suggestion
    "next": "^13.0.6",
```

sites/org/package.json
```diff
@@ -40,7 +40,7 @@
     "lodash.get": "^4.4.2",
     "lodash.orderby": "^4.6.0",
     "lodash.set": "^4.3.2",
-    "next": "latest",
+    "next": "13.0.5",
```
nchilada, line 43:
```suggestion
    "next": "^13.0.6",
```

nchilada (commented):
Oh good, I'm not the only one who was having trouble with lint in `sites/dev`, `sites/lab`, `sites/org`. Thanks for figuring this out already! (I found this PR by looking for repo actions with passing lint. 😄) And this is a little belated, but thank you @eriese for cleaning up all the lint issues.

Check out my suggested changes though — the problem wasn't with `next@13.0.6`, but something specific to `lerna bootstrap` or yarn workspaces. As you discovered, it looks like we simply need to get our three dependencies to perfectly match each other. (No-hoist installation would probably suffice too, or switching to `pnpm`.)

eriese:
> We might need to think about how we manage these dependencies between `shared` and the actual sites. But you've removed at least some stuff here that's actually used by the lab (tailwind & daisyui for example).
> 
> I suspect things will continue to work because those dependencies are _also_ in shared. But it feels counter-intuitive to me to have dependencies _hidden_ this way.
> 
> Thoughts?

I ran a dependency checker and it said these libraries weren't being imported in the sites, so I figured it was harmless to remove them, but I see your point that the markup relies on the assumption that you have tailwind and daisyui installed... maybe the better way is to have them be peer dependencies of `shared` and dependencies of the sites? I'm also fine to just put them back in there. An earlier attempt to declutter the dependencies showed me that I am not personally prepared to get very deep into making sure everything is only and exactly where it should be.


> the problem wasn't with `next@13.0.6`, but something specific to `lerna bootstrap` or yarn workspaces. As you discovered, it looks like we simply need to get our three dependencies to perfectly match each other. (No-hoist installation would probably suffice too, or switching to `pnpm`.)
> 
> Edit: oh, I guess you'd need to update `yarn.lock` even after accepting these suggestions — sorry! And I might still be wrong about next 13.0.5 vs. 13.0.6, maybe you saw something else broken in the latter which I'm not aware of.

What I found was that the `bin` file from `next@13.0.6` wasn't even being copied to the root `node_modules/bin`, so I stopped trying to troubleshoot the hoisting, which I know nothing about, and just went with the most recent version of next that did have a `bin` file copied. I believe that it not being copied in the root means that it's not a hoisting issue, but again, very little knowledge there. I do know that `lerna bootstrap --no-hoist` didn't help.

I could be mistaken but the only suggestions that I'm seeing are changing the `next` versions back, which leaves us without a `next` command again. What's the fix you're proposing?

nchilada:
> What I found was that the `bin` file from `next@13.0.6` wasn't even being copied to the root `node_modules/bin`, so I stopped trying to troubleshoot the hoisting, which I know nothing about, and just went with the most recent version of next that did have a `bin` file copied. I believe that it not being copied in the root means that it's not a hoisting issue, but again, very little knowledge there. I do know that `lerna bootstrap --no-hoist` didn't help.
> 
> I could be mistaken but the only suggestions that I'm seeing are changing the `next` versions back, which leaves us without a `next` command again. What's the fix you're proposing?

Originally the versions were `"^13"`, `"latest"`, and `"latest"` in dev, lab, and org respectively, which I agree was causing problems. The package was being hoisted to `./node_modules/next`, including its binary in `./node_modules/next/dist/bin/next`, but lerna/yarn wasn't creating a symlink for the binary anywhere on the Node path.

I found that requiring `"^13.0.6"`, `"^13.0.6"`, and `"^13.0.6"` respectively solves the problem. After wiping everything in my working tree (`yarn cleanall && lerna exec "rm -rf node_modules" && rm -rf node_modules`) and reinstalling (`yarn kickstart`), I now see `./node_modules/.bin/next`, `./sites/dev/node_modules/.bin/next`, `./sites/lab/node_modules/.bin/next`, and `./sites/org/node_modules/.bin/next`.

eriese:
> Originally the `next` versions were `"^13"`, `"latest"`, and `"latest"` in dev, lab, and org respectively, which I agree was causing problems. The package was being installed and hoisted to `./node_modules/next/`, with a binary at `./node_modules/next/dist/bin/next`, but lerna/yarn wasn't creating a symlink for the binary anywhere on the Node path.
> 
> I found that requiring `"^13.0.6"`, `"^13.0.6"`, and `"^13.0.6"` respectively solves the problem. After wiping everything in my working tree (`yarn cleanall && lerna exec "rm -rf node_modules" && rm -rf node_modules`) and reinstalling (`yarn kickstart`), I now see symlinks at `./node_modules/.bin/next`, `./sites/dev/node_modules/.bin/next`, `./sites/lab/node_modules/.bin/next`, and `./sites/org/node_modules/.bin/next`.
> 

oh I see! ok, that's fixed now then. Agree it's better not to pin it to a specific patch, so I'm glad that works. It makes a certain amount of sense to me that everyone having the same version would unbreak this, but I don't have a real technical explanation for it... something about it matching enough for the bootstrapping to not install a different version but not matching quite enough to have its scripts hoisted? A guess

nchilada:
> It makes a certain amount of sense to me that everyone having the same version would unbreak this, but I don't have a real technical explanation for it... something about it matching enough for the bootstrapping to not install a different version but not matching quite enough to have its scripts hoisted? A guess

Haha, no idea. I bet any of us could figure it out by stepping through `yarn`, but ultimately I think I want less 🧶 in my brain, not more… 😄

joostdecock:
FYI: I've added how we manage site dependencies to the contributor call agenda for today.

eriese:
> FYI: I've added how we manage site dependencies to the contributor call agenda for today.

In the mean time, I've added back the dependencies I removed from lab, so I think we should merge this and put what we discussed about dependency management on the TODO list

nchilada:
Sorry for missing the call! I forgot about those, will rejoin any month now 😆

Not sure what you all concluded about site dependencies, but I agree it's generally best not to rely on hoisting for either binaries or imports. Though some monorepos seem to explicitly hoist their dependencies, especially `devDependencies`, by explicitly adding them to the `package.json` in the root, which is probably safe but unintuitive. Interestingly, I've also seen an open feature request (see yarnpkg/yarn # 4964) to have dependencies' binaries be automatically symlinked in the dependent package's binaries; but that behavior seems icky and I hope it doesn't actually get implemented.

PR #3222, opened 2022-12-13
fix(docs): Plugins documentation updates

BenJamesBen:
This PR is a somewhat large update to dev reference Plugins-related documentation.

Other than corrections, the most significant changes are:
- Add import usage for each plugin.
- More information about conditional plugins and how to use them. This is especially seen in the `plugin-bust` changes.
- More information about how to use `plugin-i18n`.
- More information about `plugin-svgattr` and how to use it.
- Alphabetize plugins in `plugin-bundle`, to make it easier to see what plugins it includes.

Hopefully my understanding of conditional plugins, `plugin-i18n`, and `plugin-svgattr` and the information I added is correct.

Not included in this PR and still to be done:
- Add documentation for `plugin-cutlist` and `plugin-timing`
- Doc review for the Plugin Guide.

### DIFF
--- a/markdown/dev/reference/api/part/config/plugins/en.md
+++ b/markdown/dev/reference/api/part/config/plugins/en.md
@@ -17,7 +17,7 @@ const part = {
 ```
 <Tip>
 
-You should only list the plugins that required by the part itself, 
+You should only list the plugins that are required by the part itself,
 not those required by its dependencies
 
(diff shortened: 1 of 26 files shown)


joostdecock (approved):
I have put up a question about the double named exports on Discord, but this is great in any case.

PR #3238, opened 2022-12-15
fix(docs): Settings documentation updates

BenJamesBen:
I've included the previously-undocumented `settings.debug`, though it is unclear whether it was intended to exist. It doesn't seem to be used in `core` or in any designs.


joostdecock (approved):

joostdecock:
I've merged this, as there's a lot of good changes. However, the `settings.debug` is an artifact from v2 that should be removed. As you noticed, it's not (or no longer) used by core or the designs

PR #3239, opened 2022-12-15
fix(docs): Add context property to Part.draft()

BenJamesBen:
I am not sure how to explain or document what `context` is. However, it exists and should be documented (unless the intention was that is supposed to be undocumented). 

I note that `context` is the only way to access `settings.debug`, as I understand it.


joostdecock (approved):
The part `context` allows access to things higher in the tree, such as the settings, the pattern object itself, and so on.

It is a way to gain access to those things when all you have is the part object.

PR #3247, opened 2022-12-16
Yaml Edit View

eriese:
Expected behavior:

- User can visit the yaml editor view even if they have not entered measurements
- User can paste exported yaml from freesewing.org into the editor and go to the draft view without any errors
- User can manually edit and save yaml to the gist. The only guardrail is that all needed settings and the view state will automatically be added back as the default if they are removed
- User will be notified of likely issues with their yaml input even though it is saved

### DIFF
--- a/packages/i18n/src/locales/en/app.yaml
+++ b/packages/i18n/src/locales/en/app.yaml
@@ -345,3 +345,5 @@ accessoryDesigns: Accessory Designs
 blockDesigns: Block/Sloper Designs
 garmentDesigns: Garment Designs
 utilityDesigns: Utility Designs
+editError: Issues with YAML
+editErrorDesc: We saved your input, but it might not work for the following reasons
--- a/packages/i18n/src/locales/en/app.yaml
+++ b/packages/i18n/src/locales/en/app.yaml
@@ -345,5 +345,3 @@ accessoryDesigns: Accessory Designs
 blockDesigns: Block/Sloper Designs
 garmentDesigns: Garment Designs
 utilityDesigns: Utility Designs
-editError: Issues with YAML
-editErrorDesc: We saved your input, but it might not work for the following reasons
--- a/packages/i18n/src/locales/en/components/workbench.yaml
+++ b/packages/i18n/src/locales/en/components/workbench.yaml
@@ -5,4 +5,7 @@ layoutThing: 'Layout {thing}'
 pageSize: Page size
 startBySelectingAThing: 'Start by selecting a {thing}'
 testThing: 'Test {thing}'
-editThingTitle: 'Edit Pattern Configuration for {thing}'
+editView:
+  titleThing: 'Edit Pattern Configuration for {thing}'
+  error: Issues with YAML
+  errorDesc: We saved your input, but it might not work for the following reasons
--- a/packages/i18n/src/locales/en/components/workbench.yaml
+++ b/packages/i18n/src/locales/en/components/workbench.yaml
@@ -5,7 +5,6 @@ layoutThing: 'Layout {thing}'
 pageSize: Page size
 startBySelectingAThing: 'Start by selecting a {thing}'
 testThing: 'Test {thing}'
-editView:
-  titleThing: 'Edit Pattern Configuration for {thing}'
-  error: Issues with YAML
-  errorDesc: We saved your input, but it might not work for the following reasons
+yamlEditViewTitleThing: 'Edit Pattern Configuration for {thing}'
+yamlEditViewError: Issues with YAML
+yamlEditViewErrorDesc: We saved your input, but it might not work for the following reasons
--- a/packages/i18n/src/locales/en/components/workbench.yaml
+++ b/packages/i18n/src/locales/en/components/workbench.yaml
@@ -5,3 +5,4 @@ layoutThing: 'Layout {thing}'
 pageSize: Page size
 startBySelectingAThing: 'Start by selecting a {thing}'
 testThing: 'Test {thing}'
+editThingTitle: 'Edit Pattern Configuration for {thing}'
--- a/sites/shared/components/workbench/edit.js
+++ b/sites/shared/components/workbench/edit.js
@@ -6,9 +6,6 @@ import Popout from 'shared/components/popout.js'
 import { useTranslation } from 'next-i18next'
 import { capitalize } from '@freesewing/core'
 
-/** count the number of lines in some text */
-const countLines = (txt) => txt.split('\n').length
-
 /** a view for editing the gist as yaml */
 const Edit = (props) => {
   let { gist, setGist, gistReady } = props
--- a/sites/shared/components/workbench/edit.js
+++ b/sites/shared/components/workbench/edit.js
@@ -1,9 +1,7 @@
 import yaml from 'js-yaml'
-import hljs from 'highlight.js/lib/common'
 import defaultSettings from './default-settings'
 import validateGist from './gist-validator'
 import { useEffect, useState, useRef } from 'react'
-import Json from 'shared/components/json-highlight.js'
 import Popout from 'shared/components/popout.js'
 import { useTranslation } from 'next-i18next'
 import { capitalize } from '@freesewing/core'
@@ -20,7 +18,7 @@ const Edit = (props) => {
   const [gistAsYaml, setGistAsYaml] = useState(null)
   // the number of lines in the yaml
   const [numLines, setNumLines] = useState(0)
-  // any errors, parsed by hljs
+  // any errors as a json string
   const [error, setError] = useState(null)
   // success notifier
   const [success, setSuccess] = useState(null)
@@ -31,7 +29,7 @@ const Edit = (props) => {
   useEffect(() => {
     if (gistReady) {
       // get everything but the design because it's a function and can't be serialized
-      const { design, ...gistRest } = gist
+      const { _design, ...gistRest } = gist
       setGistAsYaml(yaml.dump(gistRest))
     }
   }, [gist, gistReady])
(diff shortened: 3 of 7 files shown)


BenJamesBen:
I did some initial testing. More testing to come.

I was able to cause an error:
- Started up the localhost lab, loaded Titan.
- I loaded one of the model measurements and clicked Edit YAML. An editing window appeared as expected.
- I pressed the ClearThing link and then Edit YAML. I got an Unhandled Runtime Error, `YAMLException: unacceptable kind of an object to dump [object Function]` at `../shared/components/workbench/edit.js`, line 34, `setGistAsYaml(yaml.dump(gistRest))`

Other observations:
- I was able to add nonsense text to `renderer:`, and `units:` without any warnings. `renderer: reactasdfasdfasdf` I believe this is the expected behavior at the moment?

- I saved an Aaron YAML from freesewing.org with a custom seam allowance `sa: 23.7`. It loaded just fine in the localhost lab, but this version seems to use `saMm: 10`, so my custom seam allowance was ignored. I'm not sure if this is just a v2-to-v3 API change.

- I saved an org Aaron YAML with options changes, and they loaded into the localhost lab and worked just fine.

- I saved an org Aaron YAML with `only: `, `  -back`, loaded it into localhost lab, but no parts were seen in DraftDesign view. Removing those 2 lines from the YAML fixed the issue, causing all parts to re-appear in DraftDesign. I note that the v3 lab saves the part as `  - aaron.back`, so this is also probably a v2-to-v3 API change.

- `paperless`, `margin`, `units`, and `scale` seem to save, load, and work just fine from org to localhost lab.

BenJamesBen:
More experimentation:

- I edited YAML changing to `  view: editxxx` and saved. I got a page with a blank content section except for the word `TODO`.

- Just a minor UI thing: I note that if I type a very long line into the YAML edit window, the text wraps in the window but the line numbers don't adjust. At first glance it looks like the 1 long line is 3 lines.

- Duplicate keys are caught by the YAML-checker.

- I see now that the localhost lab YAML has both an `sa` and `saMm` property.

- My localhost lab got into a state where I would get that `YAMLException: unacceptable kind of an object to dump [object Function]` error after loading model measurements and trying to Edit YAML. The error was repeatable. Restarting `yarn lab` seemed to clear up the error state.

- I believe that ClearThing might be what initiates the state that allows the error to occur.

- My org YAML had `design: benjamin`, but after loading it into localhost, the YAML was missing the `design:` property.

- Regarding that runtime error, I notice that the YAML in the non-bug state has `design: '@freesewing/benjamin'`, but the `design:` property is missing in the bug state. (As seen when using the "YAML" view.)

- There's another API mismatch where enabled seam allowance correctly loads from org YAML. However, the localhost UI incorrectly says "Include seam allowance" is "No" because of the v3 `saBool: false` in the v3 YAML.

- I tried breaking options, but the checker caught nonexistent option names and not-in-range options.

BenJamesBen (changes requested):
I believe that the Unhandled Runtime Error issues should be fixed before this PR is approved.

BenJamesBen:
When I load an org YAML with Imperial units and a seam allowance set to any value, for example
```yaml
settings:
  sa: 15.9
  units: imperial
```
... the localhost lab UI incorrectly says that the Seam allowance size is `"25/64`. I think this is another sa/saMm/saBool artifact.

BenJamesBen:
I've done more testing on this, loading YAML from a number of those "Error while drafting <design>" bug reports on GitHub.

Aside from these two previously-reported issues, no problems were seen:
1. Need to change part names manually from v2 `front` to v3 `charlie.front`.
2. The disappearing `design:` key issue.

packages/i18n/src/locales/en/app.yaml
```diff
@@ -341,3 +341,5 @@ darkTheme: Dark Theme
 hax0rTheme: Hax0r Theme
 lgbtqTheme: LGBTQ Theme
 transTheme: Trans Theme
+editError: Issues with YAML
```
joostdecock, line 344: The `editError` and `editErrorDesc` are not great translation keys.
There's several places in the UI where things can be edited and where that could cause an error.
This, in combination with the fact that this is added to the sort of catch-all `app.yaml` makes me feel we should:
- Move it to a more specific file since it should only ever be loaded on the Edit YAML view
- Give it a more descriptive name

packages/i18n/src/locales/en/components/workbench.yaml
```diff
@@ -5,3 +5,4 @@ layoutThing: 'Layout {thing}'
 pageSize: Page size
 startBySelectingAThing: 'Start by selecting a {thing}'
 testThing: 'Test {thing}'
+editThingTitle: 'Edit Pattern Configuration for {thing}'
```
joostdecock, line 8: Whenever I use `thing` it's typically because it allows replacing `thing` with something else. Like Edit `username`, or Edit `pattern`, or Edit `settings`, or edit `newsletter preferences`.

But there we're setting `editThingTitle` to something specific to the pattern configuration.
I'm afraid that would be counter-intuitive in those other scenarios where `editThing` is used.

I appreciate that managing translation keys across the project is not so obvious. But I'd like to see a more descriptive key here. Like `editPatternConfigThingTitle`.

sites/shared/components/workbench/edit.js
```diff
@@ -0,0 +1,143 @@
...
+
+/** a view for editing the gist as yaml */
+const Edit = (props) => {
+  let { gist, setGist, gistReady } = props
+
+  const inputRef = useRef(null)
+  // the gist parsed to yaml
+  const [gistAsYaml, setGistAsYaml] = useState(null)
+  // the number of lines in the yaml
+  const [numLines, setNumLines] = useState(0)
```
joostdecock, line 20: Do we care about the line numbers?
By which I mean, are we going to say things like: One line X you are missing blabla?

Because if we just want to show them, I'd prefer to handle that in CSS.

sites/shared/components/workbench/gist-validator.js
```diff
@@ -0,0 +1,104 @@
...
+    this.validateSettings()
+    this.validateMeasurements()
+    this.validateOptions()
+
+    return { valid: this.valid, errors: this.errors }
+  }
+}
+
+/** make and run a gist validator */
+export default function validateGist(givenGist, design) {
```
joostdecock, line 101: Should we worry about the performance penalty of instantiating this on every call?

joostdecock (changes requested):
We can probably merge this as-is for now and worry about it later.

The only changes I'd like to see if the translation keys. Especially to keep 'em out of `app.yaml` as I'm trying to get rid of that eventually

sites/shared/components/workbench/gist-validator.js
eriese: I don't think so because it only validates on save, which is an onClick event, but I'll make a reusable instance

packages/i18n/src/locales/en/components/workbench.yaml
eriese: How do we feel about this?

```yaml
editView:
  titleThing: 'Edit Pattern Configuration for {thing}'
  error: Issues with YAML
  errorDesc: We saved your input, but it might not work for the following reasons
```

joostdecock: We juggle these yaml files in various places, and we expect a flat key/value structure.
This would break that, so please let's not.

Can we not just do something like:

- `patternEditViewTitle`
- `patternEditViewError`
- `patternEditViewErrorDesc`

?

sites/shared/components/workbench/edit.js
joostdecock: I can add that. It's not worth spending much time on :)

sites/shared/components/workbench/edit.js
```diff
@@ -0,0 +1,115 @@
...
+
+  /** user-initiated save */
+  const onSave = () => {
+    // clear the errors
+    setError(null)
+    try {
+      // parse back to json
+      const editedAsJson = yaml.load(inputRef.current.value)
+      // make it backwards compatible so that people can paste in the yaml export from org
+      // the yaml export from org is missing some of the settings that are needed in the gist, and what it does have is under 'settings', so we merge that stuff with the existing gist view state and the default settings to make sure all necessary keys are accounted for, but we're not keeping stuff that was supposed to be cleared
```
BenJamesBen, line 48: I don't like to complain too much about style, but this line is extremely long.

BenJamesBen (dismissed):
I tested the updated PR, and the localhost lab worked fine for me.

I re-tested the Edit YAML functionality. The Unhandled Runtime Error is no longer seen. All of the v2->v3 issues are still seen, but I believe this is expected.

packages/i18n/src/locales/en/components/workbench.yaml
eriese: For sure, sorry. I hate coming up with translation keys

eriese: I went with `yamlEditView` etc, and I did leave it as `yamlEditViewTitleThing` because it has a thing in it. Hope this is acceptable!

BenJamesBen (dismissed):
I retested the PR in my localhost lab, and no regressions were seen.

packages/i18n/src/locales/en/components/workbench.yaml
eriese: @joostdecock should I take this still being here as a sign that these are still not acceptable, or just that it got buried?

eriese:
@joostdecock bump

eriese:
Right now because ignored builds are still broken, this hasn't been built by vercel so it's not testable that way, but in the mean time, @joostdecock do you want me to sidecar these translations?

joostdecock:
> @joostdecock bump

:see_no_evil: I feel called out :joy: 

We need to merge this, because the longer a PR is open, the harder it gets to merge. (which, is on me)
So let's get it done and we can always tweak it later.

joostdecock (approved):

PR #3251, opened 2022-12-17
Manage site dependencies centrally

joostdecock:
This PR implements what we discussed on the latest contributor call:

- Dependencies are centrally managed (like for the rest of the monorepo, in `config/dependencies.yaml`)
- The `package.json` files for all sites are auto-generated when running `yarn reconfigured`
- The READMEs for the sites are also auto-generated (didn't do much work on this).

The main benefit of this is that it should be a lot easier to let everything work together without version conflicts.
It will also reduce the number of packages to be installed since we're making sure we don't keep multiple versions.

Did some initial tests building some sites, seems fine. However, I did remove some dependencies in the process (because I think they are no longer used) so YMMV.


joostdecock:
Sorry for the many changed to the contributors list.

BenJamesBen:
I tested this PR on a fresh clone of `freesewing/freesewing.git`, and things seem to work fine. I performed:
```sh
$ cd workspace
$ git switch site-dependencies
$ yarn kickstart
$ yarn lab
$ # The localhost lab worked fine. I tested out a few designs.
$ cd sites/dev
$ yarn start
$ # The localhost dev website loaded fine.
```

(On the localhost dev website, I did see the `TypeError: _jsxDEV is not a function` Unhandled Runtime Error when browsing to any page other than the index. However, that is a known issue affecting `develop` and not related to this PR.)

Running localhost lab and the dev website were the only two tasks I previously knew about. However, I also tried the following, which was new to me:
```sh
$ cd ../../sites/org
$ yarn start
$ # The localhost org website started up.
```

(When the localhost org website started up, I immediately saw that Error: `Invalid <Link> with <a> child. Please remove <a> or use <Link legacyBehavior>.` error. It appears that the org web pages have not yet been updated to work with Next 13. I was able to edit files to get the index page to load without errors.  I'm seeing various other errors on other org web pages, but I think those might be due to other updates that need to be made to org web pages.)

BenJamesBen:
The org web page errors were:

designs:
```
./pages/designs/index.js:6:0
Module not found: Can't resolve 'shared/config/designs.json'
> 6 | import designs from 'shared/config/designs.json' assert { type: 'json' }
```

showcase:
```
./pages/showcase/index.js:6:0
Module not found: Can't resolve 'shared/utils.js'
> 6 | import { strapiImage } from 'shared/utils.js'
```

docs:
```
Error: Required "jargon" option is missing in remark-jargon configuration
../packages/rehype-jargon/src/index.mjs (10:10) @ Error
   8 | // Don't bother if we don't have any jargon
   9 | if (!options || !options.jargon) {
> 10 |   throw Error('Required "jargon" option is missing in remark-jargon configuration')
```

joostdecock:
Thanks @BenJamesBen 
Yes, the `org` site is currently not getting a lot of attention, so it's lagging behind a bit. 
I'm working on deploying a v3 backend (see #3253) and when that's done will turn my attention to the `org` site.

I will merge this from CLI to side-step the yarn.lock merge conflict.

Issue #3267, opened 2022-12-19
[bug]: Development environment does not ship with prebuild translation files (Error: Default namespace not found at public/locales/en/common.json)


joostdecock:
### What seems to be the problem? 🤔

When spinning up the development environment, it throws an error:

**Error: Default namespace not found at public/locales/en/common.json**

It seems that we're not shipping the pre-build translation files.

### Design / Plugin / Package 🧐

packages/new-design

### Are you a FreeSewing patron? 😃

Yes, I am a tier-8 patron ❤️  💙 💜

### Additional context

See PR #3268 for this

eriese:
I think it might be due to this: https://github.com/freesewing/freesewing/commit/ad148eb59dedad7c483df5dfb393c94c96c34f2b#diff-1d42aaeda0f4874eb7852e37861d6c520dea972c722c5e13d080f022ad4c6e83L66

We used to build the package and run the `bin` from `dist`, but now we just run the script. Which I approve of, but means we're going to need to run that i18n prebuild step at runtime

eriese:
Not currently seeing this when running `npx @freesewing/new-design@next` for `brian`, `scratch`, or `tutorial`

PR #3268, opened 2022-12-19
fix(new-design): Put development environment back on the rails

joostdecock:
There's been changes recently, both to the lab environment as well as the template files, and they have broken the development environment. Or at least the v3 alpha version.

The first fix is the easy one: Update the list of files to fetch post-install.

But after that, it seems we're also not shipping the pre-build translation files which is causing the deveopment environment to fail.
I'm not certain whether this is related to the changes in the templates, but I've created issue #3267 for this.


joostdecock:
Merged in the `alpha-release` branch. Can be closed

PR #3274, opened 2022-12-21
fix(github): Use shortcodes for labels emoji

BenJamesBen:
This PR changes the GitHub issues templates to use shortcodes for emoji, so the labels get automatically populated on new issues.

(My commit changes only 1 line in each file. All of the other changes were made by `eslint`.)


eriese (approved):
Something's wrong with our yaml linter, but that's outside the scope of this PR

PR #3275, opened 2022-12-21
Teagan inward curve at waist

anna-puk:
New option for Teagan: curve inward at the waist, creating an hourglass shape.

Also includes a fix for the 'paperless' option (suggested by @BenJamesBen).


designs/teagan/src/back.mjs
```diff
@@ -36,6 +47,7 @@ function teaganBack({
   paths.seam = new Path()
     .move(points.cfHem)
     .join(paths.hemBase)
+    .join(paths.sideSeam)
```
BenJamesBen, line 50: Seam allowance is not being calculated properly for the new side seam. With Curve To Waist disabled, the seam allowance is too close to the fabric edge, and the seam allowance doesn't change when the option is enabled, so it can be too wide.
Screenshot 2022-12-22 at 5 58 40 AM
Screenshot 2022-12-22 at 5 58 59 AM

designs/teagan/src/front.mjs
```diff
@@ -181,6 +231,8 @@ export const front = {
     backNeckCutout: { pct: 8, min: 4, max: 12, menu: 'fit' },
     // Teagan specific
     draftForHighBust: { bool: false, menu: 'fit' },
+    curveToWaist: { bool: false, menu: 'fit' },
+    waistEase: { pct: 25, min: 8, max: 40, menu: 'fit' },
```
BenJamesBen, line 235: You might want to consider renaming the `waistEase` option name to something like `curvedWaistEase` to make it more obvious that it affects only curved waists. The documentation and translation will help with this, but if the translation string goes untranslated then changing the name would help avoid confusion.

(Or, you could change the code to always use `waistEase` in the width calculations.)

BenJamesBen (commented):
I tested the code changes in the localhost lab, and the new feature seems to work with the variety of different measurements I tested. Other than the seam allowance issues, the code changes themselves would be okay to put back for further testing.

The `waistEast` documentation and i18n translations are missing for the new options.

BenJamesBen:
The `paperless` code changes are fine to keep in this PR. 

But, FYI, per https://github.com/freesewing/freesewing/pull/3285 Joost intend to revert the removal of the `Part.getId()` method that caused the error. (The newly-added shorthand `getId` will still remain, so that part of your PR will still be fine as is.)

designs/teagan/src/back.mjs
anna-puk: Thanks for pointing this out, I've fixed it.

designs/teagan/src/front.mjs
anna-puk: That's a good idea! Done.

anna-puk:
I had renamed the branch, but apparently committed the changes to the old name... I've merged them in now.

BenJamesBen:
I re-tested the changes in the lab, and I again believe that this PR is okay to merge for further/wider testing.

An issue I found is that the side seam curve can dip below the waistline when negative lengthBonus settings are used:
Screenshot 2023-01-30 at 9 39 36 AM

BenJamesBen (approved):

BenJamesBen (approved):
I re-tested the updated PR in the lab. With the same comments as before, I again believe that this PR is okay to merge for further/wider testing.

joostdecock (approved):

Issue #3282, opened 2022-12-22
[docs]: Resolve Part.getId() method documentation confusion


BenJamesBen:
### Where can we see the problem? 🤔

https://freesewing.dev/reference/api/part/getid

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Suggested content

The `Part.getId()` method was removed as a prototype of `Part`, so normally the documentation should also be removed.

The method was moved into part shorthand that can be destructured as `getId`, and this was already added to the `Part.draft()` documentation. However, the `Part.draft()` documentation just refers users back to the `Part.getId()` method documentation.

We should either remove the `Part.getId()` method documentation and improve the `Part.draft()` documentation to explain what `getID` does. Or, keep the `Part.getId()` method documentation but improve it to specify that it isn't actually a prototype method.

Removing the method doc page is the right thing to do, but it will cause confusion since `Part.draft()` isn't set up to explain `getId`. Keeping the method doc page might be the wrong thing to do, but it would better explain `getId` to users, though the existence of the documentation page is likely to confuse users since `getId()` isn't a prototype method.

Finally, we might want to consider re-adding the `Part.getId()` prototype method, the same way that `hide` and `unhide` shorthand and their prototype methods co-exist. That way no documentation changes would be needed.

joostdecock:
@BenJamesBen makes sense to add it (again) as a prototype method. I've done so in #3285

BenJamesBen:
PR https://github.com/freesewing/freesewing/pull/3285 was merged, adding back the `Part.prototype.getId()` method and resolving this bug. Closing.

PR #3285, opened 2022-12-22
feat(core): Make Part.getId() a prototype method

joostdecock:
As requested by @BenJamesBen in #3282

### DIFF
--- a/packages/core/src/part.mjs
+++ b/packages/core/src/part.mjs
@@ -79,6 +79,17 @@ Part.prototype.attr = function (name, value, overwrite = false) {
 }
 
 /**
+ * Gets a free ID to use in the part
+ *
+ * @return {string} id - A free ID to use
+ */
+Part.prototype.getId = function (prefix = '') {
+  this.freeId += 1
+
+  return prefix + this.freeId
+}
+
+/**
  * Hide the part
  *
  * @return {Part} part - The Part instance
--- a/packages/core/src/part.mjs
+++ b/packages/core/src/part.mjs
@@ -81,6 +81,7 @@ Part.prototype.attr = function (name, value, overwrite = false) {
 /**
  * Gets a free ID to use in the part
  *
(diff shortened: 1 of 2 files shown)


packages/core/src/part.mjs
```diff
@@ -78,6 +78,17 @@ Part.prototype.attr = function (name, value, overwrite = false) {
   return this
 }
 
+/**
+ * Gets a free ID to use in the part
+ *
```
BenJamesBen, line 83: From the code that was previously removed, this line is missing in the comment: 
```
 * @param {string} prefix - An optional prefix to apply to the ID
```

BenJamesBen (dismissed):

packages/core/src/part.mjs
```diff
@@ -78,6 +78,17 @@ Part.prototype.attr = function (name, value, overwrite = false) {
...
+ * @return {string} id - A free ID to use
+ */
+Part.prototype.getId = function (prefix = '') {
+  this.freeId += 1
+
+  return prefix + this.freeId
```
eriese, line 89: I'm not liking that this is repeated code from
```
/**
 * Returns a closure holding a getId method (returns an ID unused in this part)
 *
 * @return {function} getId - The getId function
 */
Part.prototype.__getIdClosure = function () {
  const self = this
  const method = function (prefix = '') {
    self.freeId += 1

    return prefix + self.freeId
  }

  return method
}
```

One has to reference the other. Your choice which

eriese (changes requested):

packages/core/src/part.mjs
joostdecock: Added. Good catch

packages/core/src/part.mjs
joostdecock: I've called the closure instead.

BenJamesBen (approved):

PR #3304, opened 2022-12-23
fix(core): Improve error messages for Firefox and Safari browsers

BenJamesBen:
JavaScript's `Error.prototype.stack` is a non-standard feature, and different browser JavaScript engines produce error stacks of varying content. 

Chrome produces a stack with useful information:
Screenshot 2022-12-21 at 1 01 48 PM

However, Firefox and Safari's stacks are less useful, specifically missing the `TypeError: part.getId is not a function` (the value of the error's `toString()`:
image
Screenshot 2022-12-21 at 4 06 04 PM

This PR adds the `TypeError: part.getId is not a function` toString() output to the stack if it is not already there. This improves the error when seen in the logs. For example, on Safari:
Screenshot 2022-12-21 at 5 05 44 PM

References:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack

### DIFF
--- a/packages/core/src/pattern.mjs
+++ b/packages/core/src/pattern.mjs
@@ -172,6 +172,12 @@ Pattern.prototype.draftPartForSet = function (partName, set) {
       }
       return result
     } catch (err) {
+      // If the browser's JavaScript engine doesn't include the error's
+      // "name: message" string at the top of the stack, manually push it
+      // onto the stack so the information is shown in the error log.
+      if (!err.stack.startsWith(err.toString())) {
+        err.stack = err.toString() + '\n' + err.stack
+      }
       this.setStores[set].log.error([`Unable to draft part \`${partName}\` (set ${set})`, err])
     }
   } else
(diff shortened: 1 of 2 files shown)


packages/core/src/pattern.mjs
```diff
@@ -172,6 +172,12 @@ Pattern.prototype.draftPartForSet = function (partName, set) {
       }
       return result
     } catch (err) {
+      // If the browser's JavaScript engine doesn't include the error's
+      // "name: message" string at the top of the stack, manually push it
+      // onto the stack so the information is shown in the error log.
+      if (!err.stack.startsWith(err.toString())) {
+        err.stack = err.toString() + '\n' + err.stack
+      }
```
eriese, line 180: I'd rather have this happen inside of `log.error()` so that this check is performed in all circumstances rather than just this one case

eriese (changes requested):

packages/core/src/pattern.mjs
BenJamesBen: I removed the code from core `pattern.mjs` and moved it into the workbench `logs.js` so all errors are handled.

BenJamesBen:
I removed the code from core `pattern.mjs` and moved it into the workbench `logs.js` so all errors are handled.

eriese (approved):

PR #3307, opened 2022-12-23
fix(core): Fix Path.join() to work with noop

BenJamesBen:
This PR fixes a bug where `Path.joinI()` produces an error if any of the paths have a `noop()` operation.

 Screenshot 2022-12-23 at 2 22 59 PM


BenJamesBen:
Closing. Accidentally submitted an old branch.

PR #3308, opened 2022-12-23
fix(core): Fix Path.join() to work with noop

BenJamesBen:
This PR fixes a bug in `Path.join()` where it produces the below error if any of the paths contains a `noop()` operation.

Screenshot 2022-12-23 at 2 22 59 PM


eriese:
@BenJamesBen please add a test for this

BenJamesBen:
I added a test for joining paths with noop operations. `yarn test` runs with no errors in the `core` package directory.

eriese (approved):

PR #3310, opened 2022-12-24
fix(bruce): Turn failing patterns into degraded patterns with warning messages

BenJamesBen:
This PR is the first in a series of pattern fixes to implement the philosophy of "Patterns should not crash with an error. Instead, they should produce a degraded pattern as best they can and print warning messages to the user."

Specifically this Bruce PR:
- Stops the iterative tusk adjustment loop as soon as it detects that the adjustments have stopped producing better results. The goal is to avoid "runaway" rotations that keep making the pattern worse the longer the iterations run.
- Avoids a crash when two circles do not intersect and the bottom left location on the side cannot be determined. Instead, it picks the best-but-arbitrary location for the point and creates it there.
- For each of these situations, a warning message is logged explaining the specific issue.
- For each of these situations, a stronger, detailed warning message is logged to tell the customer that the pattern is degraded and to provide instructions about what to do:

> We were not able to generate the Side pattern piece correctly. Manual fitting and alteration of this and other pattern pieces are likely to be needed. First, please retake your measurements and generate a new pattern using the new measurements. If you still see this warning with the new pattern, then please make a test garment, check fit, and make alterations as necessary before trying to make the final garment.

(In a future PR, we might put this message text into a library as a string or method. That way, this standard wording can be used in different patterns without having to duplicate the code in multiple files. An alternate implementation might be to print a shorter message with a link to an "error page" that would contain the longer explanation text.)

### DIFF
--- a/designs/bruce/src/front.mjs
+++ b/designs/bruce/src/front.mjs
@@ -33,8 +33,11 @@ function draftBruceFront({
   macro,
   snippets,
   Snippet,
+  log,
   part,
 }) {
+  let adjustment_warning = false
+
   // Initialize
   init(part)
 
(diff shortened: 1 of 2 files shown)


BenJamesBen:
Before:

Screenshot 2022-12-23 at 7 32 23 PM

Screenshot 2022-12-23 at 7 32 35 PM

After:

Screenshot 2022-12-23 at 7 27 48 PM

Screenshot 2022-12-23 at 7 27 34 PM

joostdecock:
I am in favor of letting things graciously decay rather than explode, but we need to consider how we want to implement this.
Specifically, this implementation is untranslatable, which is a problem as we do support i18n.

The obvious fix is to include translation keys instead, but the link between these translation keys and their content somewhere in the i18n package would be tenuous at best.

I feel that it's perhaps a better idea to move translation into the design. So that a design would have a named `i18n1 export with the various translation keys per language. But that would also require changes to the i18n package, as well as how translations are handled with crowdin.

To be clear: This relates only to the warnings being raised. Not to to actual logic changes which are good.

I also feel we might need to add a `notice` log level if we are going to start making warnings the new soft errors.

BenJamesBen:
I was also thinking that perhaps this type of warning/messaging might be part of the infrastructure. Patterns would be aware of their own status, whether they are good or degraded, and perform the appropriate actions/logging. 

Also, I was thinking that there are different levels of a "degraded" pattern. Minor adjustments/alterations needed. Major adjustments/alterations needed. Totally broken (and the customer should be warned that even attempting to construct the pattern would likely be a waste of their time, instead of being told to construct a test garment.)

joostdecock:
Would it block progress for too long if we add it to the next call agenda to hash it out?

BenJamesBen:
No, this should be discussed, to avoid duplicate work since I have a number of other designs (Bella, Charlie, Titan) that I am planning to update with this sort of fix. We should take our time on this issue.

joostdecock (approved):

PR #3316, opened 2022-12-25
fix(org): Use absolute URL paths in navigation

BenJamesBen:
Because the freesewing.org documentation is in the `docs/` directory, the sidebar navigation URLs incorrectly have duplicate `docs` in their paths. For example `http://localhost:8000/docs/docs/guide/patterns`.

The navigation seems to construct the URLs to be relative. Because the current path is `docs/`, the slug/href `docs/guide/patterns` becomes the incorrect URL `docs/docs/guide/pattens`.

This PR fixes this by changing the hrefs to absolute paths.

(I do not know whether this is the best fix or how it might affect pages not under the `docs/` hierarchy. Another way to fix this might be to strip the `docs/` from the beginning of the `slug`. Or, keep the `slug` intact, create a new `navSlug` for the stripped path, and have the navigation use `navSlug` instead of `slug`, if present.)

### DIFF
--- a/sites/org/components/navigation/primary.js
+++ b/sites/org/components/navigation/primary.js
@@ -95,7 +95,7 @@ const SubLevel = ({ nodes = {}, active }) => (
             `}
             >
               <Link
-                href={`${child.__slug}`}
+                href={'/' + `${child.__slug}`}
                 title={child.__title}
                 className={`
                   grow pl-2 border-l-2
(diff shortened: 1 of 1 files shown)


joostdecock (approved):

joostdecock:
This fix is fine, but as you may have noticed, I've lifted the navigation out of the shared folder because I intend to build a new navigation menu for freesewing.org.

It's a WIP in the `joost` branch which I merge from time to time. It will be a while before freesewing.org stabilizes, so that's perhaps something to keep in mind.

joostdecock (dismissed):

joostdecock:
Superseded by https://github.com/freesewing/freesewing/commit/a728a5a2275874deb41fe004ba93a057456aa4b5

PR #3332, opened 2022-12-26
fix(docs): Insert space between numbers and SI units

BenJamesBen:
This PR inserts a space character between numbers and SI units ("8 cm" instead of "8cm", for example).

This conforms to the SI and NIST guidelines:
https://www.bipm.org/en/publications/si-brochure
https://www.nist.gov/pml/owm/writing-si-metric-system-units#spacing

I'm writing a style guide that I am following for FreeSewing, and the guideline currently is:
- There should be a space character between numbers and SI units. For example, "5 cm" and "1.3 mm", rather than "5cm" and "1.3mm". (The exception is for CSS and SVG values. Their specifications say that the space character should not be used.)


joostdecock:
Style guide is a good idea Thanks.

joostdecock (approved):

PR #3351, opened 2022-12-30
fix(docs): Change examples to actually use the documented method

BenJamesBen:
The `Path.attr()` and `Path.length()` API documentation examples didn't actually use the `Path.attr()` or `Path.lengthI()` methods. This PR fixes that.

The new `Path.length()` example:
Screenshot 2022-12-30 at 7 41 55 AM

### DIFF
--- a/markdown/dev/reference/api/path/length/en.md
+++ b/markdown/dev/reference/api/path/length/en.md
@@ -14,7 +14,7 @@ float path.length()
 
 <Example caption="Example of the Path.length() method">
 ```js
-({ Point, points, Path, paths, macro, utils, part }) => {
+({ Point, points, Path, paths, macro, utils, units, part }) => {
 
   points.A = new Point(45, 60)
   points.B = new Point(10, 30)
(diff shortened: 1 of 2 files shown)


markdown/dev/reference/api/path/length/en.md
```diff
@@ -14,36 +14,29 @@ float path.length()
...
-  macro("pd", {
-    path: new Path().move(points.B).line(points.A),
-    d: 10
-  })
-  
-  macro("pd", {
-    path: new Path().move(points.B).curve(points.BCp2, points.CCp1, points.C),
-    d: -10
-  })
+
```
joostdecock, line 32: I'm not sure why you felt the need to replace the macros, but we should refrain from hard-coding units in our examples.

The macros will automatically format the value in the units the user requested. So I suggest we keep them

joostdecock (changes requested):

markdown/dev/reference/api/path/length/en.md
BenJamesBen: I removed the macros because I didn't want to confuse the user by presenting an "alternate" way to display path lengths (since the goal of the document was to show `Path.length()` in action). However, I'll re-add the macros and do the example a different way.

BenJamesBen:
I restored macros to the example. (I had originally removed them because I felt their presence would confuse the user, since the goal was to demonstrate `Path.length()`, not the macros.) The new example combines the use of both macros and `Path.length()`.

Screenshot 2022-12-30 at 8 20 59 AM

joostdecock (approved):

joostdecock:
>  felt their presence would confuse the user, since the goal was to demonstrate Path.length(), not the macros.) 

Which was the actual title of this PR. I should have realized, but I appreciate this best-of-both-worlds solution.

Issue #3389, opened 2023-01-06
[docs]: Add a style guide for documentation


BenJamesBen:
### Are you a FreeSewing patron? 😃

No, I am not 😞

### Suggested content

Add a style guide for FreeSewing documentation to help doc writers maintain consistency.

BenJamesBen:
A draft style guide that I have been using is below. (I am still unsure about some of the preferred spellings.)

- Regarding FreeSewing:
  - The spelling for the name of the organization is "FreeSewing", not "FreeSewing.org", and the "F" and "S" are capitalized.
  - The websites are "freesewing.org" and "freesewing.dev", with all letters in lowercase.
  - "freesewing" without capitalization is used for package names and the GitHub account name.

- Preferred spellings that might disagree with dictionary spellings:
  - "lifecycle" when specifically referring to the stages that the FreeSewing technology goes through when it runs and generates a pattern.
  - "whitespace" when specifically used to describe computer whitespace characters, (while "white space" is used to describe spacing before/after words or lines).
  - "bartack" when specfically referring to the element from the FreeSewing plugin, ("bar tack" when otherwise referring to the element) (Or, is the spelling "bartack" accepted sewing jargon?)
  - "cutonfold" when specifically referring to the element from the FreeSewing plugin, ("cut-on-fold" when refering to the line in general)

- Jargon that might not be found in dictionaries but are accepted words/spelling:
  - Sewing and fashion: "womenswear", "menswear", "sleevecap", "grainline", "outseam", "underbust"
  - Technology: "truthy", "falsy", "backend", "frontend", "config", "monorepo", "serverless", "runtime", "lockfile"
  - FreeSewing-specific variations: "Howto" ("H" capitalized), "measies", "scalebox", "miniscale", "bnotch", "cutlist", "multiset"
  - Other acceptable words: - "startpoint", "endpoint" (Or, should we standardize on "start point" and "end point"?)

- Other preferred spellings:
  - "stand-alone", not "standalone".
  - "line break", not "linebreak".
  - "website", not "web site".
  - "pop-up", not "popup"

- There should be a space character between numbers and SI units. For example, "5 cm" and "1.3 mm", rather than "5cm" and "1.3mm". (The exception is for CSS and SVG values. Their specifications say that the space character should not be used.)

- Correct capitalizations for proper nouns and abbreviations:
  - SVG, JSON, YAML (all letters capitalized)
  - NextJS, ReactJS (the first letter and "JS" capitalized for these)
[...]

PR #3390, opened 2023-01-07
fix(docs): Updates and corrections for FreeSewing in different environments

BenJamesBen:
1. Added `<ReadMore>` tag to the summary page.
2. Changed wording to the `<Tip`> message, to clarify what "generating a pattern" means versus just "wanting a pattern".
3. Standardized spelling/capitalization for freesewing.org and Node.js.

### DIFF
--- a/markdown/dev/howtos/environments/browser/en.md
+++ b/markdown/dev/howtos/environments/browser/en.md
@@ -8,14 +8,15 @@ you can generate patterns in the browser with a few lines of JavaScript.
 
 <Tip>
 
-##### Use FreeSewing.org if you just want a pattern
+##### Use freesewing.org if you just want a pattern
 
 These instructions are intended for people who want to generate
-their own patterns. If you _just want a sewing pattern_ you can
-get all our designs on [FreeSewing.org](https://FreeSewing.org/),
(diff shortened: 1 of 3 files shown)


joostdecock:
We write FreeSewing with capital F and S, because it's much easier to read that way.

Just because it's a domain name doesn't change that IMHO as domain names are case insensitive

BenJamesBen:
I will make the "FreeSewing.org" and "FreeSewing.dev" change for this PR.

Later I will also make the change throughout the rest of the documentation. (Until now, I had been changing all instances to lowercase "freesewing.org" and "freesewing.dev", following the below style guidelines which I will also correct in #3389 . )

- Regarding FreeSewing:
  - The spelling for the name of the organization is "FreeSewing", not "FreeSewing.org", and the "F" and "S" are capitalized.
  - The websites are "freesewing.org" and "freesewing.dev", with all letters in lowercase.
  - "freesewing" without capitalization is used for package names and the GitHub account name.

BenJamesBen:
I made the "freesewing" -> "FreeSewing" changes.

joostdecock (approved):

PR #3391, opened 2023-01-07
fix(docs): Pattern tutorial: Change "you/your" to "we/us/our". Fixes #2982

### LINKED ISSUE #2982: [docs]: (pattern-design-tutorial) Change instances of "your" to "our".

https://freesewing.dev/tutorials/pattern-design/structure

No, I am not 😞

### Suggested content

The v3 pattern design tutorial has been updated so the old `your-first-part` file name/directory is now `our-first-part`, and `completing-your-pattern` was renamed `completing-our-pattern`. The conclusion page was also updated to link to the new `our-first-part` and `completing-our-pattern` URLs. Based on this, it appears that the intention is to use "our" instead of "your" wording in the tutorial going forward.

The tutorial currently uses an inconsistent mix of "your" and "our" wording. And, there is now a 404 link to the old `your-first-part` URL. 

This bug is that instances of "your" should be changed to "our" in filenames, links, and text throughout the pattern design tutorial for consistency. (This would also require renaming the `testing-your-pattern` directory.)

### Additional context

_No response_

BenJamesBen:
(This PR assumes that we want #2982 fixed in this way. We could fix it in the other direction, back to "you/your".)


markdown/dev/tutorials/pattern-design/adding-measurements/en.md
```diff
@@ -8,8 +8,8 @@ we are going to draft our pattern according to the measurements provided to us.
 
 Which begs the question, which measurements?
 
-It is you, as the pattern designer, who decides which measurements are used
-to draft your pattern. For our bib, the only measurement we need is the
+It is we, as the pattern designers, who decides which measurements are used
```
joostdecock, line 11: We need to drop the `s` from `decides` here

joostdecock (approved):

markdown/dev/tutorials/pattern-design/adding-measurements/en.md
BenJamesBen: Addressed in PR #3410 .

SeaZeeZee:
I think we should return the pattern tutorial to how it was previously because I think using plural pronouns in this case takes agency away from the person following the tutorial. It makes it sound as if the writer is doing this together with the reader but the reader is the one following the tutorial and executing the work. 

Maybe we should bring this up in a contributors call.

BenJamesBen:
I added the issue to the Feb. 4 Contributor Call agenda: https://github.com/freesewing/freesewing/discussions/3460

PR #3396, opened 2023-01-07
[Fixes #3273] Get stack name using current settings rather than all settings

### LINKED ISSUE #3273: [bug]: Examples stacks are no longer stacking

The Examples design previously produced a pattern that looked like a face. The individual elements of the face were individual layers, and when stacked together the layers would produce the face. However, something has changed such that the parts no longer stack and are instead spread out.

Screenshot 2022-12-21 at 10 36 56 AM

designs/examples

No, I am not 😞

### Additional context

_No response_

eriese:


BenJamesBen:
I tested the changes in the localhost lab, and the Examples design stacks "face" now appears:
Screenshot 2023-01-09 at 12 25 07 AM

joostdecock (approved):

Issue #3388, opened 2023-01-06
[bug]: <TITLE> element missing on freesewing.dev web pages


BenJamesBen:
### What seems to be the problem? 🤔

`<TITLE>` is a required element for HTML pages. However, all the pages on freesewing.dev and lab.freesewing.dev are missing this tag in the `<HEAD>` section.

This bug also will affect freesewing.org once v3 is deployed. (The bug is currently seen on the localhost version of freesewing.org in the `develop` branch.)

### Are you a FreeSewing patron? 😃

No, I am not 😞

eriese:
I agree page titles are important for a variety of reasons, but can you please help me understand the value of appending the name of the site to every page title?

BenJamesBen:
I wasn't sure about including the site name as part of the title, and i would be fine with removing it.

I included it because I noticed that it was a common practice on other major sites, like Google, Apple, Next.js, etc. I imagine that it might be for SEO-type reasons.

BenJamesBen:
Fixes for this bug were merged. Closing.

PR #3408, opened 2023-01-09
fix(dev,lab): Add <title> elements to HTML pages

BenJamesBen:
This is a partial fix for #3388 , for the dev and lab sites. (A separate PR will be filed for the org site.)

For freesewing.dev, the titles are the page titles with " - FreeSewing.dev" appended. For example: `Getting started with Gitpod - FreeSewing.dev` and `bartackFractionAlong - FreeSewing.dev`. The index page title is `Welcome to FreeSewing.dev`.

For lab.freesewing.dev, the titles for the lab are the capitalized design name or the design category text with " - FreeSewing Lab" appended. For example: `Simon - FreeSewing Lab` and `Block/Sloper Designs - FreeSewing Lab`. The index page title is `Welcome to the FreeSewing Lab`

### DIFF
--- a/sites/dev/pages/404.js
+++ b/sites/dev/pages/404.js
@@ -8,9 +8,11 @@ import PageLink from 'shared/components/page-link'
 
 const Page404 = () => {
   const app = useApp()
+  const title = '404: Page not found'
+  const fullTitle = title + ' - FreeSewing.dev'
 
   return (
-    <Page app={app} title="404: Page not found" layout={Layout}>
+    <Page app={app} title={title} layout={Layout}>
       <Head>
         <meta property="og:title" content="Page not found" key="title" />
         <meta property="og:type" content="article" key="type" />
@@ -27,6 +29,7 @@ const Page404 = () => {
         <meta property="og:url" content={`https://freesewing.dev/`} key="url" />
         <meta property="og:locale" content="en_US" key="locale" />
         <meta property="og:site_name" content="freesewing.dev" key="site" />
+        <title>{fullTitle}</title>
       </Head>
       <div className="flex flex-col gap-4 mt-16 lg:mt-32 text-center">
         <h1>404: Page not found</h1>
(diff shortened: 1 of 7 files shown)


eriese (changes requested):
To avoid code duplication, I think this should happen in the `Page` wrapper located in `components/wrappers/page.js`, which all of these pages are using and handing a title prop.  According to the [`next/head docs`](https://nextjs.org/docs/api-reference/next/head) you can add `Head` anywhere, so while it feels counterintuitive to put a `<head>` into a `div`, in this case it's the cleanest thing to do

joostdecock:
So, well spotted by @BenJamesBen that the title is missing, because it's a simple oversight. (eg `og:title` is set, but not the title tag).

And while @eriese makes a good point that we should keep things DRY so `components/wrappers/page.js` is the common place to add this, things get complicated fast.

For example, the open graph tags. On a blog post page, the image should be the image of the blog post. On a documentation page, the image should be one of the auto-generated OG images. On other pages the image should be *something else*. We also need to handle i18n in freesewing.org and lab, but not in dev. 
In general, each page of a similar *type* (like blog post of docs page, ...) has it's own page in the `pages` folder.

And I feel that in this particular instance it might make more sense to inline the abstraction rather than pass a bunch of props to the page wrapper and then do all sorts of stuff there.

So I am leaning more towards keeping things per-page, which is the way it's implemented now. Because while there will be some repetition, the added abstraction causes more mental overhead.

So I'm going to review this in that light, but I'm happy to debate further.

joostdecock (approved):

eriese:
I'm not saying all things in `Head` need to be moved to `Page`, just that the title should be. It's already where the meta tag for title is. And then it can do the adding of the site name in just one place, because I see your Ben's point that it seems to be the done thing.

joostdecock:
> I'm not saying all things in Head need to be moved to Page, just that the title should be.

Ah, I misunderstood.

joostdecock:
Merged this as-is, then moved title to the dev site page wrapper in #3424

PR #3414, opened 2023-01-09
feat(markdown): Added docs on how to trace a pattern

joostdecock:


joostdecock:
Merging this from the `joost` branch as that gets auto-deployed

PR #3415, opened 2023-01-10
(fix) Clean up prefixing issues in plugins

### LINKED ISSUE #3197: [bug]: cutonfold macro prefix is used as a suffix

The `cutonfold` macro has a `prefix` property that gets applied to the names of the generated path and points. However, the generated path is named `paths.cutonfold${prefix}`, so the prefix is unexpectedly used as a suffix. Similarly, generated points also use the `prefix` as suffix naming scheme. This discrepancy could cause confusion. 

There are tests that currently expect the suffix naming scheme, and these tests would need to be changed if the code is changed to use `prefix` as a prefix. If any Designs expect the existing suffix naming scheme, those would also break if the change is made. (However, I did a quick search, and I believe that none of the FreeSewing Designs would be affected by the change.)

plugins/plugin-cutonfold

No, I am not 😞

### Additional context

_No response_

### LINKED ISSUE #3198: [bug]: gore macro uses generic path and point names

The `gore` macro creates a path named `paths.${prefix}seam` and points with names like `points.${prefix}p1` and `points.${prefix}Cp1`. Because the `prefix` property is an empty string by default, the names are likely to end up being `paths.seam`, `points.p1`, and `points.Cp1`. 

Such generic names increase the chances of inadvertent collision and overwriting of paths and points. To make this issue less likely, we should consider making the path/point names less generic. Perhaps something like, `paths.${prefix}goreSeam` and `points.${prefix}goreP1`.

If a change is made, we will need to check existing Designs and tests to see if they depend on the old naming scheme. If so, the Designs and tests will also need to be changed.

plugins/plugin-gore

No, I am not 😞

### Additional context

_No response_

eriese:
resolves #3198, adds an additional prefix to eliminate generic point names in gore and round plugins
resolves #3197 uses prefix as a prefix instead of suffix in cutonfold macro

### DIFF
--- a/plugins/plugin-cutonfold/src/index.mjs
+++ b/plugins/plugin-cutonfold/src/index.mjs
@@ -34,7 +34,7 @@ export const plugin = {
       so = {
         offset: 15,
         margin: 5,
-        prefix: '',
+        prefix: 'cutonfold',
         ...so,
       }
       if (typeof setCutOnFold === 'function') {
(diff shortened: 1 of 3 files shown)


joostdecock (commented):
We're switching from `camelCase` to `PascalCase` here.

As camelCase is the default throughout our codebase I am not super stoked about this.

Am I missing something about why you made those changes, or do you feel **This is the way**?

eriese:
> We're switching from `camelCase` to `PascalCase` here.
> 
> As camelCase is the default throughout our codebase I am not super stoked about this.
> 
> Am I missing something about why you made those changes, or do you feel **This is the way**?

My mistake. I forgot that sometimes there is no prefix. 

I do think that there's merit to having the prefix come first if it's going to be called a prefix, and I think there's merit to having it be called a `prefix` rather than a `suffix` because that's consistent with other macros.

I think suffixes instead of prefixes in all macros would have been great, but I think it's kind of a big breaking change to switch them now. 

I'm fine to take that one out of this PR and we decide we're not changing it, I was just trying to clear out some low hanging fruit in the bugs list

joostdecock:
> I think suffixes instead of prefixes in all macros would have been great, but I think it's kind of a big breaking change to switch them now.

Technically we can break things for v3. But it's perhaps low return on investment.
The main reason I went with prefix is that it groups points together (alphabetically) that are generated by the same plugin

eriese:
> Technically we can break things for v3. But it's perhaps low return on investment. The main reason I went with prefix is that it groups points together (alphabetically) that are generated by the same plugin

a very sensible approach, the question is how invested we are in consistency.

In `plugin-mirror`,  `'mirrored'` is the default prefix
In `plugin-gore` and `plugin-round` there is currently no default prefix, and points are named without any way of associating them with the macro that created them
In `plugin-cutonfold` there is no default prefix, the prefix is a suffix, and points are always additionally prefixed with `'cutOnFold'`

Given this, I think the approach with the greatest consistency and the least pascal casing is to have `'mirrored'`, `'gore'`, `'round'` and `'cutOnFold'` as the default prefixes in their respective macros, each to be potentially replaced with a custom prefix.

How do we feel about this?

joostdecock:
> Given this, I think the approach with the greatest consistency and the least pascal casing is to have `'mirrored'`, `'gore'`, `'round'` and `'cutOnFold'` as the default prefixes in their respective macros, each to be potentially replaced with a custom prefix.
> 
> How do we feel about this?

That sounds great. Let's do it :)

PR #3420, opened 2023-01-11
(fix) fixes multiple children error in standalone

SeaZeeZee:
@eriese walked me through a fix for this standalone issue yesterday and I am submitting it now.

### DIFF
--- a/packages/new-design/shared/pages/index.mjs
+++ b/packages/new-design/shared/pages/index.mjs
@@ -45,7 +45,7 @@ const HomePage = () => {
         <h1>FreeSewing</h1>
         <h4>{translations.sade[app.locale]}</h4>
         <Link href="/design" className="btn btn-primary btn-lg h-20 my-8 mb-12">
-          <span>
+          <>
             <span role="image" className="text-4xl px-6">
               👉
             </span>
@@ -53,7 +53,7 @@ const HomePage = () => {
             <span role="image" className="text-4xl px-6">
               👈
             </span>
-          </span>
+          </>
         </Link>
         {translations.tips[app.locale]}
       </div>


packages/new-design/shared/lab/components/header.js
```diff
@@ -66,8 +66,10 @@ const Header = ({ app }) => {
                 <span className="ml-2">{t('docs')}</span>
               </a>
               <Link href="/support" role="button" className={btnClasses}>
-                <HelpIcon />
-                <span className="ml-2">{t('support')}</span>
+                <span>
```
joostdecock, line 69: This is great, thank you.

But, since we're fixing it, we should use an empty fragment here as that would fix the issue without introducing additional markup.

Like this:
```
<Link href="/support" role="button" className={btnClasses}>
  <>
    <HelpIcon />
    <span className="ml-2">{t('support')}</span>
  </>
</Link>
```

See here for more info about fragments in React: https://reactjs.org/docs/fragments.html

packages/new-design/shared/pages/index.mjs
```diff
@@ -45,12 +45,14 @@ const HomePage = () => {
         <h1>FreeSewing</h1>
         <h4>{translations.sade[app.locale]}</h4>
         <Link href="/design" className="btn btn-primary btn-lg h-20 my-8 mb-12">
-          <span role="image" className="text-4xl px-6">
-            👉
-          </span>
-          <span className="text-xl px-2">{translations.load[app.locale]}</span>
-          <span role="image" className="text-4xl px-6">
-            👈
+          <span>
```
joostdecock, line 48: Same remark here, please use `<></>` instead of `<span></span>`

joostdecock (changes requested):

packages/new-design/shared/lab/components/header.js
SeaZeeZee: Will change it!

joostdecock (approved):

joostdecock:
Congrats @SeaZeeZee nicely done 👍

PR #3426, opened 2023-01-12
[Feature] Escape inline code while parsing it for og images

### LINKED ISSUE #3419: [feature]: Find solution for code-tags in opening paragraphs

The opening paragraph is extracted from documentation pages to be used in open graph (OG) images.
A paragraph with a tag in it currently breaks this. Like this:

> I am an opening paragraph with an `<svg>` tag in it

We should find a workaround for this, rather than restrict how people can write documentation.

@BenJamesBen suggested escaping the tags with `&lt;` ( < ) and `&gt;` ( > ) or `&lsaquo;` ( ‹ ) and `&rsaquo;` ( › ), which is a great idea.
Another option might be to run the paragraph through some sort of sanitizer.

I'm creating this issue so I/we don't forget about it.

### Additional context

See comments here: https://github.com/freesewing/freesewing/commit/256242936340ac005852ec3caec89590b0812526#commitcomment-95732920

eriese:
Resolves #3419 

Also it turned out that the text of reference links was being tossed, so that's fixed here too

### DIFF
--- a/markdown/dev/reference/plugins/svgattr/en.md
+++ b/markdown/dev/reference/plugins/svgattr/en.md
@@ -3,7 +3,7 @@ title: plugin-svgattr
 ---
 
 Published as [@freesewing/plugin-svgattr][1], this plugin takes an object of
-key-value pairs and adds them as attributes of the `svg` tag
+key-value pairs and adds them as attributes of the `<svg>` tag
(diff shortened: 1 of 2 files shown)


joostdecock (approved):

joostdecock:
Nice catch for the link text 👍

PR #3489, opened 2023-01-27
DRAFT: New snippets: notches and symbols

BenJamesBen:
(This is not a PR that is intended to be merged. Instead, it is just to keep a record of how to implement these new notches and symbols.

Screenshot 2023-01-26 at 8 39 15 PM


joostdecock:
Is there a use-case for these notches that you had in mind?
If not, would it perhaps make sense to write a (dev) documentation page on how to add a new notch, showing how to do this?

BenJamesBen:
Closing. I'll add the code in a Howto add your own notches and symbols.

PR #3492, opened 2023-01-29
feat: Port React components to mjs and named exports

joostdecock:
This is an overhaul of our React code to comply with the goals on the FreeSewing roadmap, specifically:

- ESM only, no more CJS
- Only use named exports

Note that we're already using ESM, but this PR makes it explicit by using the `.mjs` extension.
Also note that some things are still going to be CJS because that's expected by the underlying dependencies (such as the tailwind config for example).

This PR is now up to the point that the freesewing.org site works. However, it's going to be a massive breaking change for all other web frontends, such as freesewing.dev, the lab and development environment. We just have to update those too.


BenJamesBen:
freesewing.org did not work for me when I tested the PR as-is. Instead, I had to make the following changes to get it to work:
In `sites/org/hooks/useApp.mjs`:
```diff
@@ -2,10 +2,10 @@ import { useState } from 'react'
 import get from 'lodash.get'
 import set from 'lodash.set'
 // Stores state in local storage
-import useLocalStorage from 'shared/hooks/useLocalStorage.js'
-import useTheme from 'shared/hooks/useTheme'
+import { useLocalStorage } from 'shared/hooks/useLocalStorage.mjs'
+import { useTheme } from 'shared/hooks/useTheme.mjs'
 // Prebuild navigation
-import prebuildNavigation from 'site/prebuild/navigation.js'
+import { prebuildNavigation } from 'shared/prebuild/navigation.mjs'
 // Translation
 import { useRouter } from 'next/router'
 import { useTranslation } from 'next-i18next'
@@ -58,17 +58,17 @@ const buildNavigation = (lang, t) => {
 
   // Set top-level order
   nav.designs.__order = 'a'
-  nav.showcase.__order = 'b'
-  nav.docs.__order = 'c'
+  //nav.showcase.__order = 'b'
+  //nav.docs.__order = 'c'
   nav.community.__order = 'd'
-  nav.blog.__order = 'e'
+  //nav.blog.__order = 'e'
   nav.account.__order = 'f'
 
   // Translation top-level strapi pages
-  nav.showcase.__title = t('showcase')
-  nav.showcase.__linktitle = t('showcase')
-  nav.blog.__title = t('blog')
-  nav.blog.__linktitle = t('blog')
+  //nav.showcase.__title = t('showcase')
+  //nav.showcase.__linktitle = t('showcase')
+  //nav.blog.__title = t('blog')
+  //nav.blog.__linktitle = t('blog')
 
   // Translation top-level strapi pages
   nav.community.__title = t('community')
```

joostdecock:
Thanks @BenJamesBen I had to abandon this PR before it was completely finished, so I'm putting on the finishing touches now, including your input.

PR #3493, opened 2023-01-29
Titan angled waist

anna-puk:
~~_Do not merge as-is! Also contains the workaround of deleting the sites/org/components/gdpr directory._~~ (fixed)

Introducing a waist angle for `titan`: center back will be raised and center front lowered. This impacts e.g. `charlie` as well, in a way that seems to make sense.


BenJamesBen:
I tested the changes in the localhost lab with Titan, Paco, and Charlie, and everything seems to work correctly.

joostdecock:
Can we get this out of draft? @BenJamesBen has given their seal of approval, so it seems like we should be ready to move on this. But there's some conflicts and then the comment at the very top asking not to merge this.

So where do we go from here?

anna-puk:
> Can we get this out of draft? @BenJamesBen has given their seal of approval, so it seems like we should be ready to move on this. But there's some conflicts and then the comment at the very top asking not to merge this.
> 
> So where do we go from here?

This branch also contains a commit that removes the sites/org/components/gdpr directory.


> Can we get this out of draft? @BenJamesBen has given their seal of approval, so it seems like we should be ready to move on this. But there's some conflicts and then the comment at the very top asking not to merge this.
> 
> So where do we go from here?

This branch also contains a commit removing the sites/org/components/gdpr directory. I just need to revert that commit and then it should be okay to merge.

anna-puk:
I added one extra option and restored the gdpr directory. If it still passes review, please merge! :)

BenJamesBen (approved):
I re-tested the updated PR, including the new `useWaistAngleFor` option, and I believe it is ready to be merged for wider testing.

There is no documentation for the new options. We'll want to address that in a separate PR.

joostdecock:
Thanks for the reminder @HaasJona 👍

PR #3510, opened 2023-02-02
(Feature) Use xes instead of diamonds for print alignment

eriese:
requested by some folks in discord, this pull request adds a x mark at 25% and 75% across the page edges to aid in alignment of pdf printouts. It replaces the diamonds that were at 50% across the page edge

Screen Shot 2023-02-01 at 9 42 41 AM


joostdecock:
This is a great add, thanks for taking care of it @eriese 🙏

PR #3515, opened 2023-02-04
(Feature) Add action to sync dependencies bumped by dependabot

eriese:
A very simple action/script to automate the frustrating chore of updating `dependencies.yaml` on each dependabot branch before merging, and prevent the updates from being clobbered if one slips through the cracks.

I've tested the action in my fork and it works when dependabot opens a new PR or rebases an existing one.

I've identified at least one case where it does not work, which is the case where dependabot only changes the `yarn.lock`, because in that case there is nothing to update in the `dependencies.yaml`. Right now I'm letting it fail because, for the pull request it failed on, dependabot had also bumped a top level dependency in the lock file and I figured a red x could warn us that there might still be something to do before the PR can be merged.

### DIFF
--- a/.github/workflows/dependabot-sync.yml
+++ b/.github/workflows/dependabot-sync.yml
@@ -21,3 +21,10 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - name: Run Sync Script
         run: node ./scripts/sync-dependencies.mjs {{github.ref_name}}
+      - name: Commit Changes
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add .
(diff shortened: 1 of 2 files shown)


eriese:
Oh, never mind about it bumping a top level dependency while bumping a nested dependency. Looks like it's currently bumping that one in all the prs it opens. So I guess it's an open question about whether we want it to fail in that case or not. For my part I'm not sure if those bumps are also getting clobbered on `yarn kickstart` and what we would do if they are

joostdecock (approved):

joostdecock:
If this works even half of the time it will be a blessing

PR #3520, opened 2023-02-05
fix(lab) Properly call Path attributes for X-ray mode.

bobgeorgethe3rd:
- Bug
When clicking on a path (hover red) in X-ray mode this error would appear and no information would appear.

image

Upon inspection found that it was not calling the attributes from the path causing the error

- Fix

Changing `bbox` to `prop.path` so the Attributes are called from the Path. After making this change the information is now presented to the user,

image

### DIFF
--- a/sites/shared/components/workbench/draft/path.mjs
+++ b/sites/shared/components/workbench/draft/path.mjs
@@ -502,7 +502,7 @@ export const pathInfo = (props) => {
               <Tr>
                 <KeyTd>Attributes</KeyTd>
                 <ValTd>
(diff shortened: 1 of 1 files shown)


joostdecock (dismissed):

joostdecock:
@bobgeorgethe3rd as [announced on Discord](https://discord.com/channels/698854858052075530/944926520282054666/1072594923926327326) there's been lots of changes. So please rebase your fork before submitting a PR so it can be merged.

joostdecock (approved):

PR #3524, opened 2023-02-07
(Chore: deps) Give Scripts its own package.json and update sync action accordingly

eriese:
Hopefully this really is the last thing that needs to happen in this dependency syncing saga


eriese:
that failing lint is on scripts that maybe don't actually get used? I changed the way they were importing `capitalize` because they were broken, but I didn't change anything else in the scripts

joostdecock (approved):

joostdecock:
They are used. But they should import `capitalize` from `sites/shared/utils.mjs` note from core.

In any case, this is nothing for you to worry about as it's not you who wrote that garbage anyway :) (it's me)

joostdecock:
Because it is importing  `capitalize` from core. But that's a v2 relic. Core doesn't use `capitalize` in v3, so it is no longer used and should/will be removed as a named export.

`sites/shared/utils.mjs` has `capitalize` so we can import it from there. It doesn't really matter as this script is rarely used and when it is typically run manually.

PR #3536, opened 2023-02-13
fix (useLocalStorage) Simplify useLocalStorage hook

eriese:
After the mjs overhaul, lint caught some code smells in `useLocalStorage`, so this is to fix those.

This will want some pretty thorough testing to make sure that state changes that are expected to persist through a refresh do so.


BenJamesBen:
I am seeing some behaviors with the `eriese-mjs-lint` localhost lab that are different compared to `develop` localhost lab.

Behavior 1, steps to reproduce:
1. In the localhost lab.
2. In the Designs dropdown in the upper left corner, select Bob. The Bob draft appears as expected.
3. In the Designs dropdown, select Lucy.
4. Observe that Lucy does not appear as expected. Instead, the page does not seem to load.

Behavior 2, steps to reproduce:
1. In the localhost lab.
2. In the Designs dropdown in the upper left corner, select Bob. The Bob draft appears as expected.
3. Switch to the ExportThing view.
4. Switch back to the DraftDesign view.
5. Observe that the Bob draft does not appear as expected.
6. In the theme picker in the upper right corner, switch to a different theme.
7. Observe that the Bob draft now appears.

eriese:
Great catches, thanks Benjamin! I'll work on those in the morning

BenJamesBen:
Behavior 3, steps to reproduce:
1. With the localhost dev website running.
2. Load any Core API page that has an Example section. For example, http://localhost:8000/reference/api/path/smurve
3. Observe that the Example's Preview tab unexpectedly shows the X-Ray view. The Preview and X-Ray tabs both show the X-Ray view.

(I am also seeing Behavior 1 with the localhost dev website when trying to switch from some pages, for example, from the Path API page to any other page.)

eriese:
I think Behavior 3 is pre-existing, but investigating it led me to a different bug, so it's fixed now and hopefully there are no more infinite loops anywhere

BenJamesBen:
I tested the updated PR, and the 3 behaviors are no longer seen.

BenJamesBen:
I think that `<Examples>` and `sites/shared/components/mdx/examples.mjs` might be no longer used in v3? If so, it might be possible for us to remove them.

joostdecock (approved):

PR #3541, opened 2023-02-14
fix (#3539) add wrap flag to iconWrapper

### LINKED ISSUE #3539: [bug]: Part reset button is huge in LayoutThing: ForPrinting

In the `develop` localhost lab, LayoutThing: ForPrinting, the reset button for parts is unexpectedly enormous.

Screenshot 2023-02-14 at 6 52 55 AM

_No response_

No, I am not 😞

### Additional context

_No response_

eriese:
allow IconWrapper component to optionally return an unwrapped version of the icon path for use within svg objects

This resolves #3539 because it means sizing is handled by the buttons component instead of the svg that the icon is wrapped in

### DIFF
--- a/sites/shared/components/icons.mjs
+++ b/sites/shared/components/icons.mjs
@@ -5,20 +5,24 @@ export const IconWrapper = ({
   stroke = 2,
   children = null,
   fill = false,
-}) => (
-  <svg
-    xmlns="http://www.w3.org/2000/svg"
-    fill={fill ? 'currentColor' : 'none'}
-    viewBox="0 0 24 24"
-    strokeWidth={stroke}
-    stroke="currentColor"
-    strokeLinecap="round"
-    strokeLinejoin="round"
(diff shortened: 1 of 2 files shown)


joostdecock:
Am I wrong in assuming that the icons used in the SVG aren't used anywhere else on the site?

And if si, can we just move them next ti the relevant components as I suggested in the comments? Because it feels like we're coupling things here for no good reason.

Then again, I'm in a murderous mood so perhaps I'm just being cranky

eriese:
I don't know about all the icons in this file, but the one that was acting
up in #3539 is the clear icon, which is used in a lot of places, like as
the clear button for pattern options and the reset button icon for the
entire print layout, in addition to being used within svg as the reset
button for parts in the print layout

This PR can definitely sit tidy until your mood is less murderous

joostdecock (approved):

joostdecock:
Sounds good 👍

Issue #3543, opened 2023-02-14
[bug]: Jaeger's options are being overwritten by Bent's sleeves


BenJamesBen:
### What seems to be the problem? 🤔

Jaeger's options are incorrectly being overwritten by Bent, even though Jaeger inherits from Bent. 

This is especially seen in the incorrect length of Jaeger's front, back, and side parts. Note how the front extends down only halfway across the pocket.
![Screenshot 2023-02-14 at 1 19 46 PM](https://user-images.githubusercontent.com/109869956/218872036-c3eb20a4-a837-4bca-8cc0-9b594a425827.png)

This is caused because Jaeger's 19% `lengthBonus` option is overwritten with Bent's sleeve 0% `lengthBonus`.

Lines 21, 94, 118, 135, 152, and 171:

# | MESSAGE
-- | --
0 | ⚪️  jaeger.backBase options priority is 4
1 | ⚪️  jaeger.frontBase options priority is 4
2 | ⚪️  jaeger.side options priority is 1
3 | ⚪️  jaeger.collarStand options priority is 2
4 | ⚪️  jaeger.collar options priority is 1
5 | ⚪️  jaeger.underCollar options priority is 1
6 | ⚪️  jaeger.pocket options priority is 2
7 | ⚪️  jaeger.pocketLining options priority is 1
8 | ⚪️  jaeger.chestPocketWelt options priority is 1
9 | ⚪️  jaeger.chestPocketBag options priority is 1
10 | ⚪️  jaeger.topSleeve options priority is 1
11 | ⚪️  jaeger.underSleeve options priority is 1
12 | ⚪️  jaeger.innerPocketBag options priority is 1
13 | ⚪️  jaeger.innerPocketWelt options priority is 1
14 | 🔵  centerBackDart option loaded from part jaeger.backBase
15 | 🔵  hipsEase option loaded from part jaeger.backBase
16 | 🔵  waistEase option loaded from part jaeger.backBase
17 | 🔵  rollLineCollarHeight option loaded from part jaeger.backBase
18 | 🔵  reduceWaistStandardFraction option loaded from part jaeger.backBase
19 | 🔵  reduceWaistDartFraction option loaded from part jaeger.backBase
20 | 🔵  reduceHipsStandardFraction option loaded from part jaeger.backBase
21 | 🔵  lengthBonus option loaded from part jaeger.backBase
22 | 🔵  brianFitSleeve option loaded from part brian.base
[...]

PR #3564, opened 2023-02-22
feature (Pattern Config Resolution) Resolve Part Configurations Progressively

eriese:
In preparation for the cutting layout, I've begun refactoring the way we resolve part configurations because right now we resolve all parts at once and trigger a complete re-initialization if any changes are made. We have adding at *runtime* but not really at *draft time* and for all the parts duplication that will be needed in cutting layout I liked the idea of being able to add a part, have it resolve in-cycle, and put itself onto the end of the draft order.

I ended up separating all config resolution into a separate class while I was at it because `Pattern` is doing too many things and the file is getting unmanageably large

The goals of this pull request:

- [x] adding a part with dependencies still 'just works'
- [x] adding a part at runtime also 'just works' without needing intimate knowledge of the `Pattern` class to or having to reinitialize
- [x] to this end, all steps of part resolution happen on a per-part basis rather than in bulk
- [x] fewer recursion cycles to resolve dependencies and draft order (i.e. doing it per-part shouldn't make things less efficient)
- [x] deduplicate internal tracking of part resolution
- [x] minimal changes to expected behavior in public methods and properties of the `Pattern` class (as evidenced by all existing tests passing. Only changing tests that refer to private methods and properties, or are brittle)
- [x] draft order acts as a queue that can be manipulated at draft time without concurrency issues

Current state of the branch should be that all tests pass, so let me know how you feel about the implementation

### DIFF
--- a/config/dependencies.yaml
+++ b/config/dependencies.yaml
@@ -62,6 +62,9 @@ core:
   dev:
     'eslint': &eslint '8.34.0'
     'nyc': '15.1.0'
+    'mocha': *mocha
+    'chai': *chai
+    'sinon': &sinon '^15.0.1'
 diana:
   peer:
     '@freesewing/brian': *freesewing
--- a/config/dependencies.yaml
+++ b/config/dependencies.yaml
@@ -53,7 +53,7 @@ charlie:
 core:
   _:
     'bezier-js': '6.1.0'
-    'bin-pack': '1.0.2'
+    'bin-pack-with-constraints': '1.0.1'
     'hooks': '0.3.2'
     'lodash.get': &_get '4.4.2'
     'lodash.set': &_set '4.3.2'
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -51,7 +51,7 @@
   "peerDependencies": {},
   "dependencies": {
     "bezier-js": "6.1.0",
-    "bin-pack": "1.0.2",
+    "bin-pack-with-constraints": "1.0.1",
     "hooks": "0.3.2",
     "lodash.get": "4.4.2",
     "lodash.set": "4.3.2",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -60,7 +60,10 @@
   },
   "devDependencies": {
     "eslint": "8.34.0",
-    "nyc": "15.1.0"
+    "nyc": "15.1.0",
+    "mocha": "10.0.0",
+    "chai": "4.2.0",
+    "sinon": "^15.0.1"
   },
   "files": [
     "dist/*",
--- a/packages/core/src/pattern-config.mjs
+++ b/packages/core/src/pattern-config.mjs
@@ -9,8 +9,6 @@ import { __addNonEnumProp } from './utils.mjs'
 export function getPluginName(plugin) {
   const toCheck = Array.isArray(plugin) ? plugin[0] : plugin
   return toCheck.name || toCheck.plugin?.name || false
-
-  return false
 }
 
 /////////////////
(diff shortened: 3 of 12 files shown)


joostdecock:
Can we do `pattern-config.mjs` and not `patternConfig.mjs`? 

While camelCase is our standard code style, for file names we stick to lowercase (and where we don't like in the hooks, we should change that).

eriese:
Happily, especially if that's your only note

eriese:
I've looked over all the designs, comparing this branch, `develop`, and dev.freesewing.org, and with one exception, all designs look as they did in dev.freesewing.org. 

This refactor fixes the following issues apparently introduced in v3:
- missing parts issues in **Jaeger** and **Simone**
- option priority issues in **Jaeger** (resolves #3543) and probably others
- all of Carlita being drafted narrow (which I suspect is also an option priority issue)

The one exception is that It does leave `bella.back` showing in **Bee**

In the current way of hiding inherited parts, the missing parts fixes and the extra parts issue are caused by the same code, which is now working to spec and passing its tests, so I think it's worth leaving the extra part in until we agree on a new parts hiding schema

eriese:
@BenJamesBen big internal changes in this PR, so it would definitely appreciate your wonderful QA. Known issues can be found in my previous comment.

Thanks in advance!

BenJamesBen:
I tested the PR changes by running localhost lab and comparing its drafted designs with the drafted designs in the `develop` localhost lab. I didn't see any issues, so I believe that it would be okay to merge the changes.

(In fact, I suspect that the PR changes are actually drafting patterns more correctly than what is currently in `develop`.)

A few unrelated observations:
- We might want to think about removing the "simone.fbaFront  options priority is 2" type of debug messaging. Those messages now come after all the other option setting/overriding messages, and the messaging now seems superfluous.
- With the changes, there seems to be less overriding of design options. The new draft order seems to be better in the order that designs are loaded.
- I note that the change in drafting order also affects the order in which measurements are presented in the Measurements view. But, the order of the measurements also previously changed during the port from v2 to v3. In v3 measurements are attached to parts and not designs, and there is no way for a design to specify an order.
- FYI, I added the below debug logging to my local copy to help me understand how part hiding was working. We might want to think about adding some of the logging if it might be helpful.

In `packages/core/src/pattern-config.mjs`:
```diff
@@ -151,5 +151,15 @@ PatternConfig.prototype.__addPart = function (depChain) {
   // Handle various hiding possibilities
-  if (part.hide || part.hideAll) this.partHide[part.name] = true
-  if (part.hideDependencies) this.__hiding.deps[part.name] = true
+  if (part.hide) {
+    this.store.log.debug('📋 ' + part.name + ' hide: true')
+    this.partHide[part.name] = true
+    this.store.log.debug('➖ Hiding ' + part.name + ' (because "' + part.name + ' hide: true")')
+  }
+  if (part.hideDependencies) {
+    this.store.log.debug('📋 ' + part.name + ' hideDependencies: true')
+    this.__hiding.deps[part.name] = true
+  }
   if (part.hideAll) {
+    this.store.log.debug('📋 ' + part.name + ' hideAll: true')
+    this.partHide[part.name] = true
+    this.store.log.debug('➖ Hiding ' + part.name + ' (because "' + part.name + ' hideAll: true")')
     this.__hiding.all[part.name] = true
@@ -469,2 +479,3 @@ PatternConfig.prototype.__setFromHide = function (part, depName) {
     this.partHide[depName] = true
+    this.store.log.debug('➖ Hiding ' + depName + ' (because "' + part.name + ' from: ' + depName + '" and there was a hideDependencies upstream.)')
     this.__hiding.deps[depName] = true
@@ -473,2 +484,3 @@ PatternConfig.prototype.__setFromHide = function (part, depName) {
     this.partHide[depName] = true
+    this.store.log.debug('➖ Hiding ' + depName + ' (because "' + part.name + ' from: ' + depName + '" and there was a hideAll upstream.)')
     this.__hiding.all[depName] = true
@@ -492,2 +504,3 @@ PatternConfig.prototype.__setAfterHide = function (part, depName) {
     this.__hiding.all[depName] = true
+    this.store.log.debug('➖ Hiding ' + depName + ' (because "' + part.name + ' after: ' + depName +'" and there was a hideAll upstream.)')
   }
```

joostdecock (approved):

PR #3565, opened 2023-02-23
fix(workbench): Change forPrinting background fill to "none". Fixes #2649

### LINKED ISSUE #2649: [bug]: Printing layout is black in light themes on Safari browser

With the Safari desktop browser on lab.freesewing.dev or the localhost lab, in the workbench LayoutThing: ForPrinting screen, the layout incorrectly appears black when a light theme (Light, LGBTQ, Trans) is used.
Screen Shot 2022-08-24 at 8 36 19 AM

The problem isn't seen with dark themes (Dark, Haxor).
Screen Shot 2022-08-24 at 8 36 34 AM

_No response_

No, I am not 😞

### Additional context

This was with Safari Version 15.6.1 (17613.3.9.1.16).

BenJamesBen:
I believe that the intention was to have the fill set to `none` and that the previous `url(#page)` was simply left over from some previous implementation. It just happened that the failover behavior for the non-existent `#page` URL was setting fill to `none`, and because of this the issue was not noticeable (except on Safari which handles the failover incorrectly).

I'll request reviews from @joostdecock and @eriese who might have knowledge about why we are currently using the `#page` URL. Perhaps its use is intentional. Or, perhaps it actually is a valid, existent URL that I failed to locate.

### DIFF
--- a/sites/shared/components/workbench/layout/print/index.mjs
+++ b/sites/shared/components/workbench/layout/print/index.mjs
@@ -35,7 +35,7 @@ export const PrintLayout = (props) => {
   } catch (err) {
     console.log(err, props.gist)
   }
-  const bgProps = { fill: 'url(#page)' }
+  const bgProps = { fill: 'none' }
 
   const exportIt = () => {
     setError(false)


joostdecock:
@eriese I suspect the background url serves a purpose and removing it would be a (visual) regression?
But it might be a holdout from earlier work rather than the work you did on this, so I'd appreciate your take on this.

eriese:
I have no idea... it's definitely a holdover from earlier work. I also suspected a visual regression, but `fill: none` has taken care of that, so I think it's the way to go

eriese (approved):

PR #3584, opened 2023-02-28
Refactor optiondocsimages script for mjs

eriese:


joostdecock:
@eriese minor nitpick, but this is the second PR in recent memory where the diff is: this file is completely removed, this file is completely new, which makes reviewing changes kinda impossible.

I supposed it's because the filename has changed. Perhaps using `git mv` in these scenarios could make the diff more meaningful 🤔

eriese:
> @eriese minor nitpick, but this is the second PR in recent memory where the diff is: this file is completely removed, this file is completely new, which makes reviewing changes kinda impossible.
> 
> I supposed it's because the filename has changed. Perhaps using `git mv` in these scenarios could make the diff more meaningful 🤔

ah, yes, apologies! I'll be more diligent about it in future

Issue #3517, opened 2023-02-04
[feature]: Merge annotation plugins into one


joostdecock:
### What is it that you would like to see happen? 🤔

This is a todo from the contributor call #3460

### Are you a FreeSewing patron? 😃

Yes, I am a tier-8 patron ❤️  💙 💜

### Additional context

Annotation plugins are things that annotate the pattern. Like grainline, cut on fold, and so on.

joostdecock:
@woutervdub volunteered for this on Discord. Assigned to him.

bobgeorgethe3rd:
closed as of #3625

PR #3625, opened 2023-03-10
plugin-annotations

woutervdub:
I think this is done. Including tests for the new plugins.
#3517


joostdecock:
Thank you @woutervdub for your work on this 🙏 

I presume you've allowed yourself to be inspired by the structure of `plugin-bundle` as this annotations plugin sort of bundles other plugins in a way.

I think we can simplify things a bit so that it's really just a simple plugin providing multiple macros/snippets.

I should have some time this weekend to help out with that. Since it's in the `annotations` if our monorepo, I can just jump in and push some more commits. So if it's ok with you I will do that, and then we'll be able to merge this.

woutervdub:
Yes, it seemed like a good way to keep them together, and independent.

I'm not married to a solution, so if you want to change things, go ahead.

eriese:
This is great work, and obviously was a huge undertaking!

I'm sorry to be late to the party, but wanted to check where opinion falls on adding `plugin-logo`  to annotations. I think for people using freesewing for their own purposes, that's unlikely to be a necessary annotation, so it may make sense to keep it separate

`plugin-cutlist`, on the other hand, might make sense to include because when present it now adds information to `plugin-title`

woutervdub:
I left `title` out of it because of all the stuff that was under discussion to hang on to it. `logo` seemed to be a natural to add, but I can see your reasoning to leave it out.

joostdecock:
I've pushed some changes to structure it as a single plugin.

As for the remarks about logo: I do think it might be better to keep the *branded* stuff out of this. 
That would be both `logo` and `scalebox`. 

I've kept them in for now, because I wanted to hear what y'all had to say about it.

joostdecock:
An alternative would be to think about how we can make the scalebox so that it's branding can be overwritten in an easy way (since `scalebox` is not a trivial plugin to implement) :thinking:

eriese:
I like the idea of allowing a non-branded scale box

woutervdub:
I keep wondering if a `scalebox` is still necessary, given that the export is taking care of that too.

joostdecock:
> I keep wondering if a scalebox is still necessary, given that the export is taking care of that too.

I suspect that for those of us who are familiar with Inkscape, downloading the SVG and arranging the pattern manually will remain the go-to option. It's nice that we will end up supporting something similar in the browser, but if you know inkscape, it's faster/simpler/better to just fire up inkscape

So the scalebox being replaced by whatever an export does is not a replacement since not all patterns go through such an export step.

woutervdub:
@joostdecock What still needs to happen to this? I sort of lost track to where you want to move this.

joostdecock:
I'll have some time tomorrow and dealing with open PRs is on my to-do list. So thanks for the heads-up, I am pretty sure this won't need much more than what's already done 👍

joostdecock:
I've self-assigned this as there's still a bit of work to be done to migrate the entire monorepo to use plugin-annotations.
However, I've made good progress. I just ran out of time.

We'll need to update the docs too. I've created #3791 for this so it does not get forgotten.

joostdecock:
I have (finally) merged this. There's a lot of changes here, the following plugins/packages have been removed:

- plugin-banner
- plugin-bartack
- plugin-buttons
- plugin-cutonfold
- plugin-dimension
- plugin-grainine
- plugin-logo
- plugin-notches
- plugin-scalebox
- plugin-title

And they are all part of plugin-annotations now. Which also includes the following extra macros:

- sewTogether
- pleat
- bannerbox

However, for people who load our plugin-bundle package, nothing will change. So these changes will be mostly transparent.

Mostly, because the following are breaking changes:

- Renamed `sewtogether` to `sewTogether` because consistency.

I also did essentially rewrite plugin-mirror (the `mirror` macro) because the implementation was different from the documentation.
I've also added code to keep track of the number of times a point is mirrored, and log a warning when points are mirrored more than once because this leads to hard to troubleshoot bugs (found out the hard way).

What remains to be seen is what we'll do with plugin-cutlist, but that's new v3 stuff and thus alpha. We have time to figure it out.

I've also opened #3791 to further improve plugins in general

PR #3632, opened 2023-03-11
Enhancement (new-design) Ask before overwriting files when running new-design

eriese:
running `npx @freeswing/new-design@next` in the same parent folder with the same design name (or tutorial) as a previous run will overwrite changes made to the initial files without warning. If you've checked all your changes into `git`, it's possible to get them back, but if not you may simply lose your work.

This PR adds a check to see whether there is already a folder by that name inside the `cwd`, then asks how to handle it, with the following options: 

**Rename**: *Choose a new name for this design*
**Overwrite**: *Overwrite the existing design*
**Re-initialize**:  *Bring in a fresh workbench, but don't overwrite existing design files (useful for updating to the latest dev environment)*

A beneficial side effect is that this adds a way to update the workbench in standalone environments


eriese:
This will work, but currently has difficulties because we need to publish a new 3.0.0 alpha version due to all the changes

joostdecock (approved):

PR #3666, opened 2023-03-19
fix(plugin-cutlist, plugin-test, workbench): Cutting layout-related fixes

BenJamesBen:
1. Corrects the updated cutlist method names to fix the pattern crashes.
> TypeError: Cannot read properties of undefined (reading 'addCut')

2. Prevents these warning messages:
> Store.get(key) on key cutlist,carlton.front,materials,fabric, which is undefined
> Store.get(key) on key cutlist,bent.underSleeve, which is undefined

3. Corrects grainline rotation in the cutting layout. The fix is to allow rotation in both directions, clockwise and counter-clockwise, instead of always rotating in one direction.
Before and after:
Screenshot 2023-03-19 at 5 47 59 AM
Screenshot 2023-03-19 at 5 48 44 AM

### DIFF
--- a/plugins/plugin-cutlist/src/index.mjs
+++ b/plugins/plugin-cutlist/src/index.mjs
@@ -4,10 +4,10 @@ export const plugin = {
   name,
   version,
   store: [
-    [['cutlist.addCut'], addCut],
-    [['cutlist.removeCut'], removeCut],
-    [['cutlist.setGrain'], setGrain],
-    [['cutlist.setCutOnFold'], setCutOnFold],
+    ['cutlist.addCut', addCut],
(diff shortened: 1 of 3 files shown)


plugins/plugin-cutlist/src/index.mjs
```diff
@@ -42,7 +42,7 @@ function addCut(store, so = {}) {
     return store
   }
   const path = ['cutlist', partName, 'materials', material]
-  const existing = store.get(path) || []
+  const existing = store.get(path, [])
```
joostdecock, line 45: Nice to see some people actually read the API docs :)

joostdecock (approved):

Issue #3670, opened 2023-03-19
[feature]:(lab): Possible cutting layout enhancements


BenJamesBen:
### What is it that you would like to see happen? 🤔

In the Cutting Layout, have some way of making it more obvious when a part is too wide for the fabric.

Possible implementation:
![Screenshot 2023-03-19 at 2 30 26 PM](https://user-images.githubusercontent.com/109869956/226212642-e1797291-d06e-42f5-a72c-dd67b8222725.png)
![Screenshot 2023-03-19 at 2 30 47 PM](https://user-images.githubusercontent.com/109869956/226212601-eea9ccf8-b9ad-45f9-8f41-84edc5be383c.png)

In `sites/shared/components/workbench/layout/plugin-layout-part.mjs`:
```diff
@@ -256,6 +256,24 @@ const basePlugin = ({
                 'style',
                 `stroke-opacity: 0; fill-opacity: ${(col + row) % 2 === 0 ? 0.03 : 0.09};`
               )
+            if (cols === 1) {
+              paths[pageName + '-fabricBoundaryArea'] = new Path()
+                .attr('id', pageName + '-fabricBoundaryArea')
+                .move(points[`${pageName}-tr`])
+                .line(points[`${pageName}-br`])
+                .line(points[`${pageName}-br`].shift(0, 1000))
+                .line(points[`${pageName}-tr`].shift(0, 1000))
+                .close()
+                .attr('class', 'fill-various')
+                .attr('style', 'stroke-opacity: 0; fill-opacity: 0.3;')
+              paths[pageName + '-fabricBoundaryText'] = new Path()
+                .move(points[`${pageName}-tr`])
+                .line(points[`${pageName}-br`])
+                .setClass('hidden')
+                .addText('(beyond fabric width)',
+                  'text-4xl fill-various')
+                .attr('data-text-id', pageName + '-fabricBoundaryText')
+            }
           } else {
             paths[pageName].attr('class', 'interfacing stroke-xs')
             // add markers and rulers
```

(I tried but was unable to add i18n support for the "beyond fabric width" text.)

### Are you a FreeSewing patron? 😃

No, I am not 😞

BenJamesBen:
Recategorizing this issue for generic enhancements to the cutting layout. We might not necessarily choose to implement any of them. 

Prototype for a ruler along the left side of the page:
[...]

PR #3672, opened 2023-03-20
(cutting layout) fixes and enhancements to cutting features

### LINKED ISSUE #3634: [bug]: Can't use 'only' in cutting layout

Choosing specific parts to display in the cutting layout will result in blank renders

_No response_

No, I am not 😞

### Additional context

_No response_

eriese:
- [x] move plugin-cutlist to plugin-annotations
- [x] include cutting layouts in pdf exports
- [x] fix grain rotation issues on parts with no cutlist configured (e.g. hugo pocketbag)
- [x] resolves #3634 handle cases with 'only' setting
- [x] #3784 
- [x] #3777
- [x] Add an indicator to the edge of the fabric to help people see when parts exceed it #3670


BenJamesBen:
In addition to having the `addCu()` method, would it make sense to be able to put all the cutting instructions in the part configuration section? I imagine something like:
```js
export const back = {
   name: 'carlton.back',
   from: bentBack,
   measurements: ['chest', 'hpsToWaistBack'],
   cutlist: [
      {},
      {material: 'lining'},
      {cut: 4, material: 'plastic', bias: true}
   ],
...
```
and then there could be an `addCutlist()` method which would add the configured cutlist.

That way the cutlist would be in the configuration along with all the other part information rather than up in the code.

eriese:
> In addition to having the `addCu()` method, would it make sense to be able to put all the cutting instructions in the part configuration section? I imagine something like:
> 
> ```js
> export const back = {
>    name: 'carlton.back',
>    from: bentBack,
>    measurements: ['chest', 'hpsToWaistBack'],
>    cutlist: [
>       {},
>       {material: 'lining'},
>       {cut: 4, material: 'plastic', bias: true}
>    ],
> ...
> ```
> 
> and then there could be an `addCutlist()` method which would add the configured cutlist.
> 
> That way the cutlist would be in the configuration along with all the other part information rather than up in the code.

This was the original way we specified doing cutlist, but for reasons I cannot now remember, we chose not to keep it this way. I think it had something to do with @joostdecock not wanting cutlist stuff to be embedded in core. Putting cutlist in the part configuration makes it look and feel like the business of core, and would probably want methods at either the `Part` or `Pattern` level to access. 

If you want to keep cutlist in an easy-to-read place, I recommend the following pattern:

```js
const cutList = [
  {},
  {material: lining},
  {cut: 4, material: 'plastic', bias: true}
]

export const back = {
  name: 'carlton.back',
  from: bentBack,
  measurements: ['chest', 'hpsToWaistBack'],
  plugins: [pluginCutlist],
  draft: function({store}) {
    cutlist.forEach(c => store.cutlist.addCut(c))
  }
  ...
```

joostdecock:
> This is beyond the scope of this PR, but we might want to consider providing guidance for extending the part configuration with user-defined properties.
> 
> Adapting from RFC 822, we could say something like:
> 
> > The prefatory string "x_" will never  be  used  in the names of core properties.  This provides user-defined properties with a protected set of names.
> 
> It would allow an `x_cutlist` property to be placed in the part configuration while making it clear that it isn't part of core.
> 
> (I think that a similar issue of the `menu` property of options also not being part of core, instead being something added for the FreeSewing front-end UI. If this is the case, we could also require the property name to be changed to `x_menu`.)

I think we should just make it clear that it's ok to overload the config, just avoid the keys used by core.
Also not that the `x-` prefix was deprecated years ago ([in HTTP headers at least](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)) because inevitably some custom property is going to evolve into a standard, and now the `x-` prefixed thing is actually no longer custom.

So, let's learn from history and just allow people to overload the config for their own use. And simply document a list of things to not use. Which is similar to what we're already doing for `store.extend()`: https://freesewing.dev/reference/api/store/extend#notes

joostdecock:
> This was the original way we specified doing cutlist, but for reasons I cannot now remember, we chose not to keep it this way. I think it had something to do with @joostdecock not wanting cutlist stuff to be embedded in core. Putting cutlist in the part configuration makes it look and feel like the business of core, and would probably want methods at either the Part or Pattern level to access.

To be clear, I think we should avoid core handling the cutlist (it's a plugin after all) but I don't mind it being part of the config (it would just be an extended config key).
I think the bigger issue is that the cutlist can only be reliably determined at run time (because in a non-trivial pattern, it will depend on the options passed by the user).

I guess we could work around that by passing in a method that takes the options as input and returns the cutlist, but then we're putting logic in the config. So honestly, I think what we landed on makes sense if you want to make sure it works everywhere and not just in the simpler designs :shrug:

eriese:
>  I think the bigger issue is that the cutlist can only be reliably determined at run time (because in a non-trivial pattern, it will depend on the options passed by the user).
> I guess we could work around that by passing in a method that takes the options as input and returns the cutlist, but then we're putting logic in the config. 

There are other places in the config that we pass a method that takes the options as input (actually, just `option.hide`, I think), but given that we don't put other annotations in the config I think it feels inconsistent to put the cutlist stuff in the config. Additionally, right now we do not pass a part's configuration to its draft method because what's in the config is the concern of the `Pattern`, not the `Part`, so accessing cutlist at draft time (which is needed for stuff like adding cutting instructions to the title) would mean big core changes for a single use case

eriese:
@BenJamesBen I added a warning about the edge of the fabric. I didn't add a ruler, but I did up the contrast between rows, and changed the label from A1, A2 etc. to 1, 2, etc. Do you feel that could be sufficient?

Screen Shot 2023-04-18 at 4 36 52 PM

BenJamesBen:
I'm getting a runtime error when switching to the Cutting Layout view?

```
 Unhandled Runtime Error
 TypeError: Cannot read properties of undefined (reading 'draftOrder')
 Source
 ../shared/components/workbench/menu/core-settings/core-setting-only.mjs (7:34) @ draftOrder
 
    5 | export const CoreSettingOnly = (props) => {
    6 | const { t } = useTranslation(['app', 'parts', 'settings'])
 >  7 | const list = props.draft.config.draftOrder
      |                                ^
    8 | const partNames = list.map((part) => ({ id: part, name: t(`parts:${part}`) }))
    9 |
   10 | const togglePart = (part) => {
```

(Yes, I think the UI is fine.)

BenJamesBen:
(I see that the writing in titles is mirrored on mirrored parts, and I assume that this is intentional so the customer knows what pieces are right side up and which are mirrored.)

There is something odd happening with Teagan's sleeve in the Cutting Layout:
Screenshot 2023-04-18 at 7 46 00 PM

eriese:
Thanks for catching all that, @BenJamesBen 

- Yes, the mirrored text is intentional. You ascribe great motives to it, but mostly the writing is flipped on flipped parts because covering all text flip and rotation cases was causing a big headache, so we moved flipping and rotating to the realm of svg transforms.
- I fixed the runtime errors and the snippets not mirroring.
- I did take the `cutonfold` off of `brian.sleevecap` to fix the teagan thing, but in general I'm thinking that because most patterns need to be updated to include their cutting instructions anyway, that'll be the time to remove inherited `cutonfold`s and `grainlines`. Historically I think a lot of designs simply hide those paths, but now they will have to be properly removed.

BenJamesBen:
- I retested and confirm that the non-mirrored snippets, runtime errors, and Teagan sleeve issues seem to be fixed.

- I believe we need to start adding cutlist instructions to some of our designs to be able to fully test inherited cutlist functionality. I especially mean the Block/Sloper designs and others that inherit from them. But, also other designs that have a hidden base part that other parts inherit from. I believe a number of parts might need to use `store.cutlist.removeCut()` to remove an inherited cutlist before setting their own.

- I note that when I make the following code change to Brian's `designs/brian/src/base.mjs`, Carlton fails with the below error. Bent also seems to experience a similar error.
```diff
@@ -1 +1 @@
-import { pluginBundle } from '@freesewing/plugin-bundle'
+import { pluginAnnotations } from '@freesewing/plugin-annotations'
@@ -40 +40 @@ export const base = {
-  plugins: [pluginBundle, bustPlugin],
+  plugins: [pluginAnnotations, bustPlugin],
```
```Could not inject part carlton.pocket into part carlton.pocketLining
TypeError: Cannot read properties of undefined (reading 'clone')
    at Path.clone (webpack-internal:///../../packages/core/src/path.mjs:195:37)
    at Part.__inject (webpack-internal:///../../packages/core/src/part.mjs:297:39)
    at Pattern.createPartForSet (webpack-internal:///../../packages/core/src/pattern.mjs:145:39)
    at Pattern.draft (webpack-internal:///../../packages/core/src/pattern.mjs:125:18)
```

eriese:
> Okay, that was my mistake. I thought that `plugin-annotations` was a full replacement for `plugin-bundle`, but it is not. I need to add `plugin-annotation` to the plugin list, not replace `plugin-bundle`.

`plugin-annotations` is part of `plugin-bundle`, so you shouldn't need to add it if `plugin-bundle` is present

eriese:
@joostdecock is this waiting on anything? 
@BenJamesBen all the functionality is confirmed, it's just missing updates to the patterns, right? 

They only impact the cutting layout, so how do we feel about that being a separate PR?

joostdecock (approved):

joostdecock:
> @joostdecock is this waiting on anything? 

Not anymore :)

PR #3693, opened 2023-03-22
Pattern design tutorial proofreading

raphaelsiz:
The example for "without destructuring" was the destructured way it's done in the tutorial, the example for "with destructuring" was the old way to pass in a props object without destructuring. I switched the two so that the "without destructuring" example is of a function without destructuring and the "with destructuring" example is of the preferred destructured function.

### DIFF
--- a/markdown/dev/howtos/ways-to-contribute/proofreading/en.md
+++ b/markdown/dev/howtos/ways-to-contribute/proofreading/en.md
@@ -2,5 +2,5 @@
 title: Proofreading
 ---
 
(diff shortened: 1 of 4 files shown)


joostdecock (approved):

joostdecock:
@raphaelsiz Thank you for this, much appreciated. You should have received an invite to join our github organisation. It's not mandatory obvious, but we try to give visibility to all our contributors :)

On that note:

@all-contributors please add @raphaelsiz for docs

PR #3754, opened 2023-04-04
fix(sites) Missing Path Name Info

bobgeorgethe3rd:
Problem:
Noticed that the Name catergory in the info box was empty for paths.

Example:
image

Solution:
Change `bbox.name` to `props.pathName`

Fixed:

image

This works with inherted paths as well.


BenJamesBen (approved):
I tested the code changes and confirm that they fix the issue.

joostdecock (approved):

joostdecock:
Nice work :+1:

PR #3755, opened 2023-04-04
fix(sites) Fix Menu Settings sliders and Seam Allowance Default

bobgeorgethe3rd:
Problem:
- Seam Allowance and Margin sliders were stepping to intervals between values.
- Seam Allowances would not reset to default after clicking reset button.

Example:
image

^ Seen above the reset button has been clicked, the slider has been updated but the seam allowances has remained as the previous value.

Solution:
- Change step values from `0.1` to `1`
- Add `props.updateGist(['sa'], dflt)` to `const reset`

### DIFF
--- a/sites/shared/components/workbench/menu/core-settings/core-setting-sa-mm.mjs
+++ b/sites/shared/components/workbench/menu/core-settings/core-setting-sa-mm.mjs
@@ -26,8 +26,6 @@ export const CoreSettingSaMm = (props) => {
   const handleChange = (evt) => {
     const newVal = parseFloat(evt.target.value)
 
-    setValue(stepVal)
-
     setValue(newVal)
     if (props.gist.saBool)
       props.setGist({
--- a/sites/shared/components/workbench/menu/core-settings/core-setting-sa-mm.mjs
+++ b/sites/shared/components/workbench/menu/core-settings/core-setting-sa-mm.mjs
@@ -5,13 +5,28 @@ import { useTranslation } from 'next-i18next'
 
 export const CoreSettingSaMm = (props) => {
   const { t } = useTranslation(['app', 'settings'])
-  const { dflt, min, max } = props
+  const { dflt, min, max, step, dfltIm, maxIm, stepIm } = props
   const val = props.gist?.[props.setting]
+
+  let dfltVal
+  let maxVal
+  let stepVal
+  if (props.gist.units == 'imperial') {
+    dfltVal = dfltIm
+    maxVal = maxIm
+    stepVal = stepIm
+  } else {
+    dfltVal = dflt
+    maxVal = max
+    stepVal = step
+  }
 
   const [value, setValue] = useState(val)
 
   const handleChange = (evt) => {
     const newVal = parseFloat(evt.target.value)
+
+    setValue(stepVal)
 
     setValue(newVal)
     if (props.gist.saBool)
@@ -23,9 +38,9 @@ export const CoreSettingSaMm = (props) => {
     else props.updateGist(['saMm'], newVal)
   }
   const reset = () => {
-    setValue(dflt)
-    props.updateGist(['saMm'], dflt)
-    props.updateGist(['sa'], dflt)
+    setValue(dfltVal)
+    props.updateGist(['saMm'], dfltVal)
+    props.updateGist(['sa'], dfltVal)
   }
 
   return (
@@ -37,24 +52,24 @@ export const CoreSettingSaMm = (props) => {
           dangerouslySetInnerHTML={{ __html: formatMm(min, props.gist.units) }}
         />
         <span
-          className={`font-bold ${val === dflt ? 'text-secondary-focus' : 'text-accent'}`}
+          className={`font-bold ${val === dfltVal ? 'text-secondary-focus' : 'text-accent'}`}
           dangerouslySetInnerHTML={{ __html: formatMm(val, props.gist.units) }}
         />
         <span
           className="opacity-50"
-          dangerouslySetInnerHTML={{ __html: formatMm(max, props.gist.units) }}
+          dangerouslySetInnerHTML={{ __html: formatMm(maxVal, props.gist.units) }}
         />
       </div>
       <input
         type="range"
-        max={max}
+        max={maxVal}
         min={min}
-        step={1}
+        step={stepVal}
         value={value}
         onChange={handleChange}
         className={`
           range range-sm mt-1
-          ${val === dflt ? 'range-secondary' : 'range-accent'}
+          ${val === dfltVal ? 'range-secondary' : 'range-accent'}
         `}
       />
       <div className="flex flex-row justify-between">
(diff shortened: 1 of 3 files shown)


BenJamesBen:
The step bug occurred because the units are actually mm (even though the UI displays the metric measurement in cm). The previous step of 0.1 was allowing 0.1 mm increments and non-whole mm values.

I am still testing the PR code changes, but I note that with this fix, it is no longer possible to specify seam allowances or margins of exactly 0.5 inch when imperial units are selected. With imperial, a step value of 1.5875 (1/16 inch) or 3.175 (1/8 inch) might instead work better. (With imperial, we might also want to think about using different dflt and max values, ones that align with eighth/sixteenth inch increments better.)

bobgeorgethe3rd:
So the solution is to change to step for metric and imperial like so

```
let step
  if (props.gist.units == 'imperial'){
  step = 0.15875
  }
  else {
	  step = 1
  }
  ```
  That's my solution seems to work pretty well though I believe we try and avoid using let in this code so I wanna check if there is a better way of doing it before update the PR.

**Comment by BenJamesBen** [comment]
I now believe that your original PR is fine to merge as is. It doesn't really "break" anything, but instead it just exposes an existing behavior with imperial units in a slightly negative way. Any improvements for imperial could be handled in a later PR.

However, I did come up with the below possible enhancement (which has a side effect that the max, default, and step values are different between metric and imperial).

In `sites/shared/components/workbench/menu/core-settings/index.mjs`:
```diff
@@ -14,5 +14,10 @@ export const settings = {
   saMm: {
-    min: 0,
-    max: 25,
-    dflt: 10,
+    minMet: 0,
+    maxMet: 25,      // 2.5 cm
+    dfltMet: 10,     // 1.0 cm
+    stepMet: 1,      // 0.1 cm (1 mm)
+    minImp: 0,
+    maxImp: 25.4,    // 1 inch
+    dfltImp: 12.7,   // 1/2 inch
+    stepImp: 1.5875, // 1/16 inch
   },
@@ -29,5 +34,10 @@ export const settings = {
   margin: {
-    min: 0,
-    max: 25,
-    dflt: 2,
+    minMet: 0,
+    maxMet: 25,      // 2.5 cm
+    dfltMet: 2,      // 0.2 cm (2 mm)
+    stepMet: 1,      // 0.1 cm (1 mm)
+    minImp: 0,
+    maxImp: 25.4,    // 1 inch
+    dfltImp: 3.175,  // 1/8 inch
+    stepImp: 1.5875, // 1/16 inch
   },
```

In `sites/shared/components/workbench/menu/core-settings/core-setting-mm.mjs`:
```diff
@@ -8 +8,12 @@ export const CoreSettingMm = (props) => {
-  const { dflt, min, max } = props
+  const { dfltMet, minMet, maxMet, stepMet, dfltImp, minImp, maxImp, stepImp } = props
+  let dflt = dfltMet
+  let min = minMet
+  let max = maxMet
+  let step = stepMet
+  if (props.gist.units === 'imperial') {
+    dflt = dfltImp
+    min = minImp
+    max = maxImp
+    step = stepImp
+  }
+
@@ -23,2 +34,2 @@ export const CoreSettingMm = (props) => {
-    setValue(props.dflt)
-    props.updateGist([props.setting], props.dflt)
+    setValue(dflt)
+    props.updateGist([props.setting], dflt)
@@ -50 +61 @@ export const CoreSettingMm = (props) => {
-        step={1}
+        step={step}
```

In `sites/shared/components/workbench/menu/core-settings/core-setting-sa-mm.mj`:
```diff
@@ -8 +8,12 @@ export const CoreSettingSaMm = (props) => {
-  const { dflt, min, max } = props
+  const { dfltMet, minMet, maxMet, stepMet, dfltImp, minImp, maxImp, stepImp } = props
+  let dflt = dfltMet
+  let min = minMet
+  let max = maxMet
+  let step = stepMet
+  if (props.gist.units === 'imperial') {
+    dflt = dfltImp
+    min = minImp
+    max = maxImp
+    step = stepImp
+  }
+
@@ -52 +63 @@ export const CoreSettingSaMm = (props) => {
-        step={1}
+        step={step}
```

BenJamesBen (dismissed):

luizfzs (approved):
Looks good!

joostdecock:
I think this PR leaves a fair bit of ground uncovered. Handling absolute values is tricky and there are lots of use-cases to cover.
In general, the things to look out for are:

- The underlying value is stored in mm, but we most present it to the user as either cm or inch
- The slider should snap to values that make sense. The steps between those snapping points should lie on a natural boundary depending on the units chosen by the user
- When a value is (pre)loaded that does not fall squarely on a snap, it should still be displayed accurately
- Users should be able to inputs values manually. When they do, we should support both unit systems

So yeah, there's things that aren't fully implemented yet. But keep in mind that v3 is in alpha.
I think portraying this as a fix is somewhat setting the wrong expectations. Implenenting all of this will be a bunch of work, and should be done in the shared `mm` input, so that it applies to all `mm` options (which while we do not recommend them should still support).

joostdecock:
Added this to the call agenda so we can work out a consensus for how to move forward with this (and potential) other changes.

bobgeorgethe3rd:
Ok so I have made changes to this to work similary to how Benjamin was suggesting with some slight tweaks.

- min does not have two values as it is 0 for both
- Rather than `dfltMet` and `defltImp` etc. it is simply `dflt` and `dfltIm`. This is because `dflt` is default for everything bar imperial.

- Uses `dfltVal`, `maxVal` and `stepVal` to alternate when imperial.

The only thing I feel could be added is that when you switch between the Units it updates the seam allowance to dflt but at the same time I feel this is not wanted behaviour as you may already have set the allowance to something else and wish to know it in imperial or metric etc.

Either way what definitely needs to go through is the addition off `props.updateGist(['sa'], dflt)` in `const reset` as this does fix a bug of the reset button not actually fixing the reset so if it we just want that I can swtich it too that.

sites/shared/components/workbench/menu/core-settings/core-setting-sa-mm.mjs
```diff
@@ -5,14 +5,29 @@ import { useTranslation } from 'next-i18next'
...
+    maxVal = max
+    stepVal = step
+  }
+
   const [value, setValue] = useState(val)
 
   const handleChange = (evt) => {
     const newVal = parseFloat(evt.target.value)
 
+    setValue(stepVal)
```
BenJamesBen, line 29: Was this added line intended? 

I tested the PR in the lab with and without the line, and I wasn't able to figure out what the line was doing. The lab seemed to work fine either way.

BenJamesBen (approved):
I think that the question about the added `setValue(stepVal)` line should be answered and possibly changed, but because the PR seems to work fine in the lab as-is, I think that the PR is okay to merge.

sites/shared/components/workbench/menu/core-settings/core-setting-sa-mm.mjs
bobgeorgethe3rd: Yea Im not really sure where that came from. I doubled check and yep it does nothing so I have removed it. Not really sure why I added that? Seems like I may have done that on auto-pilot *shrugs*

BenJamesBen (approved):

eriese:
This is a tremendous effort, but I'm going to close it because it has been superseded by the new v3 ui. 

resetting the seam allowance is already working in the new ui, and #4190 implements units-based stepping on sliders that use mm values

Issue #3367, opened 2023-01-03
[bug]: utils.curvesIntersect() can fail to find intersections for some curves


BenJamesBen:
### What seems to be the problem? 🤔

The `utils.curvesIntersect()` method sometimes returns `false` for certain intersecting curves. The issue appears as if it might be in the underlying Bezier.js library's `Bezier.intersects()` method. 

Specifically, the issue is seen with this example:

```js
  let point3 = new Point(36, 160)
  let point3Cp1 = new Point(44, 106)
  let point4Cp2 = new Point(53, 67)
  let point4 = new Point(49, 17)
  let dartPoint0 = new Point(63, 78)
  let dartPoint0Cp1 = new Point(63, 78)
  let dartPoint1Cp2 = new Point(32, 64)
  let dartPoint1 = new Point(18, 63)

  let i = utils.curvesIntersect(
    point3,
    point3Cp1,
    point4Cp2,
    point4,
    dartPoint0,
    dartPoint0Cp1,
    dartPoint1Cp2,
    dartPoint1
    )
```
![Screenshot 2023-01-03 at 6 12 31 AM](https://user-images.githubusercontent.com/109869956/210374218-ac78962e-5512-4e53-ad9b-92e5fc24d441.png)

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

Full example code that can be tested via `yarn new design`, by replacing `box.mjs` with:

```js
import { pluginBundle } from '@freesewing/plugin-bundle'

function draftBox({
  options,
  Point,
  Path,
  points,
  paths,
  Snippet,
  snippets,
  complete,
  sa,
  paperless,
  macro,
  utils,
  getId,
  log,
  part,
}) {

  let point3 = new Point(36, 160)
  let point3Cp1 = new Point(44, 106)
  let point4Cp2 = new Point(53, 67)
  let point4 = new Point(49, 17)
  let dartPoint0 = new Point(63, 78)
  let dartPoint0Cp1 = new Point(63, 78)
  let dartPoint1Cp2 = new Point(32, 64)
  let dartPoint1 = new Point(18, 63)

  // Uncomment any of these lines to get array of 1 intersection point
  // dartPoint0 = dartPoint0.shift(0,1)
  // dartPoint0Cp1 = dartPoint0Cp1.shift(0,1)

  // Uncomment any of these lines to get 1 intersection point
  // dartPoint0 = dartPoint0.shift(1,1)
[...]

PR #3761, opened 2023-04-05
fix(docs): Add warning to curve intersection methods and functions

BenJamesBen:
To address the situation that Bug #3367 will not be fixed, this PR adds warnings to the documentation for the problem methods/functions.

### DIFF
--- a/markdown/dev/reference/api/path/intersects/en.md
+++ b/markdown/dev/reference/api/path/intersects/en.md
@@ -5,6 +5,13 @@ title: Path.intersects()
 The `Path.intersects()` method returns the Point object(s) where the path
 intersects with a path you pass it.
 
+<Warning>
+This method can sometimes fail to find intersections in some curves
+due to a limitation in an underlying Bézier library.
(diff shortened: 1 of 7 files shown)


joostdecock:
Did you test this in a local dev environment @BenJamesBen ? I'm specifically wondering if the parser will treat the content of the warning as markdown given that you have not left an empty line as explained here: https://freesewing.dev/guides/markdown/frequent-mistakes#using-custom-tag-components

BenJamesBen:
Yes, I did test the changes in the `sites/dev` dev environment, but I didn't see anything unusual. And, I tested it again just now, and I also don't see anything in the console. What should I be looking for?
Screenshot 2023-04-05 at 2 35 05 PM

I'll update the PR to add the extra lines.

However, I didn't understand why that guideline existed, since I've never seen the presence/absence of the blank lines to have any effect on the results. I always assumed that the recommendation was just for style reasons, to make the text easier to read. Does the lack of blank lines cause different behaviors sometimes?

joostdecock (approved):

PR #3762, opened 2023-04-05
Senya Shirt

timorl:
I made this design and it turned out ok, so I want to share it. Unfortunately, I completely don't get how to do anything else than the code (sewing instructions? adding pictures? making it show up somewhere? the last part kinda worked, but produced a huge diff and then I did weird things with it, now I have no idea how to have it work again) or in general make a proper PR, so this is probably useless without some help, sorry. <_<" I'm very willing to put in more work to get this polished, but I would need assistance with that.

Oh, and this was originally done on some slightly older version of `develop`, I'm not sure it in any way (even after fixing the stuff mentioned above) works on the current one, sorry. >.<

There are some photos of the finished product [here](https://social.wuatek.is/notice/AULJuIM1piunWYYfjc).

As a fun aside, with the help of a friend I used Inkscape and a, uh, "manual plotter"(?) to transfer the design onto fabric to avoid having to print a lot. Some details [here](https://social.wuatek.is/notice/ATysxcLoyEwUAghcn2), more on request if anyone's interested.

Edit: oh, I almost forgot – I am technically using the `waist` measurement family incorrectly, instead of measuring the waist I measured the thinnest part of the chest. There was no measurement for that (afaik?), so I cheated.


raphaelsiz (commented):
Unable to draft part senya.bottom:  _freesewing_sandy__WEBPACK_IMPORTED_MODULE_0__.draftRingSector) is not a function
There appear to be some issues with drafting your pattern. Have you drafted it in the lab?
edit: also yes I forgot but BenJamesBen's reply reminded me, Senya also wasn't in the designs.json so I also had to add it in order for it to even show up at all on the lab

BenJamesBen:
FYI, to get the Senya design to work in the localhost lab development environment to test it, I had to:
1. Add in `config/software/designs.json`:
```diff
@@ -237,2 +237,9 @@
     },
+    "senya": {
+      "description": "A FreeSewing pattern for a shirt",
+      "code": ["timorl"],
+      "design": "timorl",
+      "difficulty": 3,
+      "tags": ["tops"]
+    },
     "shin": {
```
2. `npm run reconfigure`
3. Edit in `designs/sandy/src/index.mjs`:
```diff
@@ -4,0 +5 @@ import { waistband } from './waistband.mjs'
+import { draftRingSector } from './shared.mjs'
@@ -13 +14 @@ const Sandy = new Design({
-export { skirt, waistband, Sandy }
+export { skirt, waistband, draftRingSector, Sandy }
```

Sandy currently does not export `draftRingSector()`, so that is an issue that will need to be addressed in Senya (or possibly Sandy).

timorl:
Thanks a lot for the help! I now made the PR actually look like a PR, mostly.

> Have you drafted it in the lab?

Yup, otherwise I wouldn't have been able to sew it. :D But the standard way of making it work creates a lot of changes I don't understand (`npm run reconfigure` changes _all_ the readmes Dx) and I didn't want to make a PR with these, so I only submitted the changes in `senya/`. Now I added all the other necessary stuff, although not everything that `run reconfigure` adds – why does it even modify all these things?

> Sandy currently does not export draftRingSector(), so that is an issue that will need to be addressed in Senya (or possibly Sandy).

Imo it makes sense for such a function to be externally accessible, it's very generic. Ideally it would be in some library both Sandy and Senya use, but I don't know where that would be and I think Sandy exporting it is fine as well, right?

I removed one TODO I forgot about, and added an option because of that.

The remaining problems I don't know what to do about:

1. My abuse of the `waist` family of measurements.
2. `back` doesn't show one of the measurements I tell it to show (there's even a comment about this in the code). I tried changing the order and it seems that it's always one, but I can influence which one... This suggests I am either hitting some undocumented(?) limit somewhere or just doing something immensely silly.

BenJamesBen:
1. For the measurement, if your design truly uses a measurement that isn't one of the standard FreeSewing ones, I think you should simply create a new measurement name and use it. Using "waist" isn't good because people will use their waist measurements instead of the intended one, and their patterns will not be the correct, intended size.

2. For the dimensions bug, I found the root cause. Basically, for `from:` inherited parts, adding dimensions in the new part replaces ones in the old part. You added 1 new dimension in Back, so 1 in Front was overwritten. If you had added 5 dimension in Back, 5 would have been overwritten in Front. It is because the Front dimensions named their paths `paths[0]`, `paths[1]`, `paths[2]`, etc. Back inherited all of these Front paths, but when a dimension is added for Back, it is named `paths[0]`, replacing the existing `paths[0]`. The same problem is seen with Teagan.

(I'll file the bug for the dimensions bug.)

FYI, until the bug is fixed, a workaround for Senya might be to add the first dimension twice in Front, so the same dimension is contained in `paths[0]` and `paths[1]`. Back still overwrites `paths[0]`, but `paths[1]` remains intact. 
In: `designs/senya/src/front.mjs`
```diff
@@ -103,10 +103,15 @@ function senyaFront({
     macro('hd', {
       from: points.cfWaist,
       to: points.waist,
       y: points.waist.y + sa + 15,
     })
+    macro('hd', {
+      from: points.cfWaist,
+      to: points.waist,
+      y: points.waist.y + sa + 15,
+    })
     macro('vd', {
       from: points.waist,
       to: points.armhole,
       x: points.armhole.x + sa + 15,
     })
```

timorl:
1. I replaced references to the waist with "midriff top", feel free to suggest a better name I already agree this one is terrible. ._. Also, I think this _might_ actually be `underbust`, but I haven't used it for several reasons:
   1. The instructions suggest to use `chest` in place of `underbust` for men, which totally misses the point in this case.
   2. `underbust` is potentially a tad higher than I would like for this design?
   3. There doesn't seem to be an equivalent family of measurements, so I would have to create these anyway + some information would now be duplicated in the measurements (maybe that's already the case?).
2. Ooh, nice bug! Now it works, thanks!

raphaelsiz:
Where exactly are you wanting people to measure for this? Like a rib measurement?

timorl:
I was going for "narrowest part of your chest", so yeah. Sorry if I'm messing up some nomenclature, I am a complete amateur. ^^"

BenJamesBen:
An alternate way of doing things, instead of having midriff measurements, might be to simply use the `underbust` or `waist` measurement (whichever one makes more sense... I don't have a good grasp on this subject). 

Then, have an option to adjust the "midriff" line down from the HPS-to-bust bustline or up from the HPS-to-waist waistline. Or, have the option be a percent 0 to 80, with 0 being the waistline and 100% being the bustline (and 80% or whatever representing the underbustline). Unfortunately there is no standard HPS-to-underbust measurement.

And then, have an option to adjust the waist/underbust circumference to represent the calculated midriff circumference.

Otherwise, you will need to document and explain what you mean by the midriff measurement and how it is different than the underbust or waist. I am still unclear about how to find my midriff to take a measurement. What if the narrowest part of my chest is my underbust? What if the narrowest part of my chest is my waist? What happens if someone intends to construct Senya to wear with a tightwaist corset? Would their midriff be located below their navel?

BenJamesBen (commented):
I've finished my review, but I think we need answers to some questions to be able to proceed.

I tested the actual design in the lab, and it seems okay, with no major issues.

raphaelsiz:
I think it would be the best idea to stick to waist and chest measurements
for something like this, since some people don't have (and as of yet, can't
add unless they say they have breasts) an underbust measurement in the
system. I'm also unclear on why you need specifically the smallest part of
the chest, but in that case waist would probably be better than chest. Then
there are people like me whose waists are larger than the smallest parts of
our chest, but I'm not sure how you could accommodate for that either way

designs/senya/src/base.mjs
```diff
@@ -0,0 +1,264 @@
...
+  store.set('s3CollarMaxFront', points.hps.dy(points.cfNeck) / 2)
+  store.set('s3CollarMaxBack', points.hps.dy(points.cbNeck) / 2)
+  store.set('s3ArmholeMax', points.shoulder.dy(points.frontArmholePitch) / 4)
+  // Let's leave the actual splitting the curves for the front/back parts
+
+  // Complete pattern?
+  if (complete) {
+    points.title = new Point(points.armholePitch.x / 2, points.armholePitch.y)
+    points.logo = points.title.shift(-90, 100)
+    snippets.logo = new Snippet('logo', points.logo)
```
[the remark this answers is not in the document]
timorl: Yuup, removed the whole `if`, I don't think there was anything of value there.

designs/senya/src/base.mjs
```diff
@@ -0,0 +1,264 @@
...
+    collarFactor: 4.8,
+    bicepsEase: 0.05,
+    collarEase: 0,
+    shoulderEase: 0,
+    shoulderSlopeReduction: 0,
+    // Fit
+    chestEase: { pct: 12, min: 5, max: 25, menu: 'fit' },
+    backNeckCutout: { pct: 8, min: 4, max: 12, menu: 'fit' },
+    // Advanced
+    acrossBackFactor: { pct: 98, min: 93, max: 100, menu: 'advanced' },
```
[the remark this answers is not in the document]
timorl: Uh, I even had a comment about this being probably not useful... Anyway, removed this and all the useless things that depended on it.

designs/senya/src/back.mjs
```diff
@@ -0,0 +1,99 @@
...
+    points.waistCp2,
+    Path
+  ).hide()
+  paths.sleevecap = sleevecap(
+    points.armhole,
+    points.sleeveEnd,
+    points.sleeveCp1,
+    points.sleeveCp2,
+    Path
+  ).hide()
```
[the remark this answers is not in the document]
timorl: Right, I think this was at least partially an artifact of me not understanding how this whole thing works, removed the unnecessary computations.

designs/sandy/src/index.mjs
```diff
@@ -10,4 +11,4 @@ const Sandy = new Design({
 })
 
 // Named exports
-export { skirt, waistband, Sandy }
+export { skirt, waistband, draftRingSector, Sandy }
```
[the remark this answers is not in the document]
timorl: // I don't know the answers either, so below just guesses based on my experience as a software dev; will defer to anyone with a stronger opinion.

1. It preferably should be documented, but I have no idea where. The function itself is relatively well commented, maybe that is enough?
2. Probably not, the license lets us do much more with the code if we want to and making one function exported is absolutely tiny. As for politeness, if they are still around they'll see it anyway, and if they aren't they are extremely unlikely to care.
3. No, having it in a separate "library" (or whatever the equivalent here is) would be best, but I don't have enough understanding of neither javascript nor this codebase to be able to do that properly. ^^"
4. Copying code sounds much worse than importing it, especially since this is very generic code and it does _exactly_ the desired thing.
5. I think I tried this form of import and it didn't work (somethingsomethingoutofmodule?), but see my comment above about my lack of JS knowledge. :P

designs/senya/src/base.mjs
```diff
@@ -0,0 +1,264 @@
...
+    snippets.logo = new Snippet('logo', points.logo)
+  }
+
+  log.info(`base prepared`)
+  return part
+}
+
+export const base = {
+  name: 'senya.base',
+  hide: true,
```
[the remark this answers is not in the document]
timorl: I did the first one for `base`, the second for `front` and `bottom`, nothing for `back`.

timorl:
I think I was vaguely going for something like the [Empire waistline](https://en.wikipedia.org/wiki/Waistline_\(clothing\)), maybe a tad lower. But the discussions above made me lean towards something mostly like what @BenJamesBen describes. I would suggest a solution like:
1. Unlock the underbust measurement for breastless people – I don't think Senya will work without this (or adding a separate measurement, but that's what we are trying to avoid).
2. To calculate the HPS-to-midriff-top use the percentage thing @BenJamesBen suggests, although I would suggest going the other way, i.e. the minimum is 20% (which should mean underbust, more or less, right?), 100% is exactly waist. I suggest going this way, since it sort of makes sense to go beyond 100% here (and maybe we want to do that?) while doing this in the other direction would require negatives, which seems ugly. But I'm fine with either approach.
3. IMO we should have some interpolation strategy for the circumference rather than a separate option – I think a separate option would require too much effort from the user, I for one would have no idea what to input there. Some possible interpolation strategies (but I'm very open to suggestions):
   1. Linear. Simple, but not particularly anatomically correct as far as I can tell.
   2. Use underbust curcumference until 50% (from point 2), then linear – seems to be closer to what anatomically happens, but that is almost exclusively based on my body, so it might be _very_ wrong. :P
   3. Always use the average (perhaps with an option for weighting – is this what you were suggesting by the option @BenJamesBen ?) – the differences will usually not be large enough to matter? (I don't like this option, it seems to ignore some body shapes, but my experience is terribly limited.)

I would be happy to implement 2 & 3, no idea how to do 1 (but I suspect I should be able to figure it out). I will only get to this after we reach some consensus as to how this should work.

@raphaelsiz I'm also among the people whose waist is larger than the narrowest part of their chest, this was even the exact motivation for this separate measurement. :P

BenJamesBen:
FYI, you can run lint locally in `designs/senya` directory via `yarn lint` or `npm run lint`. (If you want to see how the `lint` script is defined, check `designs/senya/package.json`.)

joostdecock:
I'm al (as so often) late to the party, but here's some thoughts on some of the issues raised above:

- @timorl For the measurement, please go with `underbust`. I understand that you feel that perhaps the measurement should be a tad lower, but let's not pretend you have much control of how good people are going to take measurements anyway. So I think `underbust` should be good enough, and there's not a very good argument to add another measurement that is almost exactly the same but not entirely
- For the `hpsToMidriffTop` we can just use `hpsToUnderbust` which is already there
- As for making this available to people *without breasts*, we have frozen the v2 branch, and this is a v3 contribution (it already uses the v3 code). In v3 there will be no longer any distinction between bodies. So this is a non-issue.
- For exporting `draftRingSector` from Sandy: It is obviously better to keep our code DRY, but I'm concerned that exporting it as it is now will make it seem like it's a part (which it is not). I see two solutions for this:
  - Export it as `__draftRingSector` using the same `__` prefix as we do in core that signifies it's an internal method. This would avoid confusion with the part names, but it still remains counter-intuitive to export an internal method
  - Move this method into a plugin, use that plugin in both Sandy and this design (this solution has my preference). 
  
I'd be happy to discuss this in [the contributor call](https://github.com/freesewing/freesewing/discussions/3768) if y'all can make it?

timorl:
I made the design use `underbust` and `hpsToUnderbust` (although the latter doesn't actually seem to exist?).

Unfortunately I wasn't able to get my head around how plugins are supposed to work, so I hope someone else will figure it out. :/

I also wasn't able to merge develop into this branch, sorry. ._.

BenJamesBen:
- I will look into moving `draftRingSector()` into a plugin. It is something I should be able to figure out how to do, I think.
- If your branch gets messed up and you aren't able to merge the latest `develop` back into it, you could eventually just create a new branch off of latest `develop`, carefully re-add all your code changes to the new branch, and then submit a new PR. Not ideal, but sometimes it is the easiest/best path.
- `hpsToUnderbust` does not exist. Instead, there are `hpsToWaistFront`, `hpsToWaistBack`, and `waistToUnderbust` measurements, and these could be used to derive the measurement you want.
- If you wanted to, you could add an option that adjusts the circumference up or down, so users can adjust based on the size of their midriff compared to their underbust. For example, something like 90 to 110% of the underbust circumference. Or, another possibility might be to require the waist circumference measurement and then approximate the midriff circumference as some percentage between underbust and waist circumferences. (I still don't have a clear understanding of where the midriff is located, so I don't know what might be most appropriate or if underbust circumference is good enough.)

BenJamesBen:
For the `draftRingSector` issue, it is unclear when a new plugin will be available. I might file an Issue in GitHub for the enhancement request. But, it shouldn't be your responsibility to make the plugin or to have to wait for it before releasing your design.

For now, I think you should simply copy the function locally into your Senya code and revert all Sandy changes. Yes, this would mean duplicate code for the short term. But, we can have the GitHub Issue include "change Sandy and Senya to use the new plugin" as part of the enhancement. And, if the plugin becomes available before Senya is released, you could simply change your code to use it. Either way, you would not be held up waiting for it.

timorl:
Thanks for the hints about which measurements to use – I now am using `waistToUnderbust` and it seems to be working (modulo something about mermaids that breaks all designs on current develop maybe?).

Speaking of which, I undid the changes to Sandy and copied the code (I feel a bit dirty). After that I managed to merge develop, no idea why it didn't work before.

So yeah, I think this seems to be in a reasonable state now, yay!

BenJamesBen:
I filed Issue https://github.com/freesewing/freesewing/issues/3865 for the `draftRingSector()` plugin feature.

BenJamesBen:
Something odd is happening when circleRatio = 100% and seam allowance is enabled. The skirt ring is not a perfect half-circle as expected, and there incorrectly is seam allowance along the top foldline.

Screenshot 2023-04-23 at 2 42 26 AM

BenJamesBen (dismissed):
There are some minor issues that i've noted, but I otherwise think that the PR can be merged for wider testing.

There is no documentation that is part of this PR, but I think that can be added later.

timorl:
I think the 100% thing is on purpose, this is likely a result of and `if` I stole from Sandy's logic – I think it tries to ensure you'll be able to cut it out when folded, if it was actually half a circle then the seam allowance would end up out of the fabric. In particular Sandy behaves the same in this context.

I'll gladly provide documentation, but I wasn't able to figure out where it should go. ^^"

BenJamesBen:
Yes, I see now that the seam allowance circle bug is also present in Sandy. I'll file a bug against Sandy.

For an example of where design documentation files are located and how they are written and arranged, please see https://github.com/timorl/freesewing/tree/senya/markdown/org/docs/patterns/sandy

timorl:
I don't like the solution with special-casing 100% – I thought in case of Sandy the "seamless circle" option was present exactly for this usecase. Anyway, because of that I instead limited the percentage here to 90%, which should make this not happen with (almost?) any settings.

I'll write some docs, if this is not merged by when I'm done I'll add them to this PR, otherwise open a separate one. Thanks for the pointer as to where to look for them! Perhaps I'll also try documenting this somewhere in the guide for creating designs.

BenJamesBen (approved):
I tested the updated PR in the lab, and I believe it is ready to merge for wider testing.

The documentation can be added in a separate PR.

timorl:
Eh, and now it no longer even compiles... I am completely lost I'm afraid. :/

joostdecock:
> Eh, and now it no longer even compiles... I am completely lost I'm afraid. :/

Apologies for both the long wait, and for the volatility. As mentioned before, v2 was frozen sometime last year and I've been busy with the v3 release.

Today, not only is v3 out, I have also implemented my preferred solution I mentioned above, which is to put the ring sector functionality into its own plugin.

So now I'd like to revive this effort and land it in our repo and on our website.
@timorl please let me know how you'd like to proceed with this. I see two options:

- You get this working again
- I get it working for you

I'm happy to do it, and realistically that will be the faster option.
But if you'd like to carry it over the finish line yourself, that's also fine.

timorl:
Phew, seems to be working although a lot has changed so I might have done some weird things...

Also, I'm not sure if the tag set is correct.

designs/senya/src/base.mjs
```diff
@@ -0,0 +1,219 @@
...
+  }
+  return a
+}
+
+function pointAtX(x, point, Point) {
+  return new Point(x, point.y)
+}
+
+function centerBack(measurements, options, Point) {
+  var hps = new Point(0, 0)
```
joostdecock, line 27: No need for variable declarations here

```mjs
return { 
  hps: new Point(0, 0),
  neck: new Point(0, options.backNeckCutout * measurements.neck),
  waistline: new Point(
    0,
    waistlineHeight(
      measurements.waistToUnderbust,
      measurements.hpsToWaistBack,
      options.lowerWaistline
    )
  )
}
```

With fat-arrow notation, you can even forego the `return` keyword:

```mjs
const centerBack = (measurements, options, Point) => ({
  hps: new Point(0, 0),
  neck: new Point(0, options.backNeckCutout * measurements.neck),
  waistline: new Point(
    0,
    waistlineHeight(
      measurements.waistToUnderbust,
      measurements.hpsToWaistBack,
      options.lowerWaistline
    )
  )
})
```

designs/senya/src/base.mjs
```diff
@@ -0,0 +1,219 @@
...
+  let target = neckCircumverence * (1 + collarEase)
+  let delta = 0
+  let run = 0
+  do {
+    run++
+    cf = cf.shift(90, delta / 3)
+    // frontNeckCpEdge aux
+    let frontNeckCpEdge = utils.beamsIntersect(neck, neckCp2, cf, new Point(20, cf.y))
+    // cfNeckCp1 needed
+    var cfCp1 = cf.shiftFractionTowards(frontNeckCpEdge, 0.55)
```
joostdecock, line 87: No `var`s please, stick to `const` and `let.

Declare your `let`s outside the loop, then assign them in the loop:
```mjs
let cfCp1, cp2Front
do {
    run++
    // ...
    cfCp1 = cf.shiftFractionTowards(frontNeckCpEdge, 0.55)
    cp2Front = neck.shiftFractionTowards(frontNeckCpEdge, 0.65)
}
```

designs/senya/src/base.mjs
```diff
@@ -0,0 +1,219 @@
...
+    points: [points.neckCp2Front, points.cfNeckCp1, points.cfNeck, points.cbNeck, points.neckCp2],
+    clone: true,
+  })
+
+  // How much space do we have to work with here?
+  store.set('s3CollarMaxFront', points.hps.dy(points.cfNeck) / 2)
+  store.set('s3CollarMaxBack', points.hps.dy(points.cbNeck) / 2)
+  // Let's leave the actual splitting the curves for the front/back parts
+
+  log.info(`base prepared`)
```
joostdecock, line 186: Please remove this log output

designs/senya/src/base.mjs
```diff
@@ -0,0 +1,219 @@
...
+    shoulderEase: 0,
+    shoulderSlopeReduction: 0,
+    // Fit
+    chestEase: { pct: 12, min: 5, max: 25, menu: 'fit' },
+    backNeckCutout: { pct: 8, min: 4, max: 12, menu: 'fit' },
+    lowerWaistline: { pct: 10, min: 0, max: 20, menu: 'fit' },
+    // Advanced
+    armholeDepthFactor: { pct: 55, min: 50, max: 70, menu: 'advanced' },
+  },
+  plugins: [],
```
joostdecock, line 217: You can drop this empty `plugins` key

designs/senya/src/bottom.mjs
```diff
@@ -0,0 +1,158 @@
...
+  for (const [shortId, uid] of Object.entries(ids.points)) {
+    points[shortId] = points[uid].copy()
+    // Some points are rotated, we need those too
+    if (points[uid + 'Rotated']) points[shortId + 'Rotated'] = points[uid + 'Rotated'].copy()
+  }
+
+  points.gridAnchor = points.in2Flipped.clone()
+
+  // Complete pattern?
+  if (complete) {
```
joostdecock, line 59: No need for this if loop

designs/senya/src/bottom.mjs
```diff
@@ -0,0 +1,158 @@
...
+        .join(paths.saBase)
+        .line(paths.hemBase.start())
+        .join(paths.hemBase)
+        .line(points.ex2Flipped)
+        .attr('class', 'fabric sa')
+    }
+  }
+
+  // Paperless?
+  if (paperless) {
```
joostdecock, line 104: No need for this if loop

designs/senya/src/bottom.mjs
```diff
@@ -0,0 +1,158 @@
...
+    macro('vd', {
```
joostdecock, line 105: All macros need to get an `id`. If not, you will only be able to use the same macro once per part. So give all these dimensions an id, as you've already done for the ones you want to be able to remove later.

(as it happens, making sure they can be removed is why we require an id now)

designs/senya/src/bottom.mjs
```diff
@@ -0,0 +1,158 @@
...
+      macro('vd', {
+        from: points.ex1Rotated,
+        to: points.center,
+        x:
+          options.circleRatio > 0.5 ? points.in1Rotated.x - sa - 30 : points.ex1Rotated.x + sa + 30,
+      })
+    }
+    snippets.center = new Snippet('bnotch', points.center)
+  }
+  log.info('bottom done!')
```
joostdecock, line 142: please remove this log output

designs/senya/src/front.mjs
```diff
@@ -0,0 +1,162 @@
...
+    .move(points.cfWaistline)
+    .join(paths.bottomAndSide)
+    .join(paths.sleevecap)
+    .join(paths.shoulderAndNeck)
+    .line(points.cfWaistline)
+    .close()
+    .setClass('fabric')
+
+  // Complete pattern?
+  if (complete) {
```
joostdecock, line 78: No need for this if loop

designs/senya/src/front.mjs
```diff
@@ -0,0 +1,162 @@
...
+      from: points.cfNeck,
+      to: points.cfWaistline,
+      grainline: true,
+    })
+    points.title = new Point(points.waistline.x / 2, points.armhole.y)
+    macro('title', { at: points.title, nr: 1, title: 'front' })
+    points.logo = points.title.shift(-90, 75)
+    snippets.logo = new Snippet('logo', points.logo)
+
+    if (sa) {
```
joostdecock, line 89: Keep sa out of a complete check

designs/senya/src/front.mjs
```diff
@@ -0,0 +1,162 @@
...
+        .join(paths.bottomAndSide.offset(sa))
+        .join(paths.sleevecap.offset(sa * 3))
+        .join(paths.shoulderAndNeck.offset(sa))
+        .line(points.cfNeck)
+        .setClass('fabric sa')
+    }
+  }
+
+  // Paperless?
+  if (paperless) {
```
joostdecock, line 101: No need for this if loop

designs/senya/src/front.mjs
```diff
@@ -0,0 +1,162 @@
...
+    // These dimensions will be inherited by the back part
+    macro('hd', {
```
joostdecock, line 103: Give these macros a unique ID

designs/senya/src/front.mjs
```diff
@@ -0,0 +1,162 @@
...
+      y: points.neck.y - sa - 45,
+    })
+    macro('vd', {
+      id: 'frontOnly',
+      from: points.cfWaistline,
+      to: points.cfNeck,
+      x: points.cfWaistline.x - sa - 15,
+    })
+  }
+  log.info('front done!')
```
joostdecock, line 145: Remove this log output please

designs/senya/src/index.mjs
```diff
@@ -0,0 +1,15 @@
+import { Design } from '@freesewing/core'
+import { data } from '../data.mjs'
+// Parts
+import { front } from './front.mjs'
+import { back } from './back.mjs'
+import { bottom } from './bottom.mjs'
+
```
joostdecock, line 7: You'll need to import and re-export the translations as `i18n`

markdown/org/docs/designs/senya/en.md
```diff
@@ -0,0 +1,5 @@
+---
+title: "Senya shirt"
+---
+
+<PatternDocs pattern='senya' />
```
joostdecock, line 5: This should be:
```
<DesignDocs design="senya" />
```

markdown/org/docs/designs/senya/instructions/en.md
```diff
@@ -0,0 +1,35 @@
+---
+title: "Sandy circle skirt: Sewing Instructions"
```
joostdecock, line 2: You forgot to update this title :)

markdown/org/docs/designs/senya/measurements/en.md
```diff
@@ -0,0 +1,5 @@
+---
+title: "Senya shirt: Required Measurements"
+---
+
+<PatternMeasurements pattern='senya' />
```
joostdecock, line 5: This should be:

```
<DesignMeasurements design="senya" />
```

joostdecock (changes requested):
Few small remarks about some of the more recent changes.
Please also provide the (English) translations. Check an existing design (like aaron) for example. (see the i18n folder)

designs/senya/src/back.mjs
```diff
@@ -0,0 +1,85 @@
...
+    .move(points.cfWaistline)
+    .join(paths.bottomAndSide)
+    .join(paths.sleevecap)
+    .join(paths.shoulderAndNeck)
+    .line(points.cfWaistline)
+    .close()
+    .setClass('fabric')
+
+  // Complete pattern?
+  if (complete) {
```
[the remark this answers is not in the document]
timorl: I completely removed these `if`s, as I think I don't actually need it for anything beyond the things you mention here. If this is correct, then I'm very glad this is the case. :3

designs/senya/src/back.mjs
```diff
@@ -0,0 +1,85 @@
...
+      from: points.cfNeck,
+      to: points.cfWaistline,
+      grainline: true,
+    })
+
+    macro('title', { at: points.title, nr: 2, title: 'back' })
+    points.scaleboxAnchor = points.scalebox = points.title.shift(90, 100)
+    macro('scalebox', { at: points.scalebox })
+
+    if (sa) {
```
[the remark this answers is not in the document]
timorl: Should be irrelevant with the `complete` check removed, but noted for the future.

designs/senya/src/back.mjs
```diff
@@ -0,0 +1,85 @@
...
+        .join(paths.bottomAndSide.offset(sa))
+        .join(paths.sleevecap.offset(sa * 3))
+        .join(paths.shoulderAndNeck.offset(sa))
+        .line(points.cbNeck)
+        .setClass('fabric sa')
+    }
+  }
+
+  // Paperless?
+  if (paperless) {
```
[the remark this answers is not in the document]
timorl: Removed everywhere then.

designs/senya/src/back.mjs
```diff
@@ -0,0 +1,85 @@
...
+    // Remove dimensions that are front only
+    macro('rmd', { ids: ['frontOnly'] })
```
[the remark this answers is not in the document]
timorl: Changed as suggested.

designs/senya/src/back.mjs
```diff
@@ -0,0 +1,85 @@
...
+    macro('vd', {
+      id: 'backWaistline',
+      from: points.cbWaistline,
+      to: points.cbNeck,
+      x: points.cbWaistline.x - sa - 15,
+    })
+  }
+  log.info('back done!')
```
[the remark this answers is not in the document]
timorl: :crying_cat_face: Done.

designs/senya/src/base.mjs
timorl: Went with the former, I prefer my functions being functions not constant lambdas (at least semantically, whether there is any difference under the hood I have no idea :P).

designs/senya/src/base.mjs
timorl: Hah, interesting style, definitely makes sense for readability, changed.

designs/senya/src/base.mjs
timorl: Done.

designs/senya/src/base.mjs
timorl: Dropped.

designs/senya/src/bottom.mjs
timorl: Removed.

designs/senya/src/bottom.mjs
timorl: Removed.

designs/senya/src/bottom.mjs
timorl: Huh, cool, now all of them show up, previously (before v3 I guess) it was just the first n for some n.

In bad news while doing this I realized that when the circle ratio differs from 50% then the printed distances seem... useless? Unless it's just me being unable to understand how I am supposed to be using them. I checked with Sandy, but it does the exact same thing, so maybe I'm just wrong.

designs/senya/src/bottom.mjs
timorl: Done.

designs/senya/src/front.mjs
timorl: Removed.

designs/senya/src/front.mjs
timorl: Irrelevant as complete removed.

designs/senya/src/front.mjs
timorl: Removed.

designs/senya/src/front.mjs
timorl: Added.

designs/senya/src/front.mjs
timorl: Removed.

designs/senya/src/index.mjs
timorl: Done, although I think I must've missed something, because the translations don't show up in the lab. :/

markdown/org/docs/designs/senya/en.md
timorl: Changed.

markdown/org/docs/designs/senya/instructions/en.md
timorl: Whoops, removed any evidence of my misdeeds.

markdown/org/docs/designs/senya/measurements/en.md
timorl: Changed.

designs/senya/src/bottom.mjs
```diff
@@ -0,0 +1,149 @@
...
+  })
+  points.logo = points.in2FlippedRotated.shiftFractionTowards(points.ex2FlippedRotated, -0.1)
+  snippets.logo = new Snippet('logo', points.logo)
+
+  points.title = points.in2FlippedRotated.shiftFractionTowards(points.ex2FlippedRotated, -0.5)
+  macro('title', { at: points.title, nr: 3, title: 'bottom' })
+
+  macro('sprinkle', {
+    snippet: 'notch',
+    on: ['in1Rotated', 'gridAnchor'],
```
BenJamesBen, line 66: Is `in1Rotated` where you intended the notch to be? 
Screenshot 2024-01-19 at 12 02 11 PM

designs/senya/src/bottom.mjs
BenJamesBen: I think there might be a bug with Sandy's dimensions for non-50% circles.

.prettierignore
```diff
@@ -16,6 +16,7 @@
 *.env
 *.sh
 *.json
+*.local
```
BenJamesBen, line 19: Was this file accidentally changed and added to this PR? Or, is it something that is needed because `sites/sde/env.local` was added to the repo?

designs/senya/src/front.mjs
```diff
@@ -0,0 +1,159 @@
...
+    points[`cf${key}`] = points[`cb${key}`].clone()
+  }
+
+  // Neckline
+  points.cfNeck = new Point(0, options.necklineDepth * measurements.waistToUnderbust)
+  points.cfNeckCp1 = points.cfNeck.shift(0, points.neck.x * options.necklineBend * 2)
+  points.neck = points.hps.shiftFractionTowards(points.shoulder, options.necklineWidth)
+  points.neckCp2 = points.neck
+    .shiftTowards(points.shoulder, points.neck.dy(points.cfNeck) * (0.2 + options.necklineBend))
+    .rotate(-90, points.neck)
```
BenJamesBen, line 30: There is something about the way `points.neckCp2` is calculated that causes the neckline curve seam allowance to be misshapen at times.
Screenshot 2024-01-20 at 6 09 43 AM
Screenshot 2024-01-20 at 6 11 07 AM

BenJamesBen (changes requested):

PR #3767, opened 2023-04-06
feat(core): Work on supported part-level transforms in layouting

joostdecock:
This is some initial work to support part-level (SVG) transforms when layouting the pattern.

It updates the method that calculates the bounding box to calculate a bounding box with transforms applied.
It also adds two new methods to utils to help with that, methods that could be useful for other people trying to do fancy transform stuff.


eriese:
Into this. I think for ease of use from a drafting perspective, I'd love to be able to use our convention of `{flipX, flipY, rotate, translateX, translateY}`, and have all this happen under the hood, if that makes sense

eriese:
having tested this out, I made some changes:
- moved the use of part level transforms to stack bounds calculations. This was needed because when we were directly mutating the bounds of the part without applying the transforms to everything, `withinPartBounds` in `sites/shared/components/workbench/draft/utils.mjs` was returning false for a lot of points that were actually in bounds. Given that the goal was just to figure out the bounds for packing purposes, I felt it was less computationally expensive to transform the bounds while calculating the stack bounds
- Check all four corners for bounds if a transform is applied. Once a part is rotating, top left and bottom right will no longer be the outer extremes in many cases.
- Join the combined matrix with a comma, because the parsing regex expects a comma

eriese (approved):

PR #3769, opened 2023-04-06
fix(sites) Counter Option Bug

bobgeorgethe3rd:
When using the counter option I found some weird behaviour when the option is being used in the below example

Example Given:
```
options: {
pleatNumber: { count: 3, min: 1, max: 10, menu: 'style' },
}

for (let i = 0; i < options.pleatNumber + 1; i++) {
points['pleatFromTopS' + i] = paths.straightCurve.shiftAlong(
          pleatKeep + (pleatKeep + pleatLengthStraight) * i
        )
}
```

When the counter slider is updated the code breaks and options.pleatNumber is not registered. This does not occur when the options is not in a for () bracket and functions correctly when this is not the case. 

The solution is to simply times the option by 1

```
options: {
pleatNumber: { count: 3, min: 1, max: 10, menu: 'style' },
}

for (let i = 0; i < options.pleatNumber  * 1 + 1; i++) {
points['pleatFromTopS' + i] = paths.straightCurve.shiftAlong(
          pleatKeep + (pleatKeep + pleatLengthStraight) * i
        )
}
```

However this is not instinctual for the user.  So isntead the change can be made on the site level.

in `sites\shared\components\workbench\inputs\design-option-count.mjs`

Goes from:
```
const handleChange = (evt) => {
    const newVal = evt.target.value
    setValue(newVal)
    props.updateGist(['options', props.option], newVal)
  }

```

Becomes:
```
const handleChange = (evt) => {
    const newVal = evt.target.value
    setValue(newVal)
    props.updateGist(['options', props.option], newVal * 1)
  }

```

This change fixes the bug without affecting the rest of the functionality of the option. It is also important to note that when clicked to (not reset) the default does not display display correctly or update correctly when this bug occurs. I am still preplexed why this bug occurs and how it breaks in such a manner but the solution was simple and easy to add.

### DIFF
--- a/sites/shared/components/workbench/inputs/design-option-count.mjs
+++ b/sites/shared/components/workbench/inputs/design-option-count.mjs
@@ -37,7 +37,7 @@ export const DesignOptionCount = (props) => {
   const handleChange = (evt) => {
     const newVal = evt.target.value
     setValue(newVal)
-    props.updateGist(['options', props.option], newVal)
+    props.updateGist(['options', props.option], newVal * 1)
   }
   const reset = () => {
     setValue(count)
--- a/sites/shared/components/workbench/inputs/design-option-count.mjs
+++ b/sites/shared/components/workbench/inputs/design-option-count.mjs
@@ -37,7 +37,7 @@ export const DesignOptionCount = (props) => {
   const handleChange = (evt) => {
     const newVal = evt.target.value
     setValue(newVal)
-    props.updateGist(['options', props.option], newVal * 1)
+    props.updateGist(['options', props.option], parseInt(newVal))
   }
   const reset = () => {
     setValue(count)


BenJamesBen:
I am not sure what is causing this behavior, but it is related to the `+ 1` part of `options.pleatNumber + 1`.

The issue also occurs with:
```js
let number = options.pleatNumber + 1
     for (let i = 0; i < number; i++) {
       log.warning('testing: ' + i)
       points['testing' + i] = new Point(-60, i * 7)
         .addText("Point " + i)
     }
```

It does not occur with:
```js
     let number = options.pleatNumber 
     number++
     for (let i = 0; i < number; i++) {
```

It also does not occur with:
```js
     for (let i = 0; i < options.pleatNumber; i++) {
```

And, for an example equivalent to `i < options.pleatNumber + 1`, the issue does not occur with:
```js
    for (let i = 0; i <= options.pleatNumber; i++) {
```

BenJamesBen:
I still don't know the root cause of the issue, but i see that the immediate problem is that `typeof options.pleatNumber === 'string'` when + 1 is added to it. If nothing is added to it then `typeof options.pleatnumber === 'number'`.

If the fix is needed, I think a better fix might be:
```diff
-    props.updateGist(['options', props.option], newVal)
+    props.updateGist(['options', props.option], parseInt(newVal))
```

joostdecock:
If you expect an option to be a number, you should cast it to a number.
Like such:

```js
for (let i=0; i <= parseInt(options.pleatNumber); i++) {
  // Do stuff here
}
```
String concatenation is done with the `+` sign in JS. So `"foo" + "bar"` = `"foobar"`
Anytime you use `+` you should be careful because if anything is a string, all of it will be treated as a string.

Since you're getting options from HTML forms, you can be certain it's a string, because they only support strings.

Last but not least, you don't need to add one to the for look. Just check that it's smaller or equal rather than only check that it's smaller.

Closing.

PR #3786, opened 2023-04-09
DRAFT: fix(workbench): Improve paperless imperial grid appearance

BenJamesBen:
Currently the imperial grid has large dashed lines at every 1 inch and 1/2 inch and smaller dashed lines at every 1/4 and 1/8 inch. I find it a bit busy and difficult to view. 

For the same 1 inch or 2.5 cm length, the imperial grid has 9 lines while the metric grid has only 6 lines. So, the imperial grid can tends to be more visually dense unless it is lightened.

To try to improve the imperial grid's appearance, this PR:
- Changes the 1 inch lines in the "lg" class to be solid.
- Moves the 1/2 inch lines to a new "md" class, with large dashes.
- Keeps the 1/4 inch lines in the "sm" class, with small dashes.
- Moved the 1/8th inch lines to a new "xs" class, with light dots.
- Added comments to make the code easier to read. 

Screenshot 2023-04-09 at 7 04 24 AM
Screenshot 2023-04-09 at 6 57 23 AM
Screenshot 2023-04-09 at 7 04 59 AM
Screenshot 2023-04-09 at 6 58 12 AM

There was also a small change to metric to add a missing "lg" class to one of the lines, but it did not cause any actual visual changes. This PR does not change the metric grid.

Screenshot 2023-04-09 at 7 04 34 AM
Screenshot 2023-04-09 at 6 57 40 AM
Screenshot 2023-04-09 at 7 04 49 AM
Screenshot 2023-04-09 at 6 58 00 AM

### DIFF
--- a/sites/shared/components/workbench/draft/defs.mjs
+++ b/sites/shared/components/workbench/draft/defs.mjs
@@ -2,28 +2,47 @@ const style = ` style="fill: none; stroke: currentColor;" `
 const grids = {
   imperial: `
       <pattern id="grid" height="25.4" width="25.4" patternUnits="userSpaceOnUse" key="grid">
-        <path class="gridline lg imperial" d="M 0 0 L 0 25.4 L 25.4 25.4" ${style} />
+        <!-- lg: solid lines every 1 inch -->
+        <path class="gridline lg imperial"
+          d="M 0 0 L 0 25.4 L 25.4 25.4"
+          ${style}
+        />
+        <!-- md: dashed lines every 1/2 inch -->
         <path
-          class="gridline lg imperial"
+          class="gridline md imperial"
           d="M 12.7 0 L 12.7 25.4 M 0 12.7 L 25.4 12.7"
           ${style}
         />
+        <!-- sm: small dashed lines every 1/4 inch -->
         <path
           class="gridline sm imperial"
-          d="M 3.175 0 L 3.175 25.4 M 6.32 0 L 6.35 25.4 M 9.525 0 L 9.525 25.4 M 15.875 0 L 15.875 25.4 M 19.05 0 L 19.05 25.4 M 22.225 0 L 22.225 25.4"
+          d="M 6.35 0 L 6.35 25.4 M 19.05 0 L 19.05 25.4"
           ${style}
         />
         <path
           class="gridline sm imperial"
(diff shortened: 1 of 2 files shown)


sites/shared/components/workbench/draft/defs.mjs
```diff
@@ -2,28 +2,47 @@ const style = ` style="fill: none; stroke: currentColor;" `
 const grids = {
   imperial: `
       <pattern id="grid" height="25.4" width="25.4" patternUnits="userSpaceOnUse" key="grid">
-        <path class="gridline lg imperial" d="M 0 0 L 0 25.4 L 25.4 25.4" ${style} />
+        <!-- lg: solid lines every 1 inch -->
```
joostdecock, line 5: I appreciate you want to use comments to make it more clear what's going on, but this is a not an efficient way to do so.
These comments would become part of the SVG defs section, and this would mean that they'd be included in the result of every render call,or each time we generate an SVG or PDF.

Please either include real code comments so they can be optimized out by the bundler, or remove them entirely.

sites/shared/components/workbench/draft/defs.mjs
```diff
@@ -2,28 +2,47 @@ const style = ` style="fill: none; stroke: currentColor;" `
...
+        <path class="gridline lg imperial"
+          d="M 0 0 L 0 25.4 L 25.4 25.4"
+          ${style}
+        />
+        <!-- md: dashed lines every 1/2 inch -->
         <path
-          class="gridline lg imperial"
+          class="gridline md imperial"
```
joostdecock, line 12: There's should be no `md` class. The absence of a size modifier (so just `gridline`) is what makes is the middle size

sites/shared/styles/svg-freesewing-draft.css
```diff
@@ -297,11 +297,16 @@ path.gridline.metric.sm {
 path.gridline.metric.xs {
   stroke-dasharray: 1 1;
 }
-path.gridline.imperial {
-  stroke-dasharray: 5 5;
+path.gridline.imperial.md {
+  stroke-width: 0.15 !important;
```
joostdecock, line 301: Everytine you have to resort to use `!important` in CSS is typically a good time to pause and refect on how to write better selectors

joostdecock (changes requested):
Please be mindful of the different environments in which styling gets applied. This PR currently only deals with on-screen output. To style the generated SVG/PDFs you'll also need to update the theme plugin.

If you check the theme plugin, you'll notice that the following class names are used:

- `gridline-lg`
- `gridline`
- `gridline-sm`
- `gridline-xs`

So I would refrain from adding `gridline-md` as it's not clear where it would fit in that hierarchy (it should be `gridline`).

Issue #3004, opened 2022-10-24
[bug]: (yuri) v3 options are different than v2


BenJamesBen:
### What seems to be the problem? 🤔

The v3 localhost lab `develop` Yuri seems to be missing options (like Fit-> cuffEase), compared to the v2 lab.freesewing.dev version.

### Design / Plugin / Package 🧐

designs/yuri

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

Possibly this code change in both `back.mjs` and `front.mjs` fixes the issue. However, we should first check to see what the intent was for why the code was written the way it was.
```diff
@@ -107,20 +107,15 @@ function yuriFront({
 
 export const front = {
   name: 'yuri.front',
-  from: {
-    ...brianFront,
-    options: {
-      ...brianFront.options,
-      // Overrides
-      collarEase: { pct: 20, min: 10, max: 30 },
-      cuffEase: { pct: 30, min: 20, max: 60 },
-      lengthBonus: { pct: 10, min: 5, max: 15 },
-      sleeveLengthBonus: { pct: 1, min: 0, max: 10 },
-    },
-  },
+  from: brianFront,
   hideDependencies: true,
   options: {
-    hipsEase: { pct: 0, min: 0, max: 10 },
+    hipsEase: { pct: 0, min: 0, max: 10, menu: 'fit' },
+    // Overrides
+    collarEase: { pct: 20, min: 10, max: 30, menu: 'fit' },
+    cuffEase: { pct: 30, min: 20, max: 60, menu: 'fit' },
+    lengthBonus: { pct: 10, min: 5, max: 15, menu: 'fit' },
+    sleeveLengthBonus: { pct: 1, min: 0, max: 10, menu: 'fit' },
   },
   measurements: ['hips', 'hpsToBust'],
   draft: yuriFront,
```

PR #3801, opened 2023-04-11
Update Yuri's option config

bobgeorgethe3rd:
This should solve #3004 

Some of Yuri's options were lost from v2 to v3. This PR should restore them.
I placed the options on the back as brianBack will draft before brianFront so it makes logical sense to put it on the yuri.back

Changes
- Separated from and options in `back.mjs`
- Separated from and options in `front.mjs`
- Added menus to override options in `back,mjs`

Deletions
- Removed duplicated override options from `front.mjs`

### DIFF
--- a/designs/yuri/src/back.mjs
+++ b/designs/yuri/src/back.mjs
@@ -99,21 +99,17 @@ function yuriBack({
 
 export const back = {
   name: 'yuri.back',
-  from: {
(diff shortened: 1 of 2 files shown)


joostdecock:
This is not a fix for #3004. You're adding an option to brian that does not exist.

Issue #3005, opened 2022-10-24
[bug]: (wahid) v3 options are different than v2


BenJamesBen:
### What seems to be the problem? 🤔

The v3 localhost lab `develop` Wahid has different options compared to the v2 lab.freesewing.dev version. For example, v3 has 13 options in the "Fit" menu while v2 has only 9 options. One of the extra options in v3 is `cuffEase` which was inherited from `brian.base`. 

It seems as if the intention was to have these extra options omitted (since things like `cuffEase` don't apply to the sleeveless Wahid). The issue might be that in v2 options not explicitly included were omitted while in v3 they get inherited.

### Design / Plugin / Package 🧐

designs/wahid

### Are you a FreeSewing patron? 😃

No, I am not 😞

BenJamesBen:
lab.freesewing.dev in Haxor green theme. `develop` lab in Light theme.
![Screenshot 2023-04-30 at 6 19 46 AM](https://user-images.githubusercontent.com/109869956/235355081-9c7768be-177e-4226-b9b4-7eeb16108186.png)
![Screenshot 2023-04-30 at 6 19 54 AM](https://user-images.githubusercontent.com/109869956/235355085-5a2a0ee5-b94a-406f-a45e-537c87dfa79f.png)
![Screenshot 2023-04-30 at 6 20 22 AM](https://user-images.githubusercontent.com/109869956/235355091-a2d4a25c-10ed-41ef-8a15-b4909ebfd4a0.png)
![Screenshot 2023-04-30 at 6 20 30 AM](https://user-images.githubusercontent.com/109869956/235355096-ffd2dda2-f011-4829-9077-a1be177becf5.png)

PR #3802, opened 2023-04-11
Update Wahid Options Config

bobgeorgethe3rd:
Should be a solve for #3005 

Some of Wahid's options were lost from v2 to v3. This PR should restore them.

Additions
- Added missing constant options to `options.mjs`
- Change `cuffEase` to constant
- Imported missing constants to `back.mjs`


joostdecock:
This is *not the way*. You should not include options from inherited patterns in v3. They will get added automatic.

As for #3005 we can look into a solution, but this is not it.

PR #3803, opened 2023-04-12
fix(docs): (sandy) Remove stray <Note> opening tag

BenJamesBen:
- Stray `<Note>` opening tag with no end tag.
- Also cleaned some trailing whitespace.


bobgeorgethe3rd (approved):
Ah good catch. Was probably a hold over from the streamlining

joostdecock (approved):

PR #3833, opened 2023-04-17
feat(shared): Add Mermaid support to MDX pipeline

joostdecock:
At the request of @marpants this adds support for mermaid diagrams to our shared MDX pipeline. Diagrams can be created by using a fenced code block using the `mermaid` language.

Screenshot 2023-04-17 at 20 02 23

I did not verify to what extend this bloats the bundle. Perhaps that's something we need to look into later and consider the trade-off. For now, my main concern is to keep a new contributor happy, so here it goes :)

### DIFF
--- a/config/dependencies.yaml
+++ b/config/dependencies.yaml
@@ -381,6 +381,7 @@ shared:
     'lodash.unset': *_unset
     'lodash.get': *_get
     'mdast-util-toc': '6.1.1'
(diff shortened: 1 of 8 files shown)


marpants (approved):
Thanks for these changes! I tried a few libraries and each of them had various errors.. But I learned a lot along the way :)

eriese (approved):

PR #3835, opened 2023-04-18
refactor (core) separate various Pattern concerns into helper classes

eriese:
This PR identifies the `Pattern` class as having 5 separate concerns:

1. Configuration resolution
2. Lifecycle hooks registration and management
3. Drafting
4. Rendering
5. Sampling

Configuration resolution had already been split out from the main class. This PR implements a new class for each of the remaining concerns, (hopefully) increasing readability/navigability of the code, and demystifying which functions pertain to which concerns.

The only public method in `Pattern` that has been removed is `createPartForSet`, which I implemented as a public method for use by a previous iteration of the `layout-part-plugin` in the lab, and which should probably never have been public in the first place. It has been rolled into the still-public `draftPartForSet`, which now provides a single call to draft a part out of cycle.


eriese:
@joostdecock It wants more documenting, but I think it's in a place where you can skim it to let me know if you like it or you think I've gone off the rails

joostdecock:
> @joostdecock It wants more documenting, but I think it's in a place where you can skim it to let me know if you like it or you think I've gone off the rails

I have skimmed it and I think we should have a discussion about where we want to go with core.
I think the current v3 implementation can be summarized as *make as little as changes possible to allow everything on the roadmap*. It is, in other words, a somewhat conservative approach, in anticipation of building the entire v3 environment.

My idea was to then build the frontend and re-evaluate what we are missing or could do different to allow us to do *cool stuff*.

I suspect that, since you have been doing *cool stuff* such as the cutting layout, you are at that stage where you are revisiting core to see how we can do better. And that's fine, and expected.

But I'm not there (yet). Nor is anyone else for that matter (insofar as anyone else is a core contributor).
So I am somewhat reluctant to make sweeping changes that are inspired by a potential subset of the frontend things we'd like to do and that would be easier/better/possible with changes to core.

At the same time, I struggle with the context switch that is required between frontend and core development. I feel that by going back and forward between them, I'm not getting any good work done.
So I guess what I am trying to say is that at least some of my remarks are not so much about your changes, but more about how I *feel* about the whole thing or how good or bad I am at switching between different things (bad).


With that out of the way, a list of things that gave me pause:

- Not sure we need to `pattern-x` everything. Like `pattern-plugin` for example. If we're going to split things up, why not just have `plugin.mjs` like we have `store.mjs` and not `pattern-store.mjs`.
- The way the PR is done (`patern.mjs` is removed) makes us lose all git history. That makes me nervous somewhat as I have on more than one occasion looked into the history to figure out what this or that was done in such a way.
- I don't like the (current, not your fault) way that the same terminology is used all over the place. Specifically `config` and `resolved config` and it just gets really hard to keep track of what is what. 

A (from the top of my head) list of things I'd like to think about/discuss/mull over:

- Should we add a new term for 'resolved config' (for example, `manifest`) and have a clear way to:
  - Create the initial manifest (aka resolve the config)
  - Allow run-time updates to the manifest (add parts and whatnot)
- Should we consider doing more things (or allowing people to do more things) in plugins. For example, why is SVG rendering in core?  Perhaps we should just put that in a plugin, and make rendering plugin-driven. Render to SVG with that plugin, render to React with this other plugin, ... and so on.
- The whole `sets` situation adds value, but can by hard to wrap your head around. Why is there no `sets.mjs` file?
- What a part is exactly does sort of depend on where we are in the lifecycle of a pattern. The parts that are passed a Design constructor are not really the same as what's in `Pattern.parts` and that is confusing. Perhaps we need to name them differently or find some other way to split this so it's easier to reason about


I guess to summarize, my main concerns are:
- I feel like I don't have enough time to dedicate to this now, or that the timing is awkward (for me)
- I feel that splitting files into more files can perhaps help a bit to clarify things. But perhaps the better fix a logical divide between similar but different things in the code base

But I'm confident we can figure it out together :)

eriese:
> I have skimmed it and I think we should have a discussion about where we want to go with core. I think the current v3 implementation can be summarized as _make as little as changes possible to allow everything on the roadmap_. It is, in other words, a somewhat conservative approach, in anticipation of building the entire v3 environment.
> 
> My idea was to then build the frontend and re-evaluate what we are missing or could do different to allow us to do _cool stuff_.
> 
> I suspect that, since you have been doing _cool stuff_ such as the cutting layout, you are at that stage where you are revisiting core to see how we can do better. And that's fine, and expected.
> 
> But I'm not there (yet). Nor is anyone else for that matter (insofar as anyone else is a core contributor). So I am somewhat reluctant to make sweeping changes that are inspired by a potential subset of the frontend things we'd like to do and that would be easier/better/possible with changes to core.

My concern about a conservative v3 is that if we don't take it as an opportunity to disentangle our very confusing and bloated `Pattern` class, we will very soon have to go to v4 to do so. My first refactor was about sweeping changes inspired by the frontend, but this refactor is about what I believe core needs even if it is to only do the things it already does. To that end, this is another refactor with almost no changes to tests, just more added to maintain coverage.

> * Not sure we need to `pattern-x` everything. Like `pattern-plugin` for example. If we're going to split things up, why not just have `plugin.mjs` like we have `store.mjs` and not `pattern-store.mjs`.

Totally. that's a holdover from before there was a pattern folder. But I also kind of like `pattern-x` as a convention because these are really all private helper classes for the `Pattern` class; they are not and should not be public facing, don't need API docs, etc.

> * The way the PR is done (`patern.mjs` is removed) makes us lose all git history. That makes me nervous somewhat as I have on more than one occasion looked into the history to figure out what this or that was done in such a way.

Github isn't showing the rename in the PR, but if you go to the files on this branch and you look at the history of `pattern/index/mjs`, the rename is there

> * I don't like the (current, not your fault) way that the same terminology is used all over the place. Specifically `config` and `resolved config` and it just gets really hard to keep track of what is what.
> 
> A (from the top of my head) list of things I'd like to think about/discuss/mull over:
> 
> * Should we add a new term for 'resolved config' (for example, `manifest`) and have a clear way to:
>   
>   * Create the initial manifest (aka resolve the config)
>   * Allow run-time updates to the manifest (add parts and whatnot)

Sounds good

> * Should we consider doing more things (or allowing people to do more things) in plugins. For example, why is SVG rendering in core?  Perhaps we should just put that in a plugin, and make rendering plugin-driven. Render to SVG with that plugin, render to React with this other plugin, ... and so on.

Into this too

> * The whole `sets` situation adds value, but can by hard to wrap your head around. Why is there no `sets.mjs` file?

The goal here was to separate out the main concerns, but I'm happy to get more granular with it. You know I love object oriented approaches. 

> * What a part is exactly does sort of depend on where we are in the lifecycle of a pattern. The parts that are passed a Design constructor are not really the same as what's in `Pattern.parts` and that is confusing. Perhaps we need to name them differently or find some other way to split this so it's easier to reason about

For sure


> * I feel like I don't have enough time to dedicate to this now, or that the timing is awkward (for me)

Totally fair. I think there isn't much harm in letting this sit; we'll just have to be careful if PRs come through with core changes

> * I feel that splitting files into more files can perhaps help a bit to clarify things. But perhaps the better fix a logical divide between similar but different things in the code base

Why not both?

joostdecock (approved):
After a second reading, I'm feeling YOLO

joostdecock:
Happy to merge this as-is. I mentioned some other candidates for refactoring in my comment above, but I think it's better to merge things in smaller bites, so that any unexpected issues crop up sooner rather than later.

Not merging, since it's in draft, but happy to when it's no longer.

PR #3841, opened 2023-04-18
Bob's Propsal for Macro Removal and Complete Usage

bobgeorgethe3rd:
Edit: 
This has now been change to proposal for Macro Removal and Complete Usage. 

So I am making this as a propsal on how to implement the Macro Removal and Complete usage. I feel having a static prefix string that is deleted is better for the user as they can identify what macros call what and can delete accordingly. I have also kept it as the `macro('macroname', false)` this may  be more of a personal preference but I feel `macro('rmmacroname`)` is more confusing and the only reason false is not widely adopted is because only a small percentage of plugins use it and is not documented on those that do.

I have also added so.detail. This acts as a marker that tells the plugin if it should be turned on or off by complete. It works in conjuction with complete like so `if ((complete && so.detail) || !so.detail)`

detail : true
The plugin works with complete toggle.

detail: false
the plugin ignores the complete toggle

I chose detail as complete is displayed as Details to the user.

I have not done every annonation yet as some need a bit more finesse than I can currently offer but this is a good overview of my proposal. I am aware we discussed having individual plugins toggleable by respective options, this should be compatible with that by changing the defaults of `detail`  to the option rather than its current `true`.

I have rebased this PR.

Design Plugintest
- Cleaned up the options
- Added false test
- Added isDetail test

Banner
- Added so.detail

Bannerbox
- Added complete
- Added so.detail

Crossbox
- Changed false flag to delete points based on string '_crossbox'
- Changed false flag to delete paths based on string '_crossbox'
- Removed prefix from default
- Declare id as either `'_' + so.id + '_crossbox'` or `'_crossbox'`
- Changed prefix to id
- added so.id
- Added complete
- Added so.detail

Cutonfold:
- Changed false flag to delete points based on string '_cutonfold'
- Changed false flag to delete paths based on string '_cutonfold'
- Removed prefix from default
- Declare idas either `'_' + so.id + '_cutonfold'` or `'_cutonfold'`
- Changed prefix to id
- Added so.detail

Miniscale 
- Changed false flag to delete points based on string '_miniscale'
- Changed false flag to delete paths based on string '_miniscale'
- Removed prefix from default
- Declare id as either `'_' + so.id + '_miniscale'` or `'_miniscale'`
- Changed prefix to id
- added so.id
- Added so.detail

Pleat:
- Changed false flag to delete points based on string '_pleat'
- Changed false flag to delete paths based on string '_pleat'
- Removed prefix from default
- Declare id as either`'_' + so.id + '_pleat'` or `'_pleat'`
- Changed prefix to id
- Added so.detail

Grainline
- Added so.detail
- Added id
- Added false
- Altered arrows to work the same as dimensions (yes this is not neccessary for this PR and needs to be altered either way to stop arrows overlapping paths)

Scalebox
- Changed false flag to delete points based on string '_scalebox'
- Changed false flag to delete paths based on string '_scalebox'
- Removed prefix from default
- Declare id as either `'_' + so.id + '_scalebox'` or `'_scalebox'`
- Changed prefix to id
- added so.id
- Added so.detail

SewTogether:
- Changed false flag to delete points based on string '_sewTogether'
- Changed false flag to delete paths based on string '_sewTogether'
- Removed prefix from default
- Declare id as either `'_' + so.id + '_sewTogether'` or `'_sewTogether'`
- Changed prefix to id
- Added so.detail

Title
- Changed false flag to delete points based on string '_title`
- Added complete
- Added so.detail


eriese (changes requested):
This is great! I'm thinking it might make sense for the default prefix to always start with an underscore, and then we can add the underscore to the matcher which will make me less nervous somehow about deleting non-macro points

Other than that, check out the one lint error that got raised, and some tests that got broken because of this, which will either want changes to the test expectations or the what's been broken.

joostdecock:
This is a great effort. But I'd love to see more ambition :)

Specifically, I think it would be good to make all behavior in plugin-annotations consistent in how they allow to *undo* a macro.
I'm open to suggestions on what would be the best behavior for this, but I feel the `rmd` and `rmad` macros are perhaps my favorite approach: https://github.com/freesewing/freesewing/blob/develop/plugins/plugin-annotations/src/dimensions.mjs

There are two ways we could handle this:

- Lean entirely on the `prefix` (or default prefix) for this. But that becomes harder when macros are used very often (such as dimensions)
- Use something like `id` instead of prefix which would serve the same purpose (it would be used in the name of generated points/paths) but would also take that same `id` as input for the equivalent `rmsomething` macro to remove it again.


As for Enoch's remark, it's perhaps a good idea to use `__` at teh start of all of these generated points and paths as that's how we mark *internal* stuff in core, so it would bring some consistency.

bobgeorgethe3rd:
> This is a great effort. But I'd love to see more ambition :)
> 
> Specifically, I think it would be good to make all behavior in plugin-annotations consistent in how they allow to _undo_ a macro. I'm open to suggestions on what would be the best behavior for this, but I feel the `rmd` and `rmad` macros are perhaps my favorite approach: https://github.com/freesewing/freesewing/blob/develop/plugins/plugin-annotations/src/dimensions.mjs
> 
> There are two ways we could handle this:
> 
>     * Lean entirely on the `prefix` (or default prefix) for this. But that becomes harder when macros are used very often (such as dimensions)
> 
>     * Use something like `id` instead of prefix which would serve the same purpose (it would be used in the name of generated points/paths) but would also take that same `id` as input for the equivalent `rmsomething` macro to remove it again.
> 
> 
> As for Enoch's remark, it's perhaps a good idea to use `__` at teh start of all of these generated points and paths as that's how we mark _internal_ stuff in core, so it would bring some consistency.

So we actually came to the same consensus that more can be done and I have already added this to the agenda for next meeting so a consensus can be reached on how best to implement it,

joostdecock:
That's perfect, thanks a million 🙏

bobgeorgethe3rd:
> * So, `macro('macroname', false)` is intended to remove all instances of the specified macro? There is no way to remove just one instance of a macro but leave others intact?
> 
>     * So, the `detail` property is intended to mean "allow `complete` to not display the macro, otherwise always display the macro"?

Yes, that is also what the current behaviour is suppose to do but does not. It's because you can't past the id through a false I believe but I could be wrong on that. If you need to delete and individual macro you would have to use deletes and be specific. 

Yes so when `detail` is true it is goverened by the complete value which is displayed as Details to the user. When `false` it ignores complete and will always show.

joostdecock:
Thanks for all input and discussion. This was taken on by @woutervdub and then further shoe-horned into the 3.0.0 release by myself. As such, this can be closed as it's implemented 🎉

PR #3842, opened 2023-04-18
fix(docs): Improve knit binding information and instructions

BenJamesBen:
This PR attempts to improve the knit binding documentation, especially to add information about the approximate expected length of binding that will be used. I also removed instances of "I" and "my" language, since the documentation has no listed author.

This documentation update came from the Discord discussion on/about April 1, 2023 in #sewing-help:
https://discord.com/channels/698854858052075530/757632180980547686/1092482399973163118
discord://discord.com/channels/698854858052075530/757632180980547686/1092482399973163118

### DIFF
--- a/markdown/org/docs/sewing/knit-binding/en.md
+++ b/markdown/org/docs/sewing/knit-binding/en.md
@@ -52,7 +52,8 @@ Whether you're using a knit band or a knit binding finish, you are going to have
 
 <Note>
 
-The remainder of this documentation uses the term _binding_ to refer to both knit bands and knit binding.
+The remainder of this documentation refers to _binding_ but applies equally
+to knit bands.
 
 </Note>
 
@@ -107,18 +108,25 @@ just enough, and cutting off the excess.
 ### Method 1: Ring method
 
 1. Measure the opening seam that the binding will cover.
+
 2. Check the pattern instructions for guidance on the length of binding to use.
 Or, use a length 70-85% smaller than the opening based on the stretchiness
 of the binding fabric (For example, 70% for fabrics with a stretch of 40% or greater,
 85% for those with a stretch of 10% or less).
+
 3. Cut the length of binding required, plus seam allowance at each end.
+
 4. Sew the binding together to form a closed loop or ring.
+
 5. Mark or insert pins into the ring in 4 places, starting at the seam
 to divide the ring into 4 quarters of equal length.
+
 6. Mark or insert pins into the garment opening in 4 places,
 dividing the opening into 4 quarters of equal length.
+
 7. Match the marks or pins on the ring with the ones on the garment opening, and pin the binding ring to the opening in 4 places, at each quarter
 mark/pin.
+
 8. While evenly stretching each quarter segment of the binding ring into
 its quarter segment of garment opening, sew the binding to the opening.
 
@@ -137,17 +145,23 @@ suboptimal distribution of binding in the opening.
 ### Method 2: Stretching method
 
 1. Cut your binding exactly as long as the seam it needs to be sewn into. Not 90%, but 100%. This way, you know it's going to be too long, and that's what we want.
+
 2. Mark the place you want your binding to be joined (say center back of the neckline) and place your binding 1 cm beyond this point.
+
 3. Start sewing your binding about 3 cm or so from this point, and go all the way around. While you do so, make sure to adjust the stretch to whatever the curve requires. There are no tricks for this; you'll have to practice until it comes naturally to you.
-4. When you make your way around and get close to the point you marked, stop about 3 cm before that. Now your entire binding is attached, apart from a 6 cm or so stretch at the back.
-5. Pin both edges of your binding good sides together to form the binding into a closed loop, making it exactly as long as needed.
-6. Cut off the excess binding, and sew the binding loop closed.
-7. Sew this remaining 6 cm or so binding loop to the garment.
 
 <Tip>
 While it is impossible to know the exact length of binding you will end
 up using, lengths of 70-85% of the garment opening are common.
 </Tip>
+
+4. When you make your way around and get close to the point you marked, stop about 3 cm before that. Now your entire binding is attached, apart from a 6 cm or so stretch at the back.
+
+5. Pin both edges of your binding good sides together to form the binding into a closed loop, making it exactly as long as needed.
+
+6. Cut off the excess binding, and sew the binding loop closed.
+
+7. Sew this remaining 6 cm or so binding loop to the garment.
 
 Pros: Some people feel that this is the best method to use because it always
 results in the exact length of binding needed being used.
--- a/markdown/org/docs/sewing/knit-binding/en.md
+++ b/markdown/org/docs/sewing/knit-binding/en.md
@@ -19,7 +19,7 @@ This leaves an unfinished edge on the inside that is often sewn down by edge-sti
 
 ### Knit binding
 
-![This woman's grey undershirt and white T-shirt both have a knit binding at the neckline.](knit_binding.jpg)
+![This grey undershirt and white T-shirt both have a knit binding at the neckline.](knit_binding.jpg)
 
 Real knit binding acts more like classic bias binding, in that it wraps around the unfinished edge of the garment, rather than extend from it.
 
--- a/markdown/org/docs/sewing/knit-binding/en.md
+++ b/markdown/org/docs/sewing/knit-binding/en.md
@@ -6,21 +6,22 @@ title: Knit binding
 
 Knit binding is a general term for a type of finish (typically) found on the neck or arm opening of T-shirts and other knitwear.
 
-There's two varieties of this finish, a knit band, and actual knit binding.
+There are two varieties of this finish, a knit _band_ and actual knit _binding_.
 
 ## Knit band
 
-![Like most T-shirts, the one this girl is wearing has a knit band at the neck opening.](knit_band.jpg)
+![Like most T-shirts, the one this woman is wearing has a knit band at the neck opening.](knit_band.jpg)
 
-Knit band is what you typically find at the neck opening of your T-shirt. It's a strip of fabric that is folded double and attached to the edge of the garment.
+Knit band is what you typically find at the neck opening of your T-shirt. It's a strip of fabric that is folded double and attached to the edge of the garment
+such that some of the fabric strip extends beyond the unfinished edge of garment.
 
 This leaves an unfinished edge on the inside that is often sewn down by edge-stitching around the knit band.
 
 ### Knit binding
 
-![This girl's grey undershirt and white T-shirt both have a knit binding at the neckline.](knit_binding.jpg)
+![This woman's grey undershirt and white T-shirt both have a knit binding at the neckline.](knit_binding.jpg)
 
-Real knit binding acts more like classic bias binding, in that it is wrapped around the fabric, rather than extend from it.
+Real knit binding acts more like classic bias binding, in that it wraps around the unfinished edge of the garment, rather than extend from it.
 
 The raw edge of the garment is inserted between the folded knit binding, which is edge-stitched down.
 This is often done with a coverlock machine, thereby neatly finishing the inside.
@@ -30,16 +31,16 @@ This is often done with a coverlock machine, thereby neatly finishing the inside
 
 ### Style vs function
 
-A knit band is more common, and is what you typically find on T-shirt necklines. It's the more decorative option.
+A knit band is more common and is what you typically find on T-shirt necklines. It's the more decorative option.
 
-A knit binding is less common, but often used on undershirts (as in, underwear). It's the more functional option.
+A knit binding is less common but often used on undershirts (as in, underwear). It's the more functional option.
 
 > ##### Still confused?
 >
 > Another way to distinguish them is this: Think about what happens when you decide to widen the knit binding/band.
 >
 > On a T-shirt, widening the band at the neck will make the neck opening smaller.
-> If you want a wider band, you will have to cut a larger hole in your T-shirt.
+> If you want a wider band with the same finished neck opening size, you will have to cut a larger neck hole in your T-shirt.
 >
 > On an undershirt, if you want the knitwear binding at the neck to be wider,
 > it will not make the neck opening smaller.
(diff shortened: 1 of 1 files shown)


markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -6,21 +6,22 @@ title: Knit binding
 
 Knit binding is a general term for a type of finish (typically) found on the neck or arm opening of T-shirts and other knitwear.
 
-There's two varieties of this finish, a knit band, and actual knit binding.
+There are two varieties of this finish, a knit _band_ and actual knit _binding_.
 
 ## Knit band
 
-![Like most T-shirts, the one this girl is wearing has a knit band at the neck opening.](knit_band.jpg)
+![Like most T-shirts, the one this woman is wearing has a knit band at the neck opening.](knit_band.jpg)
```
luizfzs, line 13: There in only one person in the image, so it seems redundant to mention `woman` here (and also here https://github.com/freesewing/freesewing/pull/3842/files#diff-4fa783e1a15b6063eee9d93cd61d9c8d14b5e8190afc2469ad020bb412ba8f13R22)

```suggestion
![Like most T-shirts, this one has a knit band at the neck opening.](knit_band.jpg)
```

markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -6,21 +6,22 @@ title: Knit binding
...
-Knit band is what you typically find at the neck opening of your T-shirt. It's a strip of fabric that is folded double and attached to the edge of the garment.
+Knit band is what you typically find at the neck opening of your T-shirt. It's a strip of fabric that is folded double and attached to the edge of the garment
+such that some of the fabric strip extends beyond the unfinished edge of garment.
 
 This leaves an unfinished edge on the inside that is often sewn down by edge-stitching around the knit band.
 
 ### Knit binding
 
-![This girl's grey undershirt and white T-shirt both have a knit binding at the neckline.](knit_binding.jpg)
+![This woman's grey undershirt and white T-shirt both have a knit binding at the neckline.](knit_binding.jpg)
```
luizfzs, line 22:
```suggestion
![Both T-shirts have a knit binding at the neckline.](knit_binding.jpg)
```

luizfzs (commented):
Thanks for updating this document!

joostdecock (approved):

luizfzs (dismissed):

joostdecock:
@BenJamesBen Is this still a draft, are can we merge it?

markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -79,41 +80,50 @@ Stretch your binding too much, and it will spring back too far, causing the main
...
-- Now your entire binding is attached, apart from a 6cm or so stretch at the back. Pin both edges of your binding good sides together, making is exactly as long as needed and sew them together.
-- Sew the remaining 6cm of your binding.
+1. Cut your binding exactly as long as the seam it needs to be sewn into. Not 90%, but 100%. This way, you know it's going to be too long, and that's what we want.
+2. Mark the place you want your binding to be joined (say center back of the neckline) and place your binding 1 cm beyond this point.
+3. Start sewing your binding about 3 cm or so from this point, and go all the way around. While you do so, make sure to adjust the stretch to whatever the curve requires. There are no tricks for this; you'll have to practice until it comes naturally to you.
+4. When you make your way around and get close to the point you marked, stop about 3 cm before that. Now your entire binding is attached, apart from a 6 cm or so stretch at the back.
+5. Pin both edges of your binding good sides together to form the binding into a closed loop, making it exactly as long as needed.
+6. Cut off the excess binding, and sew the binding loop closed.
```
SeaZeeZee, line 117: This part I think is the relevant part that we needed to clarify people's problems with the documentation.

markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -79,41 +80,50 @@ Stretch your binding too much, and it will spring back too far, causing the main
...
-On a slight curve, the difference between the inner and outer edge is not so big, and a little bit of stretch is all you need. But on a tighter curve, the difference between the inner and outer edge of your binding will be more significant. And you'll need more stretch to accommodate for that.
+On a slight curve, the difference between the inner and outer edge is not so big, and a little bit of stretch is all you need. But on a tighter curve, the difference between the inner and outer edge of your binding will be more significant. And, you'll need more stretch to accommodate for that.
 
-So as you follow your seam, and it curves around necklines, or shoulders, or design features, you will have to adapt the amount of stretch to the situation.
+So as you follow your seam and as it curves around necklines, shoulders, or design features, you will have to adapt the amount of stretch to the situation.
 
 ## You cannot calculate the length of your knit binding
 
-I am not saying it is theoretically impossible. But I am saying that it is practically impossible.
+It may not be theoretically impossible, but it is practically impossible.
```
SeaZeeZee, line 93: I am concerned about the fact that this change by removing personal language suggests this is always true. I don't think that is the case.

markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -79,41 +80,50 @@ Stretch your binding too much, and it will spring back too far, causing the main
...
-The length of your knit binding depends on the amount it needs to be stretched. This depends on the curves you have to follow, and the width of the binding, not to mention the strength of its stretchiness. There's just too many parameters to take into account to calculate your binding length accurately.
+The length of your knit binding depends on the amount it needs to be stretched. This depends on the curves you have to follow and the width of the binding, not to mention the strength of its stretchiness. There are just too many parameters to take into account to calculate your binding length accurately.
 
 ### Do not do this
 
 There are a bunch of tutorials out on the internet that show you how to sew on knit binding.
 Many of those suggest something like _make the neck binding 90% of the length of the seam you are attaching it to_ (the percentage varies).
 
-This idea is that you cut your binding, join its edges, and then sew it in the opening. That is (in my opinion) no good.
+This idea is that you would cut your binding, join its edges to form a closed loop, and then sew it to the opening. However, this is not the best method.
```
SeaZeeZee, line 102: I have the same issue here as above, before it said the author of this documentation (I assume Joost wrote this) thinks this method is no good. Now the documentation reads as if this is _never_ the right method. I think people should feel free to use the method that works for them and this unequivocal language sends the opposite message. Perhaps we need to change this to 'we' rather than 'I' so it still signals that this is someone's opinion rather than solid fact.

markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -79,41 +80,50 @@ Stretch your binding too much, and it will spring back too far, causing the main
...
 
-An alternative approach to sidestep the length issue is to attach the binding before you close the opening your binding goes in.
-For example, on a T-shirt you close one shoulder seam, attach the binding, then close the second shoulder seam.
+A possible alternate approach to sidestep the length issue would be to attach the binding before you close the opening your binding goes in.
+For example, on a T-shirt you would close one shoulder seam, attach the binding, then close the second shoulder seam.
 
-Obviously, that means your shoulder seam will cut through your binding which is a big no no in my book.
+However, that obviously means your shoulder seam would cut through your binding.
+So, this approach is also not desirable.
```
SeaZeeZee, line 108: The same issue as earlier, this again changes the way the documentation lands with the reader.

SeaZeeZee (changes requested):
I have some concerns about what the changes in language means for how documentation is presented. See comments.

markdown/org/docs/sewing/knit-binding/en.md
luizfzs: I've sewed 2 knit bands so far. What I do, after the pattern is printed and cut, is to use thread to measure the opening, then measure the thread length. I 'walk it off' on the opening using thread. 
Maybe this 'trick' can be written down in this document as well?

What this refers to IMO, is that it may not be possible or feasible to get that measurement in the digital version of the pattern. 

I'm not saying to not discuss it but personally I prefer documentation to be impersonal.

markdown/org/docs/sewing/knit-binding/en.md
luizfzs: What about if we have a section with `Alternative techniques`? It could include this one and others that may not be as reliable as this, which in itself is subjective.

markdown/org/docs/sewing/knit-binding/en.md
SeaZeeZee: I think the reason there's no author stated is that this was written before Freesewing was on GitHub, hence my assumption Joost wrote this. 

We can definitely have a discussion in a contributor call about how we want the freesewing documentation around sewing to sound. 

I personally do my best to not change the original writing or tone when I change or review documentation unless there is an issue around clarity or some kind of error.

BenJamesBen:
Based on feedback here and also based on the previous Discord discussion, I've made major changes to this documentation:
- Added another method for attaching binding
- Presented both methods in a neutral tone, rather than favoring one over the other

Previously, the documentation advocated one particular method as an anonymous personal opinion. I originally changed the wording to remove the personal language, keeping the recommendation but making it from the general FreeSewing community. This latest change removes the recommendation and instead presents two methods without recommending one over the other.

Again, I am just trying to improve this documentation, trying to take into consideration differing opinions, and trying to work towards consensus. If this is something that needs to be discussed on a Contributor Call, we can do so.

markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -79,41 +80,50 @@ Stretch your binding too much, and it will spring back too far, causing the main
...
+2. Mark the place you want your binding to be joined (say center back of the neckline) and place your binding 1 cm beyond this point.
+3. Start sewing your binding about 3 cm or so from this point, and go all the way around. While you do so, make sure to adjust the stretch to whatever the curve requires. There are no tricks for this; you'll have to practice until it comes naturally to you.
+4. When you make your way around and get close to the point you marked, stop about 3 cm before that. Now your entire binding is attached, apart from a 6 cm or so stretch at the back.
+5. Pin both edges of your binding good sides together to form the binding into a closed loop, making it exactly as long as needed.
+6. Cut off the excess binding, and sew the binding loop closed.
+7. Sew this remaining 6 cm or so binding loop to the garment.
 
 This way, there is no guessing how long your binding should be, and the binding seam sits where you want it to sit.
 
+<Tip>
```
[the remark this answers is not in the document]
anna-puk: That's exactly the point that's being made here... the only thing we know for sure is that it shouldn't be _longer_ than the seam itself.

markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -49,11 +50,17 @@ A knit binding is less common, but often used on undershirts (as in, underwear).
 
 Whether you're using a knit band or a knit binding finish, you are going to have to stretch your binding fabric for good results. That is, assuming you're not sewing a straight seam. Here's why:
 
+<Note>
+
+The remainder of this documentation uses the term _binding_ to refer to both knit bands and knit binding.
```
anna-puk, line 55: I think it's fine to use only one term, since it makes the document easier to read, but 'binding' and 'band' are not the same thing. I suggest rephrasing to "The remainder of this documentation refers to _binding_, but applies equally to knit bands."

markdown/org/docs/sewing/knit-binding/en.md
```diff
@@ -79,41 +86,78 @@ Stretch your binding too much, and it will spring back too far, causing the main
...
+1. Cut your binding exactly as long as the seam it needs to be sewn into. Not 90%, but 100%. This way, you know it's going to be too long, and that's what we want.
+2. Mark the place you want your binding to be joined (say center back of the neckline) and place your binding 1 cm beyond this point.
+3. Start sewing your binding about 3 cm or so from this point, and go all the way around. While you do so, make sure to adjust the stretch to whatever the curve requires. There are no tricks for this; you'll have to practice until it comes naturally to you.
+4. When you make your way around and get close to the point you marked, stop about 3 cm before that. Now your entire binding is attached, apart from a 6 cm or so stretch at the back.
+5. Pin both edges of your binding good sides together to form the binding into a closed loop, making it exactly as long as needed.
+6. Cut off the excess binding, and sew the binding loop closed.
+7. Sew this remaining 6 cm or so binding loop to the garment.
+
+<Tip>
+While it is impossible to know the exact length of binding you will end
```
anna-puk, line 148: Can we move this tip up to step 3, which currently only says "you'll have to practice"? I think this will be more helpful.

markdown/org/docs/sewing/knit-binding/en.md
BenJamesBen: I changed the wording to "This grey undershirt and white T-shirt both" (since apparently one of the shirts isn't a T-shirt, technically).

SeaZeeZee:
I've taken myself off of the review list because I don't currently have the energy or time to review this documentation. I think it would be good to have a conversation in a contributor call about the tone we want the sewing related documentation to have and whether  folks feel personal language is appropriate or not.

markdown/org/docs/sewing/knit-binding/en.md
BenJamesBen: I moved the Tip up to Step 3.

markdown/org/docs/sewing/knit-binding/en.md
BenJamesBen: I changed the wording as suggested.

BenJamesBen:
I added the topic to the April 29 Contributor Call.

SeaZeeZee:
Hey @BenJamesBen since this is your pr, with what we talked about in the contributor call what do you think we should do?

BenJamesBen:
I will be revising this PR entirely based on the guidance from the last Contributor Call, and I'll resubmit the PR then.

(Process-wise, I'm not sure if I should close this PR and re-open it? Or, just leave it open for now?)

BenJamesBen:
I reverted all my previous edits to the knit binding documentation. Instead, I've simply added a new section to address the alternate method.

(I still think that the existing documentation can be improved by adding the explicit step of cutting off excess binding, but I'll let someone else handle it, if they feel it is warranted.)

joostdecock (approved):

PR #3843, opened 2023-04-18
fix (banner) make banner macro able to use i18n

eriese:
The issue: because the banner macro was adding the entire banner to the path attributes at once, passing an i18n key to the banner macro could not result in translation. This PR adds the text separately from the spacers to allow for i18n during rendering.

A minor side effect is that because `path.attributes.get('data-text')` uses a `[].join(' ')` there are spare spaces in the final string, but because they are breaking spaces, they do not make a noticeable difference as far as I can tell


joostdecock (approved):
Good catch!

Issue #3844, opened 2023-04-19
[bug]: (new-design) Dev environment missing Mermaid support and unable to run


BenJamesBen:
### What seems to be the problem? 🤔

`npx @freesewing/new-design@next` standalone environment installs are unable to run due to missing Mermaid support.
```sh
% yarn dev
...
error - ./shared/components/mdx/highlight.mjs:3:0
Module not found: Can't resolve './mermaid.mjs'
  1 | import { CopyToClipboard } from 'shared/components/copy-to-clipboard.mjs'
  2 | import { HttpStatusCode } from './http.mjs'
> 3 | import { Mermaid } from './mermaid.mjs'
  4 | 
  5 | const names = {
  6 |   js: 'Javascript',
```

### Design / Plugin / Package 🧐

packages/new-design

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

A workaround is to:
1. copy the file at `https://raw.githubusercontent.com/freesewing/freesewing/develop/sites/shared/components/mdx/mermaid.mjs` to `shared/components/mdx/mermaid.mjs` and 
2. `yarn add mermaid` in the created folder

marpants:
I believe the bug will be resolved with the changes in #3845 , however I see a new problem now which I haven't yet found a root cause for:
```
tutorial$ yarn dev
yarn run v1.22.19
$ next dev -p 8000
ready - started server on 0.0.0.0:8000, url: http://localhost:8000
warn  - Invalid next.config.mjs options detected: 
warn  -     The value at .i18n has an unexpected property, defaultNS, which is not in the list of allowed properties (defaultLocale, domains, localeDetection, locales).
warn  - See more info here: https://nextjs.org/docs/messages/invalid-next-config

🌼 daisyUI components 2.51.5  https://daisyui.com
  ✔︎ Including:  base, components, 4 themes, utilities
  ❤︎ Support daisyUI:  https://opencollective.com/daisyui 
  
event - compiled client and server successfully in 3.7s (218 modules)
wait  - compiling / (client and server)...
event - compiled client and server successfully in 1364 ms (389 modules)
error - Error: Default namespace not found at public/locales/en/common.json
    at createConfig (/home/marpants/freesewing/packages/new-design/tutorial/node_modules/next-i18next/dist/commonjs/config/createConfig.js:165:19)
[...]

PR #3845, opened 2023-04-19
(bugfix #3844) mermaid missing from new-design package

marpants:
Verified with local build/installing of new-design package to see that package.json is correct and mermaid.mjs file is copied into output code, however the actual site is still not running correctly, I believe due to another bug. The error I'm getting is 
"Error: Default namespace not found at public/locales/en/common.json"
Still, I think this code could be merged to fix the issue #3844


BenJamesBen (commented):
I don't know enough about how these packages are built to be able to comment on the code changes, if they are sufficient, etc.

joostdecock (approved):

joostdecock:
> "Error: Default namespace not found at public/locales/en/common.json"

@marpants You should run `yarn prebuild` in the `sites/dev` folder before running `yarn dev`. 
That will fix it as it will generate the language files that are missing ;)

joostdecock:
@all-contributors please add @marpants for code

Issue #3854, opened 2023-04-20
[bug]: develop branch broken after 3.0.0-alpha.9 update.


marpants:
### What seems to be the problem? 🤔

@BenJamesBen reported an issue via the discord with the latest update of the develop branch.
```
% yarn kickstart
yarn run v1.22.19
$ npx lerna bootstrap && yarn reconfigure && yarn buildall && yarn prepare && yarn tips
lerna notice cli v6.6.1
lerna info bootstrap root only
[1/4] 🔍  Resolving packages...
error Couldn't find any versions for "@freesewing/plugin-banner" that matches "3.0.0-alpha.9"
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: Couldn't find any versions for "@freesewing/plugin-buttons" that matches "3.0.0-alpha.9"
...
```
It appears that certain plugins were removed or bundled together at some point but the dependencies/package.jsons were not updated appropriately.
The commit where the packages were removed is here:
https://github.com/freesewing/freesewing/commit/f4f7562dd7068429ee9fac2b147f3c5f1a4b9633

### Design / Plugin / Package 🧐

plugins/plugin-banner, plugins/plugin-buttons, plugins/plugin-dimension, plugins/plugin-grainline, plugins/plugin-logo, plugins/plugin-scalebox

### Are you a FreeSewing patron? 😃

No, I am not 😞

PR #3855, opened 2023-04-20
Remove offending packages contributing to build errors (bug: #3854)

marpants:
So I've only been able to to minimal testing with this... 
I've started the lab and dev environments from the mono repo, but additional validation might be needed...
Starting this PR because at least with this fix we can build and identify any remaining issues


joostdecock (approved):

joostdecock:
Hey thanks. I saw the error this morning and fixed it before realizing you had put up this PR (hadn't had coffee yet) but it turns out it took both of our efforts to get everything, so nice work :)

FWIW we manage dependencies locally (as you've noticed) to ensure we only use one version for shared packages.
To re-generate the various `package.json` files you can run `yarn reconfigure` in the monorepo root. If you do so after you make changes in the `config/dependencies.yaml` file it will ensure those changes go where they need to go.

marpants:
Great thank you for the info :)

Issue #3819, opened 2023-04-15
[feature]: Allow overwriting of snippets by plugins


joostdecock:
### What is it that you would like to see happen? 🤔

This comes on the heels of #3625 and what I'd like to see is to make it easier tor plugins to overwrite what other plugins provide.

Specifically, there was a bit of discussion on whether or not to keep `logo` in the plugin-annotions package, or keep it out because it's typically something people will want to overwrite when they implement their own designs.

However, we must assume that almost everyone loads the plugin-bundle package, which also includes the logo. So keeping it out of the annotations plugin is in practice probable not all that helpful, not to mention that it's a bit counter-intuitive.

So, instead it would be better if people could just load their own plugin that would provide a `logo` snippet that simply overwrites the previous logo.

However, the way things are handled right now this would not work because plugins set snippet definitions for the SVG `defs` section by adding it to `svg.defs` which is a string.

We should isntead extend the svg object so that defs can by set together with an ID (and concatenated for rendering) so that people can do something like 

```js
svg.defs.set('logo', 'my new logo here')
```

This would make it a lot simpler for people to overwrite some of the default snippets. And it would mean we don't have to worry about what we do or do not bundle.

### Are you a FreeSewing patron? 😃

Yes, I am a tier-8 patron ❤️  💙 💜

### Additional context

This would predominantly impact core and the new annotations plugin.

woutervdub:
@joostdecock ,

Where / how were you envisioning this code would be used by someone who would want to replace our logo?
`svg.defs.set('logo', 'my new logo here')`

joostdecock:
So in the plugin, we now typically have something like:

```js
svg.defs += "[svg string here]"
```

I'd like to use something like this:

```js
svg.defs.setIfUnset('logo', "[svg string here]")
```

For this, I'd propose to make a light version of the current attributes facility that has `set`, `setIfUnset`, `unset`, and `toString` methods.

We might want to re-use it for setting CSS too. Since that's also kept as one big string right now.

joostdecock:
By hooking into the lifecycle and setting the snippet themselves. Either through a plugin, or via the `pattern.on()` method.

That part doesn't change. It's just that now it will be possible to do so without generating an ID conflict in the SVG output.

PR #3863, opened 2023-04-20
Allow `defs` in the `svg` to be overwritten.

woutervdub:
Implements #3819


eriese (changes requested):
Great work!

My one nitpick is that I think it'd be ideal if all the snippets were treated the way you've done `buttons`, i.e. each individual `snippet` gets added to `defs` via its own `id`, rather than all being grouped by function

woutervdub:
The buttons one is different because there were multiple `defs` in the original. Each of those `defs` is itself made up from separate `id`s. So the `cutOnFold` is really the same as `snaps`, whereas `cutOnFold` is in its own mjs source file, while `snaps` is combined in the buttons mjs source file.

woutervdub:
I guess I broke a bunch of tests.

packages/core/tests/svg.test.mjs
```diff
@@ -42,7 +43,7 @@ describe('Svg', () => {
     expect(svg.freeId).to.equal(0)
     expect(svg.body).to.equal('')
     expect(svg.style).to.equal('')
-    expect(svg.defs).to.equal('')
+    expect(svg.defs instanceof Defs).to.equal(true)
```
eriese, line 46: there's a handy assertion you can use here that's a little more readable. `expect(svg.defs).to.be.an.instanceof(Defs)`. It'll throw a better error

plugins/plugin-annotations/tests/buttons.test.mjs
```diff
@@ -11,7 +11,7 @@ pattern.draft().render()
 describe('Buttons Plugin Test', () => {
   for (const snippet of ['button', 'buttonhole', 'snap-stud', 'snap-socket']) {
     it(`Should add the ${snippet} snippet to defs`, () => {
-      expect(pattern.svg.defs.indexOf(`<g id="${snippet}">`)).to.not.equal(-1)
+      expect(pattern.svg.defs.render().indexOf(`<g id="${snippet}">`)).to.not.equal(-1)
```
eriese, line 14: change this to `expect(pattern.svg.defs).to.have.property(snippet)`

plugins/plugin-annotations/tests/logo.test.mjs
```diff
@@ -9,7 +9,7 @@ describe('Logo Plugin Tests', () => {
     const Pattern = new Design()
     const pattern = new Pattern().use(annotationsPlugin)
     pattern.draft().render()
-    expect(pattern.svg.defs).to.contain(
+    expect(pattern.svg.defs.render()).to.contain(
```
eriese, line 12: same here. `expect(pattern.svg.defs).to.have.property('logo')`

plugins/plugin-annotations/tests/notches.test.mjs
```diff
@@ -17,11 +17,11 @@ pattern.draft().render()
 
 describe('Notches Plugin Test', () => {
   it(`Should add the snippets to defs`, () => {
-    expect(pattern.svg.defs).to.contain('<g id="notch">')
+    expect(pattern.svg.defs.render()).to.contain('<g id="notch">')
   })
 
   it(`Should add the notches snippet to defs`, () => {
-    expect(pattern.svg.defs.indexOf(`<g id="notch">`)).to.not.equal(-1)
+    expect(pattern.svg.defs.render().indexOf(`<g id="notch">`)).to.not.equal(-1)
```
eriese, line 24: These as well

eriese (dismissed):

eriese:
It's looking really good. Only thing is test coverage. I'd say if you haven't seen a use case for `Defs.prototype.clone()` you could get rid of it and save yourself a test

eriese (approved):

BenJamesBen:
I do not know if these are related to this PR's changes or if they simply might be due to this PR being based on older code, but I notice that in Cutting Layout:
1. Parts no longer align with grain when the grainline is toggled to horizontal.
2. Cutonfold parts are no longer mirrored/splayed.
3. Some parts no longer show via `complete`, like Carlita's `carlton.back`.

eriese:
I think that's because this branch is 230 commits behind `develop`, and those things were broken in the move to `plugin-annotations`, but I think it's probably best for @woutervdub to merge in develop to find out

eriese:
@joostdecock I'm good with this, but don't want to merge it without your approval because I think you had a point of view about how this should be implemented

joostdecock:
I had a small remark but it's not something @woutervdub did, just something I noticed while reviewing this.
So I've merged this (great work) and created: #3965 3965

Issue #3834, opened 2023-04-18
[bug]: Each click on 'Preview' or 'X-ray' enlarges the 'Sample' example on dev 'Sets' page


marpants:
### What seems to be the problem? 🤔

Pretty sure this is unintended behavior, wasn't able to figure out why it's happening yet. I also couldn't find an open bug for this already but can't be sure.

Bug:
The 'sample' example preview on the following page https://freesewing.dev/guides/patterns/sets gets bigger when clicking between 'Preview' and 'X-Ray'.

Behavior:
On page load:
![image](https://user-images.githubusercontent.com/61366665/232669160-8b0014ed-b285-41ce-aa3d-e35e30b52eaa.png)
Click 'X-ray' then 'Preview'
![image](https://user-images.githubusercontent.com/61366665/232669179-a3088f63-8e44-4bcf-a55c-01fcefb61272.png)
Multiple clicks later
![image](https://user-images.githubusercontent.com/61366665/232669205-308c8d86-4d0d-4b85-a810-51425559f49b.png)

### Are you a FreeSewing patron? 😃

No, I am not 😞

PR #3875, opened 2023-04-23
fix (#3834) Use one dRaft sample for all example tabs

eriese:
The bug was happening because that particular example was sampling a measurement. 

`Pattern.__measurementSets()`, called during measurement sampling, creates a new array of settings sets starting at `this.settings[0].measurements[measurementName] * 0.9`. Because the `Example` component was being passed an existing `Pattern` instance and then sampling it on render, every time the tabs were switched a new `Example` component was being mounted, causing the instance to be re-sampled and measurement being sampled to be reduced by 10%. The logo appeared to be getting bigger because it was staying the same size and the box was getting smaller.

This PR fixes the issue by drafting or sampling the pattern at the `TabbedExample` level, and passing the returned renderProps to the `Example` components instead.

It also fixes a compiler complaint about `fillOpacity` that came up while I was investigating the pattern sets documentation

### DIFF
--- a/markdown/dev/guides/patterns/sets/en.md
+++ b/markdown/dev/guides/patterns/sets/en.md
@@ -18,16 +18,17 @@ _combine_ multiple drafted variants into a single pattern container.
 Here's a simple example:
 
 <Example settings="sample: { type: measurement, measurement: head }" withHead caption="A simple example of sampling the `head` measurement">
+
 ```js
-({ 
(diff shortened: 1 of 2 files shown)


eriese:
This branch is apparently an old one that now has an unrelated history with `develop`, so I'm closing this one and opening a new one with a working branch

PR #3877, opened 2023-04-24
Fix (#3834) use one draft for all example tabs

eriese:
The bug was happening because that particular example was sampling a measurement.

Pattern.__measurementSets(), called during measurement sampling, creates a new array of settings sets starting at this.settings[0].measurements[measurementName] * 0.9. Because the Example component was being passed an existing Pattern instance and then sampling it on render, every time the tabs were switched a new Example component was being mounted, causing the instance to be re-sampled and measurement being sampled to be reduced by 10%. The logo appeared to be getting bigger because it was staying the same size and the box was getting smaller.

This PR fixes the issue by drafting or sampling the pattern at the TabbedExample level, and passing the returned renderProps to the Example components instead.

It also fixes a compiler complaint about fillOpacity that came up while I was investigating the pattern sets documentation

### DIFF
--- a/markdown/dev/guides/patterns/sets/en.md
+++ b/markdown/dev/guides/patterns/sets/en.md
@@ -18,16 +18,17 @@ _combine_ multiple drafted variants into a single pattern container.
 Here's a simple example:
 
 <Example settings="sample: { type: measurement, measurement: head }" withHead caption="A simple example of sampling the `head` measurement">
+
 ```js
-({ 
-  Point, 
-  points, 
-  Path, 
-  paths, 
-  Snippet, 
-  snippets, 
-  measurements, 
-  part 
+({
+  Point,
+  points,
+  Path,
+  paths,
+  Snippet,
+  snippets,
+  measurements,
+  part
 }) => {
 
   const size = measurements.head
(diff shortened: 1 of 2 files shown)


markdown/dev/guides/patterns/sets/en.md
```diff
@@ -145,17 +148,17 @@ This is illustrated below:
 
   // Stacks
   drawBox('Stack 0', -30, -33, 50, 169, 'b', 'mark', 'fill-mark')
-  paths['Stack 0'].attr('fill-opacity', 0.2)
+  paths['Stack 0'].attr('fillOpacity', 0.2)
```
joostdecock, line 151: Pretty certain this will break SVG output.

joostdecock (changes requested):
`Path.attr()` is not react code. We should set the proper attribute name. We can rename these sort of attributes in in our react renderer so it works there too. But this is not the place to set `fillOpacity` rather than `fill-opacity`.

eriese:
Makes sense. So, something that camel cases attributes? Probably I'll
remove this from this PR and do that separately

joostdecock (approved):

PR #3888, opened 2023-04-25
Test (new-design) add a way to test new-design

eriese:
- Adds capability to optionally copy e2e testing into generated designs
- Adds a github workflow that generates designs from **scratch**, from **tutorial**, and from **bent** and runs the e2e tests on them


joostdecock (approved):

joostdecock:
We should probably tell eslint ti ignore `*.mustache` files.

PR #3906, opened 2023-04-26
fix(sandy): Correct 100% full circle angle and seam allowance

BenJamesBen:
There was a behavior in Sandy where if a 100% circle is used and seam allowance is enabled, the skirt pattern is not the expected half-circle. Instead, the skirt falls short of a half-circle. This PR fixes this so 100% circle skirts have an exact half-circle pattern and there is no seam allowance along that 2nd edge (because it is on the fold).

Before the fix:
Screenshot 2023-04-26 at 2 07 58 PM

After the fix:
Screenshot 2023-04-26 at 2 05 24 PM

This has a side effect that the pattern jumps between 99.9% and 100.0%:

At 99.9%:
Screenshot 2023-04-26 at 2 02 45 PM

At 100.0%:
Screenshot 2023-04-26 at 2 02 29 PM

### DIFF
--- a/designs/sandy/src/skirt.mjs
+++ b/designs/sandy/src/skirt.mjs
@@ -76,7 +76,7 @@ function sandySkirt({
      * maximum angle calculated so the seam allowance fits in the
      * fabric
      */
-    if (an > 90 && sa) {
+    if (an > 90 && an !== 180 && sa) {
       const maxAn = utils.rad2deg(Math.atan(radiusWaist / sa))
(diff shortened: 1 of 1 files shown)


BenJamesBen:
Per discussion, the current behavior is desired. Closing.
https://discord.com/channels/698854858052075530/757631205804998759/1101159580597235915
discord://discord.com/channels/698854858052075530/757631205804998759/1101159580597235915

PR #3908, opened 2023-04-27
Add cutlists to blocks/slopers

BenJamesBen:
PR to add cutlist info to the block/sloper designs. It is important to do this first, before adding cutlists to other designs that inherit from them.

- I chose to add a minimal amount of `addCut()`s, only adding code where needed to produce the correct cutting layout. (The alternative I considered was adding explicit cutlist code to every part, even when not required, in the name of potential clarity.)
- Some parts have a `removeCut()` that was needed to clear an inherited cutlist. I elected to not add a comment explaining these situations.
- I also considered adding an explicit `removeCut()` to every inheriting part regardless of whether it was needed, but I chose not to do so.
- I added a missing `cutonfoldAndGrainline` to Breanna's front. The foldline and grainline are mentioned in the instructions and code, so it appears that the omission was accidental. 
- I added a matching `grainline` to Breanna's back, assuming that it was also intended (given that the front is aligned on grainline) but accidentally omitted.

Screenshot 2023-04-26 at 7 08 35 PM
Screenshot 2023-04-26 at 7 09 09 PM
Screenshot 2023-04-26 at 7 08 15 PM
Screenshot 2023-04-26 at 7 09 26 PM
Screenshot 2023-04-26 at 7 10 07 PM
Screenshot 2023-04-26 at 7 17 31 PM

### DIFF
--- a/designs/bella/src/back.mjs
+++ b/designs/bella/src/back.mjs
@@ -253,6 +253,11 @@ export const back = {
       .close()
       .hide()
 
+    macro('grainline', {
+      from: new Point(points.hps.x / 2, points.shoulder.y),
+      to: new Point(points.hps.x / 2, points.waistSide.y),
+    })
+
     store.cutlist.addCut()
 
     if (complete) {
@@ -261,10 +266,6 @@ export const back = {
         nr: 2,
         title: 'back',
         at: points.titleAnchor,
-      })
-      macro('grainline', {
-        from: new Point(points.hps.x / 2, points.shoulder.y),
-        to: new Point(points.hps.x / 2, points.waistSide.y),
       })
       macro('sprinkle', {
         snippet: 'bnotch',
--- a/designs/bella/src/front-side-dart.mjs
+++ b/designs/bella/src/front-side-dart.mjs
@@ -204,7 +204,7 @@ export const frontSideDart = {
       .curve(points.hpsCp2, points.cfNeckCp1, points.cfNeck)
       .hide()
 
-    store.cutlist.addCut()
+    store.cutlist.addCut({ cut: 1 })
 
     if (complete) {
       points.titleAnchor = new Point(points.armholePitch.x / 2, points.armholePitchCp2.y)
--- a/designs/bella/src/front-side-dart.mjs
+++ b/designs/bella/src/front-side-dart.mjs
@@ -204,6 +204,12 @@ export const frontSideDart = {
       .curve(points.hpsCp2, points.cfNeckCp1, points.cfNeck)
       .hide()
 
+    macro('cutonfold', {
+      from: points.cfNeck,
+      to: points.cfHem,
+      grainline: true,
+    })
+
     store.cutlist.addCut({ cut: 1 })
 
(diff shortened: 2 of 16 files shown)


joostdecock:
Do we have enough consensus on this to go ahead with this?
I seem to remember there was a discussion on this during the last call, and I'm not sure it was entirely resolved. (bad memory)

eriese:
Up to @BenJamesBen. If we end up deciding cutlists should not be inherited, nothing in this PR will break just because it becomes unnecessary.

It's just whether you want to move things out of `if (complete)` in the same PR or not

BenJamesBen:
I moved `cutonfold` and `grainline` macros out of the `if (complete)` blocks. The PR is ready to merge.

(Please note that moving the macros isn't always a simple operation. The change actually broke Noble when `complete = false` because the macros in Noble's back parts depended upon `points.grainlineFrom` and `points.grainlineTo` being inherited from Bella's back from a previous `grainline` macro. Previously it was guaranteed that the points existed, back when all the macros were in `if (complete)`. With the change, the points don't exist when `complete = false`. To fix Noble, I added code to create the points.)

(A best practice is to test patterns to verify they still work with both `complete = true` and `complete = false` after moving macros out of `if (complete)`.)

(When re-creating the grainline points for Noble, I took the opportunity to move the backoutside part grainline to a better location for a greater range of measurements.)

Before:
Screenshot 2023-04-27 at 2 55 10 PM

After:
Screenshot 2023-04-27 at 2 55 23 PM

woutervdub (approved):

eriese:
> I moved `cutonfold` and `grainline` macros out of the `if (complete)` blocks. The PR is ready to merge.
> 
> (Please note that moving the macros isn't always a simple operation. The change actually broke Noble when `complete = false` because the macros in Noble's back parts depended upon `points.grainlineFrom` and `points.grainlineTo` being inherited from Bella's back from a previous `grainline` macro. Previously it was guaranteed that the points existed, back when all the macros were in `if (complete)`. With the change, the points don't exist when `complete = false`. To fix Noble, I added code to create the points.)

Sounds like maybe we should be making the points and perhaps even the paths, but then hiding the paths if `complete === false`

> (A best practice is to test patterns to verify they still work with both `complete = true` and `complete = false` after moving macros out of `if (complete)`.)

good idea. We should add it to the standard tests

BenJamesBen:
> Sounds like maybe we should be making the points and perhaps even the paths, but then hiding the paths if complete === false

The documentation for `cutonfold` and `grainline` currently says that the paths and points are created when the macros are used. It does not mention `complete` as a condition.

We could either change the behavior of the macros as suggested above, or we could change the documentation to describe the current behavior. Since grainline and foldline information is something that we want to be inherited regardless of `complete` and regardless of whether the markings are drawn on the pattern, it sounds as if the first option should be done?

joostdecock:
I think it makes sense to let the macros take `complete` into account but developers should be able to force the to do it regardless

joostdecock (approved):

PR #3924, opened 2023-04-28
DRAFT: chore(hi): Add cutlist

BenJamesBen:
Draft PR, to ask questions about best practices for cutting instructions. This PR is for Hi, to add the cutting instructions listed here: https://freesewing.org/docs/patterns/hi/cutting/

What are the best practices for using custom materials in cutlists?
1. If there are no `fabric` pieces in the cutlist, switching to the Cutting Layout results in a blank screen, until one of the material tabs is selected. Is having `fabric` as one of the materials a best practice?
Screenshot 2023-04-27 at 5 53 32 PM

2. If I use a space character in the material name, it appears prefixed with "Plugin". Is the best practice to use words with no spaces? Possibly to assist with translations? Example when setting the material as `color 3`:
Screenshot 2023-04-27 at 5 56 00 PM

3. I'm not sure what determines the order of the material tabs. It might be beyond the scope of current functionality, but it might be nice to be able to specify an order so the most important material is listed first, etc.

### DIFF
--- a/designs/hi/src/aboveMouth.mjs
+++ b/designs/hi/src/aboveMouth.mjs
@@ -85,6 +85,8 @@ function draftHiAboveMouth({
   store.set('aboveMouthBottomLength', paths.bellyAndMouthAttachment.length())
   store.set('aboveMouthFinLength', points.aboveMouth02.dist(points.aboveMouth03))
 
+  store.cutlist.addCut({ material: 'color2Belly' })
+
   // Complete?
   if (complete) {
     points.aboveMouthSnippet = new Path()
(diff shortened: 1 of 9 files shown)


BenJamesBen:
4. Another variation of 1: If the design has a cutlist with only one material which is not `fabric`, the Cutting Layout is blank and there is no way to switch to the material.
Screenshot 2023-05-03 at 3 37 19 PM

Pinging @eriese . I need your guidance about how materials should be named in cutlists.

BenJamesBen:
As requested in today's Contributor Call meeting, pinging @eriese again to request guidance on cutlist material practices.

BenJamesBen:
Marking as ready for review. The issues raised are outside of the scope of this PR and don't require any changes to this PR.

(For now for Hi, the behavior will be that the Cutting Layout is blank when the user switches to it and until a material tab is selected.)

PR #3930, opened 2023-04-28
fix(benjamin): Part name changes, instructions, and cutlist

BenJamesBen:
- Change part names for consistency and improvement, to make the pattern, cutting instructions, and construction instructions consistent in the names they use ("Bow", "Short Bow", "Medium Bow", "Long Bow, and "Collar Band").
- Improve and clarify instructions.
- Fix cutlist instructions (2 Medium Bow/Knot 2 and only 1 Long Bow/Knot 3).

### DIFF
--- a/designs/benjamin/src/bow1.mjs
+++ b/designs/benjamin/src/bow1.mjs
@@ -10,6 +10,7 @@ function draftBenjaminBow1({
   sa,
   store,
   paperless,
+  options,
   part,
 }) {
   points.bandBottomLeft = points.bandBottomLeft.shift(0, 0)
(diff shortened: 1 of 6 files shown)


woutervdub (approved):

woutervdub:
I do wonder if we now should rename the `ribbon.mjs` to `collarBand.mjs` (and rename the functions within too). It seems inconsistent to not do so.

BenJamesBen:
I don't think that renaming the file or functions is necessary, though we could do it if we want. They aren't visible to customers using the design or reading its documentation.

BenJamesBen:
I moved this PR out of Draft status and changed its name.

SeaZeeZee:
@joostdecock this one's been reviewed and is ready to merge!

eriese:
While I'm emotionally with Wouter on renaming the files, it would also mean renaming the part exports, the parts, and their translation keys, so I think Ben's right and we skip it.

PR #3932, opened 2023-04-28
fix (lab) don't rely on browsers to compute transforms in layouts

eriese:
From [This discussion on discord](https://discord.com/channels/698854858052075530/698862765053575188/1101526717148971088) Firefox was not properly calculating transforms on the printing layout, which led to weird behaviors like allowing 6 pages for a 17-page pattern

The culprit was in `sites/shared/components/workbench/layout/draft/stack.mjs`

```js
// get the bounding box and the svg's current transform matrix
    const stackRect = innerRef.current.getBoundingClientRect()
    const matrix = innerRef.current.ownerSVGElement.getScreenCTM().inverse()

    // a function to convert dom space to svg space
    const domToSvg = (point) => {
      const { x, y } = DOMPointReadOnly.fromPoint(point).matrixTransform(matrix)
      return { x, y }
    }
```

This was returning different values in different browsers.

This PR uses our new matrix transform utils (ht @joostdecock) to avoid having to convert from DOM space to SVG space and return consistent transformations across browsers


BenJamesBen (approved):
I tested the PR in a localhost lab using the Safari and Chrome desktop browsers, and the behavior in the different layout views seems to be the same as in `develop`, with no problems seen.

I didn't test other browsers, including Firefox.

PR #3940, opened 2023-04-30
fix(wahid): Hide inadvertently exposed inherited options. Fixes #3005 

### LINKED ISSUE #3005: [bug]: (wahid) v3 options are different than v2

BenJamesBen:
In v2, only explicitly specified inherited options appeared in menus in the lab UI.
In v3, inherited options also inherit their menus and appear in the lab UI by default.

In Wahid, this exposed some options, causing them to appear in the lab UI. This PR adds overrides to hide them to match the previous behavior.

I tested the PR in the lab to verify that the same patterns are generated and the same options are seen between the PR and lab.freesewing.dev.

### DIFF
--- a/designs/wahid/src/front.mjs
+++ b/designs/wahid/src/front.mjs
@@ -22,6 +22,14 @@ import {
   lengthBonus,
   acrossBackFactor,
   frontArmholeDeeper,
+  bicepsEase,
+  collarEase,
+  cuffEase,
+  shoulderEase,
+  s3Collar,
+  s3Armhole,
+  shoulderSlopeReduction,
+  backNeckCutout,
 } from './options.mjs'
 
 function wahidFront({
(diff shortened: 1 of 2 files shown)


bobgeorgethe3rd:
#3005
 So appears on issue github doesn't do it with titles for some reason

joostdecock (approved):

PR #3942, opened 2023-04-30
fix(shared): Silence no-menu warnings for constant options

BenJamesBen:
This PR eliminates `console.log()` warnings for constant options. The current behavior is to give a warning when there is no menu configured for the option.

Example, for Benjamin:
```
utils.mjs:192 Warning: Option adjustmentRibbonWidth does not have a menu config. Either configure it, or set it to false to hide this option.
utils.mjs:192 Warning: Option bandLength does not have a menu config. Either configure it, or set it to false to hide this option.
utils.mjs:192 Warning: Option transitionLength does not have a menu config. Either configure it, or set it to false to hide this option.
```
Resulting from these contants in `designs/benjamin/src/base.mjs`:
```js
 options: {
    // Static options
    transitionLength: 2, //Twice the knot
    bandLength: 0.17,
    adjustmentRibbonWidth: 20,
...
```

This PR takes advantage of the the fact that constant options have only one key (`option.name`). From testing it was seen that all other options configured as objects have at least 2 keys.

### DIFF
--- a/sites/shared/utils.mjs
+++ b/sites/shared/utils.mjs
@@ -187,8 +187,9 @@ export const optionsMenuStructure = (options) => {
   // Fixme: One day we should sort this based on the translation
   for (const option of orderBy(sorted, ['menu', 'name'], ['asc'])) {
     if (typeof option === 'object') {
+      const isAConstant = Object.keys(option).length === 1
       if (option.menu) set(menu, `${option.menu}.${option.name}`, optionType(option))
-      else if (typeof option.menu === 'undefined') {
+      else if (typeof option.menu === 'undefined' && !isAConstant) {
         console.log(
           `Warning: Option ${option.name} does not have a menu config. ` +
             'Either configure it, or set it to false to hide this option.'


BenJamesBen:
This fix also eliminates similar messages spuriously generated for legitimate options when the ClearThing button is used.

sites/shared/utils.mjs
```diff
@@ -187,8 +187,9 @@ export const optionsMenuStructure = (options) => {
   // Fixme: One day we should sort this based on the translation
   for (const option of orderBy(sorted, ['menu', 'name'], ['asc'])) {
     if (typeof option === 'object') {
+      const isAConstant = Object.keys(option).length === 1
       if (option.menu) set(menu, `${option.menu}.${option.name}`, optionType(option))
-      else if (typeof option.menu === 'undefined') {
+      else if (typeof option.menu === 'undefined' && !isAConstant) {
```
eriese, line 192: I think this check, and actually also the menu check, would work better in the previous loop on line 182. We're manipulating the options there anyway, may as well only include ones that should actually go in the menu. That way instead of counting keys, you can do a type check on the options. Then this loop will just do the business of line 192

eriese (changes requested):

sites/shared/utils.mjs
joostdecock: @BenJamesBen please change line 183 to:

```js
if (typeof option === 'object') sorted[name] = { ...option, name }
```

The current behaviour is a bug, and this proposed change treats the symptom, rather than the cause.

Static options are a scalar, and a scalar can't be spread (`...5` is not valid JS).
The reason it doesn't throw an error is because it's in an object literal (`{ ...option}`). However, if you were to use `console.log(...option)` it would throw an error as, once again, you can't spread a scalar.

Because it's in the object literal, this error is not thrown, but it's still a bug and that's what should be addressed.

So, if you feel up for it, you could revert this change and instead update line 183 as shown above.

joostdecock (changes requested):

BenJamesBen:
Changes made as requested, and re-tested in the lab.

eriese (approved):

PR #3943, opened 2023-04-30
fix(workbench): Do not pre-draft patterns without measurements

BenJamesBen:
In the browser Javascript console I see messages that indicate that the lab is drafting patterns without measurements. This happens in certain circumstances, like when the ClearThing button is used or when visiting the Logs view immediately after browsing to a new design.

For example, if I go to Carlton and use the ClearThing button I see messages like `Warning: The round macro only handles 90 degree angles correctly.` and `Called part.setCutOnFold() but at least one parameter is not a Point instance`. And also things like, `Unable to draft part 'bent.sleeve' (set 0) TypeError: Cannot read properties of undefined (reading 'x')`.

### DIFF
--- a/sites/shared/components/wrappers/workbench.mjs
+++ b/sites/shared/components/wrappers/workbench.mjs
@@ -139,7 +139,7 @@ export const WorkbenchWrapper = ({
 
     // draft it for draft and event views. Other views may add plugins, etc and we don't want to draft twice
     try {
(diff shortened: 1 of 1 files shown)


eriese (approved):
I have some thoughts about changing how we handle the per-view drafting, but this is a good fix for the current way

PR #3974, opened 2023-05-04
fix(bob,penelope,docs): Access optional measurements via `measurements`

### LINKED ISSUE #3933: [bug]: (bob,penelope): Optional measurements are accessed via `measurements`, not `optionalMeasurements`

Steps to Reproduce:
1. In lab.freesewing.dev, select the Bob design.
2. Enable `paperless`, and note that the neck opening width is 10.08 cm.
3. In the Measurements view select With Breasts Size 28, and note that the Head circumference is set to 53 cm.
4. Switch back to DraftDesign view, and note that the neck opening width is still 10.08 cm.
5. In the Measurements view select Without Breasts Size 50, and note that the Head circumference is set to 65.5 cm.
6. Switch back to the DraftDesign view, and note that the neck opening width is still 10.08 cm.

The same behavior is seen in the `develop` localhost lab.

_No response_

No, I am not 😞

### Additional context

I note that the below code addition seems to fix the issue, though I don't fully understand the code so I don't know if it is an appropriate fix.

In `packages/core/src/part.mjs`:
``` diff
@@ -175,6 +175,12 @@ Part.prototype.shorthand = function () {
     },
     set: (measurements, name, value) => (self.context.settings.measurements[name] = value),
   })
+  shorthand.optionalMeasurements = new Proxy(this.context.settings.measurements, {
+    get: function (optionalMeasurements, name) {
+      return Reflect.get(...arguments)
+    },
+    set: (optionalMeasurements, name, value) => (self.context.settings.optionalMeasurements[name] = value),
+  })
   shorthand.options = new Proxy(this.context.settings.options, {
     get: function (options, name) {
       if (typeof options[name] === 'undefined')
```

BenJamesBen:
Fixes #3933.

### DIFF
--- a/markdown/dev/reference/api/part/config/measurements/en.md
+++ b/markdown/dev/reference/api/part/config/measurements/en.md
@@ -38,3 +38,12 @@ const part = {
 }
 ```
 
+<Tip>
+
+Although they are specified via the part configuration `optionalMeasurements`
+property, optional measurements are accessed via the 'measurements'
+settings property.
+
+(There is no `optionalMeasurements` settings property.)`
+
+</Tip>
--- a/designs/bob/src/bib.mjs
+++ b/designs/bob/src/bib.mjs
@@ -16,7 +16,7 @@ export const bib = {
     points,
     Path,
     paths,
-    optionalMeasurements,
+    measurements,
     options,
     macro,
     log,
@@ -27,7 +27,7 @@ export const bib = {
     part,
   }) => {
     // Head size
-    const head = (optionalMeasurements?.head || 360) * options.headSize
+    const head = (`head` in measurements ? measurements.head : 360) * options.headSize
 
     // Construct the neck opening
     let tweak = 1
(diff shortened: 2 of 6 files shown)


woutervdub:
I checked the functionality, and all seems right.

woutervdub (approved):

SeaZeeZee:
@joostdecock same as previous, wouter reviewed this so its ready to go!

markdown/dev/reference/api/part/config/measurements/en.md
```diff
@@ -38,3 +38,12 @@ const part = {
...
 
+<Tip>
+
+Although they are specified via the part configuration `optionalMeasurements`
+property, optional measurements are accessed via the 'measurements'
+settings property.
+
+(There is no `optionalMeasurements` settings property.)`
+
+</Tip>
```
eriese, line 49: Sorry to nitpick, but I don't think this goes here. It's not relevant to a user who has come to the docs to figure out how to specify which measurements are required by a part, and I worry it's confusing to see in this context

I think it should move to `markdown/dev/api/part/draft/en.md`, where it is relevant to a user who is looking at how to access measurements in the draft method and there is currently no note

eriese (changes requested):

Issue #3232, opened 2022-12-14
[docs]: Settings documentation missing for absoluteOptions, sample, and stackPrefix


BenJamesBen:
### Where can we see the problem? 🤔

https://freesewing.dev/reference/settings

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Suggested content

Settings documentation missing for `absoluteOptions`, `debug`, `sample`, and `stackPrefix` and should be added.

- `absoluteOptions` is not something that users or designs will set. The infrastructure sets it automatically when appropriate. However, it is something that patterns can access and should be documented.
- `debug` exists in `packages/core/src/config.mjs`, where `debug: false,` is set as the default value. As far as I can determine it is not used anywhere in the FreeSewing core or web page infrastructures. And, I am not sure if patterns are able to access the setting when drafting parts.
- `sample` is used by `Pattern.sample()` and should be documented.
- `stackPrefix` is new and seems to be used only within the infrastructure. We might need to wait for stacks/sets to be documented before documenting `stackPrefix`.

BenJamesBen:
Per Joost, `debug` was left over from v2 and should be removed. So, it is no longer a part of this bug.

BenJamesBen:
Just to provide an update on this bug. Once PR #4009 has been merged, this bug can be closed.

BenJamesBen:
PR #4009 was merged on Oct 14, 2023. Closing.

PR #3978, opened 2023-05-04
chore(docs): Move sample setting docs from Pattern.sample() to its own doc

BenJamesBen:
(The `sample` setting was previously undocumented in the reference Settings section.)


SeaZeeZee (approved):
*thumbs up*

BenJamesBen:
This partially covers #3232.

SeaZeeZee:
@joostdecock I have also looked at and approved this PR so it can be merged

PR #3989, opened 2023-05-04
enhancement(title): Add alignment property

### LINKED ISSUE #2745: [feature]: Add text alignment functionality to title macro/plugin

An enhancement request to add alignment functionality to the `title` macro/plugin. For example, something like:
```
     macro('title', {
       nr: 1,
       at: points.title,
       title: 'Front',
       align: 'center',
     })
```

No, I am not 😞

### Additional context

I notice that a number of patterns are using the following hack to achieve this functionality:
```
     points.__titleNr.attr('data-text-class', 'center')
     points.__titleName.attr('data-text-class', 'center')
     points.__titlePattern.attr('data-text-class', 'center')
```

However, the hack is somewhat tedious, and when the title macro changes (like when it did when the export date was added), the hack results in mixed alignment and a not-as-good-looking title.

BenJamesBen:
Fixes #2745.

Although this is something we could implement, I'm not sure whether we want to do so. However, because a number of Designs (Cornelius, Hortensia, Lunetius, Noble, Tiberius, Walburga) already include workaround code to center align titles, I'm submitting this PR to see how we feel about it.

### DIFF
--- a/markdown/dev/reference/macros/title/en.md
+++ b/markdown/dev/reference/macros/title/en.md
@@ -9,6 +9,7 @@ It is provided by the [annotations plugin](/reference/plugins/annotations).
 
 ```js
 macro('title', {
+  String align,
   Boolean append,
   Point at,
   Boolean cutlist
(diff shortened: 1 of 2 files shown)


BenJamesBen:
Not quite ready yet... Modifying and adding tests...

PR #3998, opened 2023-05-05
[proposal]: Keep macro ID in store for removal

joostdecock:
As agreed in the most recent contributor call, I'd put up a proposal to outline how we could handle removing changes made by macros.

This PR has a minimal implementation of this, which boils down to:

- A macro passes an `id` parameter (with a default, so it's optional)
- The plugin stores this id in the store, along with any info it needs to roll back the changes
- The `rm` prefixed macro exists that rolls back these changes after looking up the info in the store

I would also suggest we strive to harmonize the way data is passed to plugins:

- Use `id` everywhere (no more prefix)
- Allow passing objects or strings refering objects (paths or points). Strings are preferred because they point to named Point/Path objects which makes it easier to find and remove them afterwards, but we can support both
- add `rmmacroname` instead of the less-intuitive way of passing `false` to the macro

Marking as DRAFT because this is not supposed to be merged as-is, it's for discussion.

### DIFF
--- a/designs/plugintest/src/plugin-annotations.mjs
+++ b/designs/plugintest/src/plugin-annotations.mjs
@@ -34,12 +34,15 @@ const pluginAnnotations = ({
     points.banner_to = new Point(320, y)
     paths.banner = new Path().move(points.banner_from).line(points.banner_to)
     macro('banner', {
-      path: paths.banner,
+      path: 'banner',
       text: 'banner macro',
       dy: options.bannerDy,
       spaces: options.bannerSpaces,
       repeat: options.bannerRepeat,
     })
+    macro('rmbanner', 'banner')
+    paths.banner.addText('banner removed')
+
     macro('bannerbox', {
       topLeft: points.banner_from,
       bottomRight: points.banner_to,
--- a/plugins/plugin-annotations/src/banner.mjs
+++ b/plugins/plugin-annotations/src/banner.mjs
@@ -1,8 +1,9 @@
 // Export macros
 export const bannerMacros = {
-  banner: function (so) {
+  banner: function (so, pattern) {
     // Mix defaults with settings object
     so = {
+      id: 'banner',
       text: '',
       dy: -1,
       spaces: 12,
@@ -10,14 +11,29 @@ export const bannerMacros = {
       className: '',
       ...so,
     }
-    so.path.attr('data-text-dy', so.dy).attr('data-text-class', `${so.className} center`)
-    const spacer = '&#160;'.repeat(so.spaces)
+    // Allow passing of path by ID or the actual object
+    const path = typeof so.path === 'object' ? so.path : pattern.paths[so.path]
 
+    // Store id/path in store for later removal with rmbanner
+    if (path.name) pattern.store.set(`data.plugins.annotations.banner.${so.id}`, path.name || path)
+
+    // Set test dy value
+    path.attr('data-text-dy', so.dy).attr('data-text-class', `${so.className} center`)
+
+    // Set text on path
+    const spacer = '&#160;'.repeat(so.spaces)
     for (let i = 0; i < so.repeat; i++) {
-      so.path.attr('data-text', spacer)
-      so.path.attr('data-text', so.text)
+      path.attr('data-text', spacer)
+      path.attr('data-text', so.text)
     }
 
-    so.path.attr('data-text', spacer)
+    path.attr('data-text', spacer)
+  },
+  rmbanner: function (id = 'banner', pattern) {
+    let path = pattern.store.get(`data.plugins.annotations.banner[${id}]`)
+    if (typeof path !== 'object') path = pattern.paths[path]
+
+    // Remove attributes from path
+    path.attributes.remove('data-text-dy').remove('data-text-class').remove('data-text')
   },
 }


woutervdub:
>> A macro passes an id parameter (with a default, so it's optional)

Should the macro return the `id`? So you can do:
```
const mirrorId =   macro('mirror', {clone: true,mirror: [ points.m1, points.m2 ],paths: Object.keys(paths),})
if( weDidntExpectThis ) { rmmirror( mirrorId ) }
```

BenJamesBen:
The concept sounds good. Some thoughts:
- Macros would need to store the actual Path/Point objects if they are unnamed/anonymous.
- Macros would need to check for and handle situations where an `id` gets reused, if simply overwriting points/paths isn't sufficient.

eriese:
I'm for it, I'm just stuck on `rmmacroname`. maybe it could be `macro('macroname', {remove: true})? That way you could still pass other data, but it doesn't have to be a whole separate macro. 

I'm also with Wouter that it couldn't hurt to return the applied macro id.

woutervdub:
> I'm for it, I'm just stuck on `rmmacroname`. maybe it could be `macro('macroname', {remove: true})? 

We could even make that `macro( 'macroname', {remove: <macroId>})

woutervdub:
@joostdecock Do we still want to do this in v3?

joostdecock:
I'd say yes

joostdecock:
@woutervdub to clarify where I'd like to land on this: This idea of passing an ID to already partially implemented. In particular the current `plugin-annotations` already takes an ID for the various dimensions.

In a design I am working on currently, this looks like this:

```js
    macro('hd', {
      id: 'belLoopLeftFront',
      from: points.leftFrontTop,
      to: points.leftFrontLoopTop,
      y: points.cbTop.y - 30 - sa,
    })
 ```
 
 This is a godsend when trying to remove specific dimensions from an inherited pattern, and we should try to apply it consistently to all macros.
Things that are still to be done are:

- Add a macro-specific hard-coded prefix. The way it is implemented now, if you have `paths.seam` and then you pass `seam` as an ID in this macro, your path will be overwritten. We should make it clear these paths are auto-generated, perhaps with a `__macroname_` prefix or something.
- Add these IDs to the store. This will allow people to iterate over them. They should go into their own little corner, like `macros.name.ids` or something.
- Add an `rm` alternative for each macro so they can easily be rolled back.
- Macros should return the ID (good idea)
- Removal should be through `rmblah(id)`

Overloading the macro properties with removal instructions will cause cognitive overhead and complicate documenation.
Each plugin that provides a macro should also provide the macro to undo said macro, with the `rm` prefix.
Once that's established, it will be easy to remember and also a lot easier to find where a macro is actually used (rather than used to remove something).

While we're mucking about with macros, it might also be more elegant from an API point of view to use:

```js
macros.blah({ options })
```

Rather than:

```js
macro('blah', { options })
```

Which is also more consistent with how we deal with points/paths/snippets.

woutervdub:
> While we're mucking about with macros, it might also be more elegant from an API point of view to use:
> 
> ```js
> macros.blah({ options })
> ```

If we're going that route, wouldn't it be more consistent to do something like this:

```js
macros.dimLeft2Right = new Macro( 'hd', {from: points.left, to: points.right})
```

And then to remove do something like:

```js
macros.dimLeft2Right.hide()
```

woutervdub:
@joostdecock 

> This is a godsend when trying to remove specific dimensions from an inherited pattern, and we should try to apply it consistently to all macros.

The dimension macros only have one removal in the current code. Shouldn't we have removal macros for each dimension macro?

Another thought; why would we not have something like `rmmacro( 'hd', hdMacroId )`?

woutervdub:
By this:

> Add these IDs to the store. This will allow people to iterate over them. They should go into their own little corner, like macros.name.ids or something.

you mean something like `macros.hd.ids` and `macros.bartack.ids`?

woutervdub:
@joostdecock 

I hate to take you away from other things you're working on, but could you give me some direction here? I'm not sure exactly where you want to go with this. Can you comment on the questions above?

@eriese, @BenJamesBen, your insights are appreciated too.

joostdecock:
@woutervdub no worries, good move to remind me.

> *If we're going that route, wouldn't it be more consistent to do something like this:*

No, that would pretend macros are like points/paths, but they are rather different.
Let's not worry about that right now and keep calling `macro('name', { })`. It's not all that important

> *The dimension macros only have one removal in the current code. Shouldn't we have removal macros for each dimension macro?*

Good point. Let's do that. Although I still think the `rmad` macro is useful, and perhaps it could take a list of IDs? (to remove a bunch of them in 1 call.

> *you mean something like `macros.hd.ids` and `macros.bartack.ids`?*

Yes, exactly like that :)


Let me know if I overlooked something

woutervdub:
@joostdecock 

Some more questions/remarks.

Some current macros already have a remove option (`macro('cutonfold', false)`). Do we want to maintain this behaviour, or replace those occurrences with the new `rmcutonfold(macroID)`?

I could write a `rmmacro(macroID)` that just removes all paths and points of a certain macro, removing the need to have individual removal methods.

woutervdub:
> I could write a `rmmacro(macroID)` that just removes all paths and points of a certain macro, removing the need to have individual removal methods.

That would not work for all macros. Or, at least, may be very cumbersome for some macros like `banner`.

joostdecock:
@woutervdub 

> *Some current macros already have a remove option (macro('cutonfold', false)). Do we want to maintain this behaviour, or replace those occurrences with the new rmcutonfold(macroID)?*

I'd say replace it so its consistent

> *I could write a rmmacro(macroID) that just removes all paths and points of a certain macro, removing the need to have individual removal methods.*

That would assume you know what all macros are doing. But users may add their own macros, so I don't think this could work in practice.

woutervdub:
@joostdecock There are macros with a _prefix_ that somewhat functions as an _id_. Shall we get rid of that, in favour of a consistent use of _id_?

joostdecock:
@woutervdub yeah, I think it's best to have consistency

woutervdub:
@joostdecock 

For the removal part, using the same `macro` method is somewhat awkward. 

When a macro is generated, there may or may not be an `id` parameter. If there is, the `id` needs to get a prefix of the macro that is being generated (ie: `__macro_grainline_7`) so as not to overwrite a different macro with the same `id`. One could have a `banner` macro and a `hd` macro with the same `id` 'topHem'. And if there isn't an `id` parameter, one will be created. And this `id` will be saved in the `store`.

With the removal/reversal of a macro, the `id` is required. Otherwise, there is no way to know what needs to be removed. So the macro method would have to test every time when it is called if the macro name starts with 'rm', to understand if it needs to handle a removal or generation. With a removal, no `id` is stored, but instead is retrieved. No prefix is added to the `id`.

It would be simpler to have a `rmmacro` method that would do the removal. And it would allow for things like `rmmacro('hd',store.get('macros.hd.ids'))` to remove all `hd` macros at once.

Thoughts?

woutervdub:
> ```js
> macro('rmd', 'hemWidth')
> ```
I'm afraid that `rmd` has been scrapped in favour of `rmld`, `rmpd`, `rmhd`, and `rmvd`. All in the name of consistency. 

I guess I could include a non-`rm` macro that would remove all the dimension macros, but doing a `rmd` is not really all that easy, because (tl;dr) of consistency. 

(the longer version is that every macro with an rm prefix is regarded to be a removal macro of the corresponding non-prefix macro. And the `id` for the to-be-removed macro needs to be in the `macros.<macroName>.ids` store. And there are none in the `macro.d.ids` path, since there is no `d` macro.)

woutervdub:
@joostdecock @eriese @BenJamesBen 

Summarising where I am with this. 
- Every macro is assigned an `id` 
- This `id` is stored in `store.macros.<macroName>.ids`
- This `id` is returned by `macro`.
- Calling an unknown macro will return `undefined`
- Macros can be called with their own `id` (`macro('cutOnFold',{id:'top'})`). If none is present on the call, one will be assigned. 
- The `id` is prefaced by __macro_<macroName>_ before being passed to the macro code.
- All of this is done in `part.mjs`, and does not have to be handled in the macro itself.
- Macros can have remove versions. These should have the same name as the original macro, but be prefaced with 'rm'.
- This means that there can be no 'normal' macros that start with 'rm'.
- Macros do not need to have a 'rm' version (it is not enforced through tests or something).
- Remove macros should remove the instance that corresponds with the `id` they are passed.
- The call to `macro('rm<macroName>`)` can have an individual `id` of the macro to be removed. Or an array of `id`. Or nothing at all (in which case all macros of that type will be removed). The removal macro will be called multiple times with successive `id`s. This is all handled by `part.mjs`
- Some macros may use `store.macros.<macroName>` to store additional information that may be useful for removal purposes.

I'm afraid that the `rmd` and `rmad` macros have been dropped. These have been replaced with `rmhd`, `rmld`, 'rmpd` and 'rmvd'

- The `cutOnFold` macro has been adapted to work with the cutting layout system. The `id` of the macro is used to keep thing in line.

Thoughts? Concerns?

joostdecock:
I love the work on this. I did not even expect all this auto-if stuff but nice work.
(The reason I didn't expect is is that I think it makes most sense to use an explicit ID. That way people who extend your pattern can undo that macro. Not sure how that would work with a runtime assigned ID, but as long as passing in an ID works as expected, I'm happy 😄 )

woutervdub:
I'm removing this:
- The id is prefaced by __macro_<macroName>_ before being passed to the macro code.

And replacing it with adding a macro specific prefix to all paths and points inside the macro itself. That way it is easier to do this:
```
macro('vd',{from:f, to:t, id: 'dim')

macro('rmvd','dim')
```

So each macro is now responsible for adding a prefix to things. And it can choose its own prefix.

eriese:
Sounding good to me, as long as test make sure that inheritance all works nicely so that you can add a macro in a parent part and remove it in a child part, etc.

woutervdub:
@joostdecock 

Instead of storing the macro ids in the `store`, can we store them in the `part`, like points, paths, and snippets? I'm running into some issues with the `store` route. First of, one part could (try to) remove macros that were created in another part. But more so, the `getId()` returns an unique id for this part. So you can have multiple macros with the same id in the same pattern, just belonging to different parts. 

I'm not sure if there is a practical reason for one part to be able to refer to, or manipulate, a macro that was created in a different part, unless that part is in it's inheritance chain (the same as with points and paths).  So it would make more sense to keep them local and all the info stored in the part, not the store. Adding some reference to the store (ie. `macros.<part>.<macroName>.ids` would not work, since you would not know which part you need to reference to in an inherited part.

joostdecock:
@woutervdub can't you just add the oart name to the ID?

woutervdub:
I could. But then I can’t do something like `macro(‘rmtitle’)` to remove all title macros in parts that this part is based on.

joostdecock:
Perhaps allow the `rm...` maceo to be explicit about the part?

```js
// Assuming `tip` is the ID

// Remove top hd macro in this part 
macro('rmhd', 'top')

// Remove top hd macro in brian:front part 
macro('rmhd', 'top', 'brian:front')
```

Thoughts?

woutervdub:
I feel like I've taken this as far as it should be. It allows for macros to be added with specified or automatically generated ids. And it allows for those to be removed, by id, or all of a certain type. This last part works for the part, and all parts it is based on. This works fine, even with shared base parts. And the system plays well now with the plugin-cut-layout. I tested it with multiple cutOnFold macros, defined in separate parts, and it behaves as you would expect.

The designs in this Macro-ID branch have been adjusted to the new implementation.

Anyone want to take a view of things before I start adjusting the documentation?

Thanks!
@joostdecock @eriese @BenJamesBen

woutervdub:
Oh, I probably should mention this:
`macro()` has an optional parameter: `id`. If omitted, a part-unique value will be generated ('1', '2', etc.)
A `part` now has a `macros` property. When calling a macro, `part` adds the macro name and `ids` to it. The `ids` is an array with all the individual id values that were either specified, or generated when none were specified. Example: `macros.hd.ids['2','top','7']`
All annotation macros now have a `rm` counterpart that removed the annotation the macro made. `rmhd` to remove the `hd` macro, `rmtitle` to remove the `title`. The `false` parameter that some macros allowed to remove the annotation has been eliminated. The `macro()` call for a 'rm' version takes an optional parameter; the `id` of the macro to be removed, or an array of ids to be removed. When the parameter is omitted, all macros of that type defined up to this moment will be removed. Including those that have been defined in parts that were included into this part. 
All annotation macros can now have multiple version of themselves in the same part (two titles, two scaleboxes, etc). 
The `rmad` macro has been removed. A `removedim` macro has been provided. The latter just calls the `rmhd`, `rmld`, `rmpd`, and `rmvd` in turn without any parameters, removing all dimension macros.

joostdecock:
Thanks @woutervdub for your valuable work on this.
I've just wrapped up my work on the site, and I'm running low on fuel right now, but I'll pick this up over the course of next week.

eriese:
Hey @woutervdub, where can we see your work on this?

woutervdub:
> Hey @woutervdub, where can we see your work on this?

@eriese In the Macro-ID branch:
https://github.com/freesewing/freesewing/tree/Macro-ID

joostdecock:
Closing in favor of #4912

Issue #3902, opened 2023-04-26
Add scale support for text in React render


joostdecock:
### What seems to be the problem? 🤔

As pointed out by @woutervdub the React view (aka the website) only partially supports the wcale setting, with text not adapted. 
Thus is unlike the SVG output which uses CSS that takes the scale into account.

Similar CSS should be generated in the fly in the frontend, taking the scale setting into account.

### Are you a FreeSewing patron? 😃

Yes, I am a tier-8 patron ❤️  💙 💜

bobgeorgethe3rd:
Implemented as of #3999

Issue #3579, opened 2023-02-27
[bug]: Some Snippets do not obey pattern scaling


BenJamesBen:
### What seems to be the problem? 🤔

Some Snippets like buttons, buttonholes, notches, bnotches, and bartacks do not change size when the overall pattern scale setting is changed. Others like the logo do scale.

Steps to Reproduce:
1. In the localhost lab development environment, using the Simon pattern, select cisFemale Doll 10% size, and go to the DraftDesign view.
3. Observe that some graphical elements are large.
4. In the Settings menu on the right, change the Scale setting to 0.1.
5. Observe that most graphical elements get smaller.
6. However, also observe that the notches, bnotch, buttons, and buttonholes did not get smaller.
7.  Move the Scale setting slider back and forth, and observe that while other elements shrink and grow, the buttons, buttonholes, etc. remain the same size.

### Are you a FreeSewing patron? 😃

No, I am not 😞

BenJamesBen:
The logo is a special snippet in that it scales itself. Other snippets like buttonholes don't scale themselves and instead depend on the core infrastructure to scale them.

A possible fix for the issue in exported SVG/PDFs is:
In `packages/core/src/svg.mjs`:
```diff
@@ -322,7 +322,7 @@ Svg.prototype.__renderSnippet = function (snippet) {
   let x = round(snippet.anchor.x)
   let y = round(snippet.anchor.y)
   let scale = snippet.attributes.get('data-scale') || 1
-  scale = scale * (this.pattern.settings.scale || 1)
+  scale = scale * (this.pattern.settings[0].scale || 1)
   if (scale) {
     snippet.attributes.add('transform', `translate(${x}, ${y})`)
     snippet.attributes.add('transform', `scale(${scale})`)
```
A possible fix for the issue in the lab UI is:
In `sites/shared/components/workbench/draft/snippet.mjs`:
```diff
@@ -51,7 +51,8 @@ export const Snippet = (props) => {
     x: props.snippet.anchor.x,
     y: props.snippet.anchor.y,
   }
-  let scale = props.snippet.attributes.get('data-scale')
+  let scale = ( props.snippet.attributes.get('data-scale') || 1)
+  scale = scale * (props.gist.scale || 1)
   let rotate = props.snippet.attributes.get('data-rotate')
   if (scale || rotate) {
     snippetProps.transform = ''
```

BenJamesBen:
[...]

PR #3999, opened 2023-05-05
Implements #3579 & #3902

woutervdub:


bobgeorgethe3rd:
#3579,  #3902

BenJamesBen:
I tested the PR in the lab, using the RenderTest and PluginTest designs. I also exported SVGs and PDFs. I believe that the PR works and is okay to merge.

It might be outside the scope of this PR, but I saw the below unexpected behavior, and I was wondering if the behavior is correct?

When working in the lab, on the screen text would scale down and become smaller, as expected. However, when I exported SVG and PDF files, I saw some text remain the same size regardless of scale. This is most evident on `banner` text, the `cutonfold` and `grainline` indicators and dimension measurements. Other text, like in `title` did scale down.

SVG scale: 1.0:
Screenshot 2023-05-07 at 3 01 17 PM

SVG scale 0.1:
Screenshot 2023-05-07 at 3 00 56 PM

plugins/plugin-annotations/src/scalebox.mjs
```diff
@@ -148,9 +156,10 @@ export const scaleboxMacros = {
       .line(points.__scaleboxMetricTopRight)
       .close()
     // Lead
-    points.__scaleboxLead = points.__scaleboxLead
+    points.__scaleboxLead
       .attr('data-text', so.lead || 'FreeSewing')
       .attr('data-text-class', 'text-sm')
+      .attr('data-text-transform', transform(points.__scaleboxLead))
```
eriese, line 162: Given that font size should be scaling automatically now, are these transforms necessary?

eriese (changes requested):
good work! Just one question and a little cleanup

plugins/plugin-annotations/src/scalebox.mjs
woutervdub: They were the last time I tested. I'll try to get this branch in line with `develop` and have a look.

eriese:
Shouldn't matter about `develop`. The text scaling is implemented in this
branch

plugins/plugin-annotations/src/scalebox.mjs
woutervdub: Tested the branch, transform is needed.

eriese:
please fix the failing test on cross-box, and for completeness please fix the bug Ben pointed out (take a look at line 33 in `plugins/plugin-theme/index.mjs`)

plugins/plugin-annotations/src/scalebox.mjs
eriese: Help me understand? Is this about visual space?

With additional transforms at scale 0.2 and scale 0.5
Screen Shot 2023-05-17 at 4 04 18 PM
Screen Shot 2023-05-17 at 4 05 26 PM

Without additional transforms at scale 0.2 and scale 0.5
Screen Shot 2023-05-17 at 4 04 35 PM
Screen Shot 2023-05-17 at 4 05 05 PM

woutervdub: You're right, and I'm an idiot. Transforms removed.

woutervdub:
> I tested the PR in the lab, using the RenderTest and PluginTest designs. I also exported SVGs and PDFs. I believe that the PR works and is okay to merge.
> 
> It might be outside the scope of this PR, but I saw the below unexpected behavior, and I was wondering if the behavior is correct?
> 
> When working in the lab, on the screen text would scale down and become smaller, as expected. However, when I exported SVG and PDF files, I saw some text remain the same size regardless of scale. This is most evident on `banner` text, the `cutonfold` and `grainline` indicators and dimension measurements. Other text, like in `title` did scale down.
> 
> SVG scale: 1.0: Screenshot 2023-05-07 at 3 01 17 PM
> 
> SVG scale 0.1: Screenshot 2023-05-07 at 3 00 56 PM

The issue is with how the svg is created. The `transform` properties work differently when applied to the <text> and the <tspan> tags. Some text uses a point (title) and that results in the `data-text-transform` attribute to be set on the <text> tag. When the text is part of a path, that same attribute goes to the <tspan> tag.

plugins/plugin-annotations/src/scalebox.mjs
eriese: Not an idiot at all. I assume that you fixed the scalebox before fixing the font size and that's how this happened

BenJamesBen:
After restoring `sites/shared/components/workbench/menu/core-settings/index.mjs` in my local repo, I re-tested this PR by scaling Plugintest. I verified that the text in scale boxes seems to scale properly, and that the text in SVG files also scale. I also verified that `plugin-annotations` passes its tests.

BenJamesBen (changes requested):

sites/shared/components/workbench/menu/core-settings/index.mjs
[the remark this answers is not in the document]
woutervdub: I removed it because it caused a merge problem (and the change is not really needed for this PR). But I must have done something wrong. I only intended to remove it from the PR.

eriese:
Sorry @woutervdub, we're so close here. There's now a broken test in `plugin-theme`. 

And to remove your changes to `sites/shared/components/workbench/menu/core-settings/index.mjs`, you'll run `git checkout $(git merge-base develop HEAD) sites/shared/components/workbench/menus/core-settings/index.mjs`

woutervdub:
> The issue is actually in the [theme plugin, on line 33](https://github.com/freesewing/freesewing/blob/7cd9d43a23259ce4935814c5dadcd8dc308852c1/plugins/plugin-theme/src/index.mjs#L33). It is trying to use `pattern.settings.scale` to set scaled values, but `pattern.settings` is an Array of settings objects, so it needs to be changed to `pattern.settings[0].scale`. Please make that change, and then we'll be ready to go.

This breaks the test in plugin-theme

eriese:
> This breaks the test in plugin-theme

Yes, but it fixes the bug, so the test now needs to be fixed as well. The problem with the test is that it's also assuming an object for `settings` rather than and array of objects

eriese (approved):

PR #4010, opened 2023-05-07
fix(docs): (cutonfold,grainline) Update docs to reflect new behavior

BenJamesBen:


BenJamesBen:
It occurs to me that perhaps all of our macros are being changed to automatically show/hide themselves based on `complete`? If so, then I am not sure whether the `<Note>` I added might be redundant? Perhaps the behavior would be best documented in one single place elsewhere, perhaps in the documentation for `complete`?

(My understanding is that the behavior change hasn't yet been applied to all macros.)

SeaZeeZee (approved):
This looks good to me but I can't answer your question Benjamin I hope someone else can!

BenJamesBen:
I changed the PR status from Draft to ready. (If the `<Note>` is redundant and needs to be removed or if the information needs to be added to the documentation in a different place, please let me know.)

joostdecock:
I'm going to keep this open for a while until we clarify how we handle macros

joostdecock (approved):

PR #4013, opened 2023-05-07
fix(core): (part) Add warning message if invalid macro name used

BenJamesBen:
(This will help in situations such as the pattern coder accidentally using `sewtogether` as the macro name, rather than the correct `sewTogether`.)

Screenshot 2023-05-07 at 1 19 59 PM

### DIFF
--- a/packages/core/src/part.mjs
+++ b/packages/core/src/part.mjs
@@ -345,6 +345,7 @@ Part.prototype.__macroClosure = function (props) {
   const method = function (key, args) {
     const macro = utils.__macroName(key)
     if (typeof self[macro] === 'function') self[macro](args, props)
+    else self.context.store.log.warning('Unknown macro `' + key + '` used in ' + self.name)
   }
 
   return method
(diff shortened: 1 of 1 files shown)


BenJamesBen:
I made a change to the proposed fix to work around a test failure, caused when a part has no `context` and thus no `store` (such as during the test).

```
1) Part
       Should not run an unknown macro:
     TypeError: Cannot read properties of undefined (reading 'store')
      at method (file:///home/runner/work/freesewing/freesewing/packages/core/src/part.mjs:348:23)
      at Context.<anonymous> (file:///home/runner/work/freesewing/freesewing/packages/core/tests/part.test.mjs:31:12)
      at processImmediate (node:internal/timers:466:21)
```

joostdecock (approved):

joostdecock:
@BenJamesBen This is a good change, but there's a bigger issue here: The (unwritten) rule is that macro names are always lowercase (see `bnotch`, `cutonfold`, `bannerbox`)

This should be fixed for the `sewTogether` macro (which as far as I know is the only offender) and perhaps be documented.

PR #4014, opened 2023-05-07
fix(bannerbox): Removed hardcoded values in call to `banner`

BenJamesBen:
The `bannerbox` macro was calling the `banner` macro with hardcoded `repeat` and `dy` values. The PR fixes this, using the values passed to the macro.

The default `repeat` value was changed to `99` to reflect the actual hardcoded value that was previously used.

Documentation was updated with the changes. 

I also added additional documentation: 
- The bannerbox is larger than the provided box.
- Information about the offset used to make the bigger box (since during testing I was confused why changing margins seemed to increase offsets by a greater amount than expected). 
- Information about what happens if there is not enough or too much text.

### DIFF
--- a/markdown/dev/reference/macros/bannerbox/en.md
+++ b/markdown/dev/reference/macros/bannerbox/en.md
@@ -16,11 +16,11 @@ macro('bannerbox', {
   String text='',
   Number margin=15,
   Number dy=4,
-  Number repeat=10,
+  Number repeat=99,
   Number spaces=12,
 })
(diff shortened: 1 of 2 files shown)


joostdecock (approved):

SeaZeeZee (approved):
No idea about the code but the extra documentation is clear and readable to me!

BenJamesBen:
Changed the PR status from Draft to ready.

Issue #3956, opened 2023-05-01
[bug]: List options are `undefined` when set to the default value


BenJamesBen:
### What seems to be the problem? 🤔

Using this debug code in a design and view the design it in the lab UI:
```js
bottomArmReduction: {
     pct: 90,
     min: 75,
     max: 125,
     menu: 'style',
     //hide: ({ options }) => options?.type == undefined
     hide: ({ options }) => {console.log('t ' + options); console.log ('v ' +    options?.type); if (options?.type == undefined) return true; return false}
   },
```

We see that when a non-default setting is selected fo `options.type`, it is correctly set to the value:
```
t [object Object]
v octopus
```

However, when the default setting is selected, `options.type` is unexpectedly `undefined`:
```
t [object Object]
v undefined
```

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

This prevents the use of this code:
```js
hide: ({ options }) => options.type == 'octoplushy'
```

The value of `options.type` will never be set to `octoplushy` (the default).

joostdecock:
This was resolved a while ago. The first parameter of the `menu` function receives the settings as provided by the user.
If you want to know what an option's current value is -- and not what the user asked -- you should use the second parameter which holds `mergedOptions` which merges the user's choices with the defaults.

Here's how it is currently implemented:
https://github.com/freesewing/freesewing/blob/7bf47552a7ea9fdb92a5f74fc4515bf5c0db6e4d/designs/octoplushy/src/head.mjs#L649-L655

Issue #3955, opened 2023-05-01
[bug]: `options` are `undefined` in the reset state when setting up lab menus


BenJamesBen:
### What seems to be the problem? 🤔

Using this debug code in a design:
```js
bottomArmReduction: {
     pct: 90,
     min: 75,
     max: 125,
     menu: 'style',
     //hide: ({ options }) => options?.type == undefined
     hide: ({ options }) => {console.log('t ' + options); console.log ('v ' +    options?.type); if (options?.type == undefined) return true; return false}
   },
```

We see that when a design is first loaded in the lab or when the ClearThing button is used, the `options` object is `undefined`:
```
t undefined
v undefined
```

As soon as any option is selected, the options object then starts to exist and continues to exist:
```
t [object Object]
v octopus
```

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

This issue prevents the use of this code when a design is first loaded in the lab.
```js
hide: ({ options }) => options.type == 'octoplushy'
```
And, it generates messages in the console when trying to access the `type` property of `undefined`.

joostdecock:
See #3956

PR #4037, opened 2023-05-11
fix(noble,octoplushy,simon): Add/modify options hide, with workarounds

### LINKED ISSUE #2779: [bug]: [v3] [lab] Options with method-based menu: properties not correctly displayed by UI

New in v3 are options with a method for the `menu:` property, to dynamically determine under which menu the option should appear or, if set to `false`, whether to hide the option. (This replaces `hide:` properties with methods used in v2.)

For example, in Noble:
```
export const armholeDartPosition = {
  pct: 50,
  min: 10,
  max: 90,
  menu: ({ options }) => (options.dartPosition == 'shoulder' ? 'darts' : false),
}
```

However, the UI currently does not display these method-based `menu:` properties correctly yet. Instead, the function string is shown as a series of nested menus by the UI. Example:
Screen Shot 2022-09-09 at 4 36 50 PM

_No response_

No, I am not 😞

### Additional context

The two designs affected by this bug are Noble and Octoplushy.

Simon and Simone still use the old, v2 'hide:` method and are not yet affected. However, my understanding is that they will be converted to the new `menu:` technique and will be affected eventually.

BenJamesBen:
This PR closes #2779 by updating Hi and Noble's part configurations to use functions in `hide` rather than `menu`.

However, this PR also includes workarounds for bugs #3955 and #3956, including an update to Simon's part configuration with one of the workarounds. It might be that we want those bugs to be fixed before merging this PR. If that is the case, then the workarounds can be removed from this PR.

### DIFF
--- a/designs/noble/src/options.mjs
+++ b/designs/noble/src/options.mjs
@@ -11,11 +11,16 @@ export const shoulderDartPosition = {
   pct: 50,
   min: 10,
   max: 90,
-  menu: ({ options }) => (options.dartPosition != 'shoulder' ? 'darts' : false),
+  menu: 'darts',
+  hide: ({ options }) => options?.dartPosition === 'armhole',
 }
 export const armholeDartPosition = {
   pct: 50,
   min: 10,
   max: 90,
-  menu: ({ options }) => (options.dartPosition == 'shoulder' ? 'darts' : false),
+  menu: 'darts',
+  hide: ({ options }) =>
+    options === undefined ||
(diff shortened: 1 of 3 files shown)


eriese:
I think this should wait on the bug fix. I know Joost is refactoring the workbench right now

joostdecock:
Thank you @BenJamesBen This has since been fixed and properly implemented by myself in Simon, and @woutervdub in octoplushy/noble.

Closing.

PR #4040, opened 2023-05-11
fix(workbench): Allow blank optional measurements

### LINKED ISSUE #3973: [bug]: Cannot clear optional measurements once entered in the lab UI

In the lab UI, once optional measurements are entered for a pattern, it is not possible to remove or clear them.

Steps to Reproduce:
1. In either the localhost development environment or lab.freesewing.dev, browse to the Unice design.
2. Select a model to pre-populate the measurements.
3. Switch to DraftDesign view.
4. Switch back to Measurements view, and delete the optional Cross Seam and Cross Seam Front measurements.
5. Switch to DraftDesign view.
6. Switch back to Measurements view, and observe that the deleted measurements re-appeared.
7. Use ClearThing to reset the pattern.
8. In Measurements view, enter all the required measurements, but leave the two optional measurements blank.
9. Switch to DraftDesign view, and observe that the pattern drafts just fine without the optional measurements.
10. Switch back to Measurements view and observe that the optional measurements are blank.
11. Enter the 2 optional measurements.
12. Switch to DraftDesign view.
13. Switch back to Measurements view, and delete the optional Cross Seam and Cross Seam Front measurements.
14. Switch to DraftDesign view.
15. Switch back to Measurements view, and observe that the deleted measurements re-appeared.

A workaround is to use ClearThing to clear all the measurements, but this can require the user to re-enter all the required measurements again manually.

_No response_

No, I am not 😞

### Additional context

_No response_

BenJamesBen:
Fixes #3973.

I don't know if this is an appropriate way or the best way to fix the issue. This PR adds an `optional` parameter allows zero values if optional. (From testing I observed that the value is set to zero when the field is blanked out.)

### DIFF
--- a/sites/shared/components/workbench/inputs/measurement.mjs
+++ b/sites/shared/components/workbench/inputs/measurement.mjs
@@ -19,10 +19,7 @@ export const MeasurementInput = ({ m, gist, app, updateMeasurements, focus, opti
   const isDegree = isDegreeMeasurement(m)
   const factor = useMemo(() => (isDegree ? 1 : gist.units == 'imperial' ? 25.4 : 10), [gist.units])
 
-  const isValValid = (val) => {
-    if (optional && val === 0) return true
-    return typeof val === 'undefined' || val === '' ? null : val != false && !isNaN(val)
-  }
+  const isValValid = (val) => (val == false ? optional : !isNaN(val))
   const isValid = (newVal) => (typeof newVal === 'undefined' ? isValValid(val) : isValValid(newVal))
 
   const [val, setVal] = useState(gist.measurements?.[m] / factor || '')
--- a/sites/shared/components/workbench/inputs/measurement.mjs
+++ b/sites/shared/components/workbench/inputs/measurement.mjs
@@ -17,10 +17,16 @@ export const MeasurementInput = ({ m, gist, app, updateMeasurements, focus, opti
   const title = t(`measurements:${m}`)
 
   const isDegree = isDegreeMeasurement(m)
-  const factor = useMemo(() => (isDegree ? 1 : gist.units == 'imperial' ? 25.4 : 10), [gist.units])
+  const factor = useMemo(
+    () => (isDegree ? 1 : gist.units == 'imperial' ? 25.4 : 10),
+    [gist.units, isDegree]
+  )
 
-  const isValValid = (val) => (val == false ? optional : !isNaN(val))
-  const isValid = (newVal) => (typeof newVal === 'undefined' ? isValValid(val) : isValValid(newVal))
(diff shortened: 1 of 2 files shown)


sites/shared/components/workbench/inputs/measurement.mjs
```diff
@@ -11,16 +11,18 @@ import { measurementAsMm } from 'shared/utils.mjs'
...
   const title = t(`measurements:${m}`)
 
   const isDegree = isDegreeMeasurement(m)
   const factor = useMemo(() => (isDegree ? 1 : gist.units == 'imperial' ? 25.4 : 10), [gist.units])
 
-  const isValValid = (val) =>
-    typeof val === 'undefined' || val === '' ? null : val != false && !isNaN(val)
+  const isValValid = (val) => {
+    if (optional && val === 0) return true
+    return typeof val === 'undefined' || val === '' ? null : val != false && !isNaN(val)
```
eriese, line 24: I would just replace these lines with `return val  == false ? optional : !isNaN(val)` because `undefined`, `0`, and `''` are all falsy, and any of them should be allowed on an optional measurement

sites/shared/components/workbench/inputs/measurement.mjs
```diff
@@ -11,16 +11,18 @@ import { measurementAsMm } from 'shared/utils.mjs'
...
  * measurement and I always have some typo in it because dyslexia.
  */
-export const MeasurementInput = ({ m, gist, app, updateMeasurements, focus }) => {
+export const MeasurementInput = ({ m, gist, app, updateMeasurements, focus, optional }) => {
   const { t } = useTranslation(['app', 'measurements'])
   const prefix = app.site === 'org' ? '' : 'https://freesewing.org'
   const title = t(`measurements:${m}`)
 
   const isDegree = isDegreeMeasurement(m)
   const factor = useMemo(() => (isDegree ? 1 : gist.units == 'imperial' ? 25.4 : 10), [gist.units])
```
eriese, line 20: I know it's not in the scope of the PR, but while you're here can you please fix these lint warnings? If fixing them messes everything up we can bump it to the workbench refactor

eriese (changes requested):

sites/shared/components/workbench/inputs/measurement.mjs
BenJamesBen: I believe that I eliminated the `eslint` warnings in a way that is correct. I had to add more `useCallback()` hooks to deal with additional `eslint` warnings that popped up due to the changes.

sites/shared/components/workbench/inputs/measurement.mjs
BenJamesBen: Replaced code as suggested.

eriese (approved):

PR #4041, opened 2023-05-11
fix(simon): Correct cuff button placement

### LINKED ISSUE #3967: [bug]: (simon) Cuff Button Rows option isn't working properly

By default, Simon's `cuffButtonRows` option is set to 1, and the button/hole placement looks correct:
Screenshot 2023-05-02 at 5 12 35 PM

However, when `cuffButtonRows` option is set to 2, the button/hole placement appears to be incorrect:
Screenshot 2023-05-02 at 5 12 43 PM

And, when `cuffButtonRows` is set to 1 by using the slider, the button/hole placement again appears to be incorrect, and it is different than the default (which was also 1, so they should have been identical):
Screenshot 2023-05-02 at 5 12 50 PM

The issue is seen in both lab.freesewing.dev and `develop`.

designs/simon

No, I am not 😞

### Additional context

_No response_

BenJamesBen:
Fixes #3967.

`options.cuffButtonRows` was getting cast to a string, so instead of correct row spacing of 1/2 or 1/3, the incorrect spacing of 1/11 or 1/22 was used.

After:
Screenshot 2023-05-11 at 12 25 18 PM
Screenshot 2023-05-11 at 12 25 27 PM

### DIFF
--- a/designs/simon/src/shared.mjs
+++ b/designs/simon/src/shared.mjs
@@ -147,7 +147,7 @@ export const decorateBarrelCuff = (part) => {
   for (let i = 1; i <= options.cuffButtonRows; i++) {
     points['button' + i] = points.buttonLineTop.shiftFractionTowards(
       points.buttonLineBottom,
-      (1 / (options.cuffButtonRows + 1)) * i
+      (1 / (Number(options.cuffButtonRows) + 1)) * i
     )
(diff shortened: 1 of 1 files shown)


joostdecock (approved):

joostdecock:
Really great work @BenJamesBen 👍

PR #4073, opened 2023-05-18
[docs] Added docs for core settings

joostdecock:
This PR is merging into the `joost` branch. I'll change that once #3636 is merged. But this makes it easier to review.

### DIFF
--- a/markdown/org/docs/site/draft/core-settings/complete/en.md
+++ b/markdown/org/docs/site/draft/core-settings/complete/en.md
@@ -2,5 +2,11 @@
 title: Pattern Details
 ---
 
+<Fixme> 
+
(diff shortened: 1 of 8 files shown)


BenJamesBen:
I assume that this is documentation for the org site UI? If so, it looks good, though without having the actual UI to look at it is somewhat difficult to review. For example, I am assuming that the UI uses the wording "Included Parts" and not something else like "Contents".

I'm also assuming that the org site UI is different than the lab workbench UI?  And, this is just documentation for the org site UI?

joostdecock:
Merged as-is because done is better. If you'd like to include the defaults, I guess we could be I think it's not as useful as the default is listed in the UI when working on patterns.

PR #4074, opened 2023-05-18
[docs] Designs not patterns

joostdecock:
This PR is merging into the `core-settings-docs` branch. I'll change that once #4073 3636 is merged. But this makes it easier to review.


BenJamesBen (approved):

BenJamesBen:
This PR affects a large number of files. The GitHub web pages were very slow to load and not everything was loaded because there were so many files. However, I looked it over the best i could, and it appears that the only changes in this PR are:
1. Rename doc files to move them from the old "patterns" docs subdirectory to the new "designs" subdirectory.
2. Change the top-level `en.md` file to reflect the new "Designs" wording.

So, the PR looks fine to me. (But, this is assuming that the org website will keep the same organization that was used previously in v2. That is, all org site documentation is under the "docs" directory and that our design documentation is in a subdirectory immediately under "docs". I'm assuming that we don't want to organize things differently.)

PR #4089, opened 2023-05-19
fix(docs): Move optional measurements info from Part config to Part draft

BenJamesBen:
This PR addresses the concerns in #3974, suggesting that information about optional measurements did not belong in the Part config documentation and instead should be moved to Part draft documentation.

### DIFF
--- a/markdown/dev/reference/api/part/draft/en.md
+++ b/markdown/dev/reference/api/part/draft/en.md
@@ -3,14 +3,14 @@ title: "The part's draft method"
 ---
 
 Each part **must** have a `draft` property that holds a method that will draft the part.
-In other words, this method is where the actual work happens. The method's signature 
+In other words, this method is where the actual work happens. The method's signature
 is as follows:
 
 ```js
 function draft(props)
 ```
 
-The draft method receives a single parameter, an object which you can _destructure_ to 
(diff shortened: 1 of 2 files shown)


markdown/dev/reference/api/part/draft/en.md
```diff
@@ -45,3 +45,11 @@ access the following properties:
 | `Bezier`          | The [bezier-js](https://pomax.github.io/bezierjs/) library's `Bezier` named export |
 || **_Return value_**   |
 | `part`            | Your draft method **must** return this |
+
```
joostdecock, line 48: I find this adds more confusion than anything else.
There is nowhere ever that it's mentioned that there's an `optionalMeasurements` key. 

Just because somebody made a mistake once and then others blindly copy/pasted the same mistake, I don't think we need to add this to the docs.

BenJamesBen: I don't know if the addition to the docs is absolutely needed. However, I think that it helps to add clarity, and it does not add confusion.

In #3933, before learning that `optionalMeasurements` did not exist, I was convinced that it did actually exist and was somehow missing from our code. I believed it to the point that I spent time working on a code "fix" to add `optionalMeasurements` back into part shorthand to try to resolve the problem.

joostdecock (approved):

Issue #3909, opened 2023-04-27
[bug]: Titan cisFemale sizes 28-42 are not packed correctly in Cutting Layout


BenJamesBen:
### What seems to be the problem? 🤔

Regardless of the width of fabric used, a Titan pattern generated from cisFemale sizes 28 to 42 is packed suboptimally in the Cutting Layout. Instead of being arranged to fill the available width, parts are arranged vertically:
![Screenshot 2023-04-26 at 8 28 29 PM](https://user-images.githubusercontent.com/109869956/234752292-920c2eb8-909a-4dbe-a4e8-c8de8cfe8948.png)

This problem isn't seen with other sizes, for example cisFemale size 44:
![Screenshot 2023-04-26 at 8 29 40 PM](https://user-images.githubusercontent.com/109869956/234752526-0cdbc5da-bccd-4e66-b2c5-cbc05bd4e219.png)

Steps to Reproduce:
1. In the `develop` lab, select the Titan design, cisFemale Size 28.
2. Switch to the Cutting Layout view, and change the Fabric Width to 999 cm.
3. Observe that the layout is packed vertically instead of packing horizontally to fill the available width.
4. Switch to the Measurements view, and select cisFemale Size 46.
5. Switch to the Cutting Layout view (still set to 999 cm width), and observe that the layout is packed correctly, horizontally.
6. Switch to the Measurements view, and select cisFemale Size 30.
7. Siwtch to the Cutting layout view, and observe that the layout is packed vertically instead of filling out the available horizontal width.

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

I see similar incorrect-looking packing behavior with other designs As I understand it, we use a packing strategy of trying to pack into an approximate square, and I am not sure if this might be related.

eriese:
This is not a bug. Our extremely simplistic packing algorithm in a square was sub-optimal as it was, but for cutting layout I added a width constraint so that it would not pack wider than the fabric and that has made the packing exponentially more terrible. I assume that what is happening here is that at the front piece is slightly taller than the back piece, so the algorithm believes it does not fit next to the back piece. The only way to fix this issue would be to use a smarter algorithm. If the consensus is that this is truly not presentable, we can slap a warning on the page that we do not vouch for the efficiency of the default layout.

[...]

PR #4092, opened 2023-05-20
fix(core): Add strictMax property to bin pack() call

BenJamesBen:
This PR adds a `strictMax` property to the bin `pack()` call to take advantage of packing improvements to be implemented in https://github.com/freesewing/bin-pack-with-constraints/pull/1 .

With that PR and this one, there will be improvements in parts packing in cutting layouts, as described in #3909.

This PR can be merged before or after https://github.com/eriese/bin-pack-with-constraints/pull/1. Currently, the `strictMax` property is ignored by `bin-pack-with-constraints`, so merging this PR before that one will result in no change in behavior.

### DIFF
--- a/packages/core/src/pattern/pattern-renderer.mjs
+++ b/packages/core/src/pattern/pattern-renderer.mjs
@@ -131,7 +131,11 @@ PatternRenderer.prototype.__pack = function () {
   }
   if (settings[activeSet].layout === true) {
     // some plugins will add a width constraint to the settings, but we can safely pass undefined if not
-    let size = pack(bins, { inPlace: true, maxWidth: settings[0].maxWidth })
+    let size = pack(bins, {
+      inPlace: true,
+      maxWidth: settings[0].maxWidth,
+      strictMax: settings[0].maxWidth ? true : false,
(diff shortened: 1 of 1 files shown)


joostdecock:
This is all feeling a bit ad-hoc and brings back memories to the time before the monorepo where changes in various repositories had to carefully click together to make things work. I really don't want to go back to that.

So @eriese forked the bin-pack package for their fabric cutting layout. Do we want this? Not sure. I'm sort of letting things evolve as things are in alpha and I appreciate that we're still figuring out what works best.

But if we are serious about maintaining our own fork of `bin-pack` it should be added as a package in our monorepo.
On the other hand, core doesn't care because the cutting layout thing is not a core feature. So it probably makes more sense to see how we can plug a different layout engine into core so that people have options for the layout. Perhaps we want a different layout engine for the cutting layout, and so on.

I feel that's something we need to figure out because my concern is that people are somewhat cavalier about the API boundary between core and our frontend.

joostdecock (approved):

joostdecock:
Superseded by #5133

PR #4098, opened 2023-05-21
fix(octoplushy): Three fixes and add cutlist

BenJamesBen:
This PR fixes 3 issues with Octoplushy, and it also adds the cutlist.
1. The eye parts paths were mistakenly hidden during the hide/render core change awhile ago.
2. With the new cutlist feature, empty parts need to be explicitly hidden. Otherwise, they show up in the printing and cutting layouts as just a floating button.
Screenshot 2023-05-21 at 11 21 15 AM
3. The pattern crashes if the type is Squid and Contents/Only is just armSection2. This is because armSection2 depends on `armSkirtToTopAngle` being set in the store, and armSection2 is after: headSection2. However, `armSkirtToTopAngle` only gets set in headSection1. There were 2 ways to fix this: 1. Make armSection2 after: headSection1 as well. 2. Fix head.mjs so it sets the needed store variable in headSection2. (I chose to implement the first fix.)

Submitting as draft to allow the pattern designer time to look over the fixes.


woutervdub:
Looks good to me. Haven't tested it.

joostdecock (approved):

PR #4154, opened 2023-05-26
[wip] org: Work on patterns pages and saving/editing of patterns

joostdecock:


joostdecock:
Note that linter issues are all in the XRay code which is still v2.5 stuff.

PR #4156, opened 2023-05-26
fix(docs): (snap) Metric/imperial properties can hold a number, in addition to holding an array

BenJamesBen:


BenJamesBen:
See: https://github.com/freesewing/freesewing/blob/develop/packages/snapseries/src/index.mjs and the `smallSteps`, `steps`, and `bigSteps` exports for examples of the scenario added to the documentation.

joostdecock (approved):

PR #4181, opened 2023-05-31
DRAFT: fix(bella): Add bustDartAngle option and keep bust dart location along side seam

BenJamesBen:
This PR:
1. Adds a `bustDartAngle` option to Bella that allows the customer to move the bust dart to a different location on the side seam.
2. Forces the bust dart location to start at the armhole in situations where the calculations would otherwise tell it to start above the armhole.
3. Forces the bust dart location to end at the bottom of the side hem in situations where the calculations would otherwise tell it to be located below the hem.
4. Adds `log.info` messages when the bust dart location is restricted in situations 2 and 3 above.
5. Adds `bustDartAngle` option constants to Bee and Noble, to remove the option, so there is no change in behavior for those two designs.

Edit: Added:
6. Changes `bustDartAngle` min from 0% to -100%, to allow convex side darts.

This fixes a current bug in Bella where, if the dart is located below the bottom of the hem, the side seam is incorrectly increased so the front side seam is longer than the back side seam.
Screenshot 2023-05-30 at 4 15 17 PM
https://discord.com/channels/698854858052075530/757632180980547686/1113245412132921374
discord://discord.com/channels/698854858052075530/757632180980547686/1113245412132921374

---

With the fix: Bella's front when `bustDartAngle` settings force the design to restrict bust dart location at top and bottom of seam and the log messages:
Screenshot 2023-05-30 at 7 52 55 PM
Screenshot 2023-05-30 at 7 53 11 PM
Screenshot 2023-05-30 at 7 53 26 PM
Screenshot 2023-05-30 at 7 53 46 PM

---

Side dart curvature with `bustDartCurve` set to the new -100% minimum:
Screenshot 2023-05-31 at 5 51 17 AM

### DIFF
--- a/designs/bee/src/cup.mjs
+++ b/designs/bee/src/cup.mjs
@@ -15,6 +15,7 @@ export const cup = {
     sideCurve: { pct: 0, min: -50, max: 50, menu: 'fit' },
     frontCurve: { pct: 0, min: -50, max: 50, menu: 'fit' },
     bellaGuide: { bool: false, menu: 'fit' },
+    bustDartAngle: 0,
   },
   draft: ({
     store,
--- a/designs/bella/src/back.mjs
+++ b/designs/bella/src/back.mjs
@@ -37,7 +37,7 @@ export const back = {
     bustDartLength: { pct: 90, min: 75, max: 100, menu: 'darts' },
     waistDartLength: { pct: 90, min: 75, max: 100, menu: 'darts' },
     bustDartAngle: { count: 0, min: -20, max: 55, menu: 'darts' },
-    waistDartCurve: { pct: 100, min: 0, max: 100, menu: 'darts' },
+    waistDartCurve: { pct: 100, min: -100, max: 100, menu: 'darts' },
     // Armhole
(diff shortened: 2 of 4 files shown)


BenJamesBen:
Filed as Draft. I still need to add doc files to this PR for the new `bustDartAngle` option.

The addition of the `bustDartAngle` option is entirely optional, and it can be removed from this PR if desired. (If we remove it, we can/should also remove the fix for scenario 2 because the dart never gets located above the armhole using the current 0 degree angle.)

BenJamesBen:
7. Also adds `waistDartCurve` option to shape waist dart and changes max length to 100% to allow the dart tip to be located at the apex (to facilitate dart manipulation).

joostdecock:
Similar question here as for #4329 @BenJamesBen how do we proceed with this draft PR?

joostdecock:
Two month later, so I'm going to close this. Anyone who wants to pick this up can do so.
If you like keeping PRs open for 6 months, then keep in mind you can keep the PR in your fork until it's ready.

PR #4190, opened 2023-05-31
Refactor (Sites/Shared) Workbench menus use shared components

eriese:
This refactor brings in more shared building blocks for workbench menus, (currently Core Settings and Design Options) to make building additional view-specific menus faster and making ui changes easier.

The main interventions are in the new `sites/shared/components/workbench/menus/shared` folder. We have base inputs, base value display components, base MenuItems, and a base Menu which can either display children or loop recursively through a config to figure out what to display.

Everything works on org right now


joostdecock:
Can you target the tip of the UI work which is in the `joost` branch.

Tx

PR #4201, opened 2023-06-01
wip(shared): Adding the pattern inspector

joostdecock:
Aka X-Ray mode


sites/shared/components/workbench/views/draft/inspector/shared.mjs
```diff
@@ -0,0 +1,631 @@
...
+  return showInfo
+}
+
+export const KeyValTable = ({ rows }) => (
+  <table className="border-collapse h-fit">
+    <tbody>
+      {rows.map((row) => (
+        <tr>
+          <td className="font-bold p-2 pr-0 text-right">{row[0]}:</td>
+          <td className="p-2">{row[1]}</td>
```
eriese, line 26: obnoxiously, these need keys

eriese (commented):
Looking good so far. I like the package for our pattern components.

`workbench/views/draft/inspector/shared.mjs` is a good candidate for getting broken into smaller concerns

eriese:
@joostdecock how would you feel about me preparing a branch off of this one that can be merged into `develop` without you having to be done with the pattern inspector? Then we can put up a build and start getting feedback on the rest of the ui while you continue this work

eriese:
Seems I did not fix everything. I'll go again

joostdecock:
It is slow, that's a problem that needs TLC. But people don't have to use it if they think it's too slow.

eriese:
we're getting a major security warning for using `dangerouslySetInnerHtml` in the new `Defs` because it doesn't do any safety checks or escaping. I'm struggling to wrap my head around what's best/safer

joostdecock:
> we're getting a medium security warning for using `dangerouslySetInnerHtml` in the new `Defs` because it doesn't do any safety checks or escaping. I'm struggling to wrap my head around what's best/safer

The *library input* in this case is what's returned from `getRenderProps()`, which is not user input but our own core library output which is sanitized. This is a false positive and it I feel we can just dismiss the alert.

I didn't really like how you changed the type under the hood without any way for the user to know what's going on.
So I've changed that in https://github.com/freesewing/freesewing/pull/4201/commits/01722da3911154fc0e2d8759d5c214ebd26f8ea3 to show an error which we needed in any case because when the backend is down or there's some other problem things just kept loading for eternity:

image

eriese:
> The _library input_ in this case is what's returned from `getRenderProps()`, which is not user input but our own core library output which is sanitized. This is a false positive and it I feel we can just dismiss the alert.

It's a false positive when we use it for our front end with known input, but we're now publishing a library, so when someone uses it as a such they may choose to allow unknown input. We can't cover all use cases, but we may want to take a tip from react and do something to make it clearer that the safety of the strings you put into defs is your responsibility. 



> I didn't really like how you changed the type under the hood without any way for the user to know what's going on.
> So I've changed that in [01722da](https://github.com/freesewing/freesewing/commit/01722da3911154fc0e2d8759d5c214ebd26f8ea3) to show an error which we needed in any case because when the backend is down or there's some other problem things just kept loading for eternity:

Great. For some reason I always forget about error pages

joostdecock:
@eriese I've addressed the XSS warning in #4250

PR #4227, opened 2023-06-05
WIP Feature (Shared) Port Print and Cut Layout to v3 ui

eriese:
Reimplements the printing layout and cutting layouts using the new ui components


joostdecock (changes requested):
Happy to merge this, with the exception of the changes to the `react-components` package.

I perhaps have failed to clarify the purpose of this package, but the idea is that it can be used by people who use FreeSewing to render a pattern in their own frontend.
As such, there should be a clear separation of concerns (nothing that's specific to our frontend should go in this package) nor should we make assumptions of how it is loaded.

In practice, that means that these changes should be re-evaluated:

**Removing the React import**

This breaks compatibility with React versions prior to v17. We can perhaps debate whether we want to support this yes or no, but it seems a rather opinionated change to slide into this PR 

**Changes to components inside that package**

Long story short: Don't. If you need more than merely rendering the pattern, extend the component on your own turf and pass it in to override the default. You can check how it's handled in the inspector as an example: https://github.com/freesewing/freesewing/blob/joost/sites/shared/components/workbench/views/draft/inspector/pattern.mjs

**Adding to utils in the package**

Same reasoning. If it's not required to render a pattern, it doesn't belong here.
However, if you pass in your custom components, there will also no longer be a need to mage changes to this utils file. You can import from a local utils/shared file instead.

Here too, you can refer to the inspector example: https://github.com/freesewing/freesewing/blob/fb9e9cf5e1331f9dc8cb618ed14cd9afc650f934/sites/shared/components/workbench/views/draft/inspector/path.mjs#L3



To summarize, happy to merge this functionality. But we have to keep our own frontend fireworks separate from the basic render functionality provided by this component.

eriese:
> **Removing the React import**
> 
> This breaks compatibility with React versions prior to v17. We can perhaps debate whether we want to support this yes or no, but it seems a rather opinionated change to slide into this PR

Oops, didn't understand the purpose of that. I'll put it back

> **Changes to components inside that package**
> 
> Long story short: Don't. If you need more than merely rendering the pattern, extend the component on your own turf and pass it in to override the default. You can check how it's handled in the inspector as an example: https://github.com/freesewing/freesewing/blob/joost/sites/shared/components/workbench/views/draft/inspector/pattern.mjs

I think you'll if you look closer at the changes I made you'll find that they are almost all fixes that are within the concerns of the package:

- added a `forwardRef` to `Group`. possibly debatable, but I think this could be useful for more than just the `movable` use case
- add `Circle` to the list of default components in `Pattern`. Anything that used a circle was breaking without this.
- add `forwardRef` to `Pattern`. `ref` was being passed as a prop, but that doesn't work, so this fixes it.
- attach `ref` passed to `Pattern` to the `Svg`. This was being done via `optionalProps`, but again, `ref` doesn't work when passed as a prop
- access `settings.units` correctly in `Part`. Probably we should change this so that the settings set that applies to a part is in its ` renderProps`, but in the mean time `settings` are being passed all the way down the chain from the `pattern`, which means they're an array

> **Adding to utils in the package**
> 
> Same reasoning. If it's not required to render a pattern, it doesn't belong here. However, if you pass in your custom components, there will also no longer be a need to mage changes to this utils file. You can import from a local utils/shared file instead.

Understood and agreed

joostdecock:
> Oops, didn't understand the purpose of that. I'll put it back

Ah ok, then perhaps I should explain a bit: Prior to v17 of react, `React` needed to be in scope (as in, imported) to use JSX syntax.
Since v17, this is no longer required, but it doesn't hurt (as you need React in the bundle anyway to use these components) and it's kinder to people running older versions. So while we don't import it anymore in our own frontend code (because we run a recent react) I'm taking a more conservative approach with the published components.

Ok to do the ref thing if it can't be passed via props (I avoid refs if at all possible, so I'm not that familiar)

joostdecock (approved):

PR #4233, opened 2023-06-06
Implement Test View in new UI

eriese:
Not super impressed with what I came up with visually here, but the functionality for testing options and measurements is in there, so someone who wanted to change the look and feel could now more easily do so.


sites/org/pages/new/pattern/[design]/[type]/[id].mjs
```diff
@@ -14,7 +13,7 @@ import { Null } from 'shared/components/null.mjs'
 import { DynamicOrgDocs as DynamicDocs } from 'site/components/dynamic-org-docs.mjs'
 
 // Translation namespaces used on this page
-const namespaces = [...new Set(['aaron', ...wbNs, ...pageNs])]
+const namespaces = [...new Set([...wbNs, ...pageNs])]
```
Comment (joostdecock) [inline sites/org/pages/new/pattern/[design]/[type]/[id].mjs:16]: Since this is a common pattern, I've added the `nsMerge()` method to `shared/utils.mjs` which does this. (merge all namespaces without dupes)

joostdecock (approved):

sites/org/pages/new/pattern/[design]/[type]/[id].mjs
Comment (joostdecock) [reply sites/org/pages/new/pattern/[design]/[type]/[id].mjs:16]: This is FYI, not blocking the merge because of this

PR #4255, opened 2023-06-09
Migrate translations to v3

joostdecock:
This merge request implements the new way we handle translations in v3.
It also removed that `@freesewing/i18n` package and updates the `@freesewing/plugin-i18n` plugin.

Currently. these translations are kept in the i18n package, and there's a bit of prebuild magic going on to take care of shared optinos and so on. It's all a bit brittle and hard to wrap your head around, plus it suffers from the same issue with component translation which is that it is not very obvious where the translations go.

This PR changes that, by (as a proposal) implementing translation for 2 designs: Brian and Aaron.
I picked these because Aaron is our go-to example pattern, and because it extends Brian, re-uses some of its options, but not all. So it is a good enough test case to handle inheritance in translation.

## Where are translations now?

In this proposal, a design will (apart from a `src` folder) have a `i18n` folder. Inside, there should be a `.json` file for each language. For example `en.json`.

I have opted to not go for YAML here for a couple of reasons:
- Both NodeJS and Esbuild can load JSON natively, so we don't need a prebuild step
- Crowdin supports JSON, so it's not like translators have to change these files
- The English files will (originally) be created by pattern designers. If you can write code, you can also write JSON

The `i18h` folder also holds a `index.mjs` file that imports all the languages and then re-exports them as the `i18n` named export.

In the main entrypoint `src/index.mjs` we then import this and re-export this as the `i18n` named export.
At least in Brian we do, in Aaron, the situation is more complicated because we are need some translations from brian, but not all.

So, in Aaron, we import the `i18n` names export from Brian (in addition to our own `i18n` and then merge them together.

## How are translations merged

Core has a new `mergeI18n` names export. This is method that takes two parameters:
- an array of translation objects
- a configuration object to indicate how you want to merge the translations (see below)

## How are these translation objects structured?

The JSON and resulting POJO is an object with a key for each language:
```js
{
  en: {}. // english here
  de:  {}, // german here
  // and so on
}
```

Each language entry should be structured as such:

```js
{
  t: "The title of the design",
  d: "The description of the design",
  s: {
      someSting: "Translation for someSting that is used in the pattern",
      // more strings here
  },
  p: {
    front: "The title of part front",
    back: "The title for part back",
    // and so on
  },
  o: {
    optionA: {
      t: "The title for optionA",
      d: "The description for optinoA",
    },
    // other options here
  }
}
```

The `t` and `d` keys are mandatory. The `p`, `s`, and `o` keys are optional.

## How do I configure the merging of translations?

By passing a config object to `mergeI18n`. This object can have one or more of the `s`, `p`, `o` keys. For each of these it takes an object as follows:

```js
{
  keep: ['a', 'b', 'c'], // Keep only these
  drop: ['d', 'e', 'f']. // Drop these
}
```

So if you want to drop a couple of options, you configure it like this:

```js
{
  o: {
    drop: ["notYou", "alsoNotYou" ]
  }
}
```

Or if you want to only retain some parts, you do this:

```js
{
  p: {
    keep: ["thisPart", "andThatPart"]
  }
}
```

## mergeI18n signature

```js
function mergeI18n(array, object) 
```

The array can hold as many designs as you want. Later designs will overwrite the translations from earlier designs if they both provide them.

The second parameter is the config object. If you want to merge more than 2 designs (not sure why you'd need that but ok) and you want to use a different merge config, you cannot do that in 1 call. But you can call `mergeI18n` once and then call it again with the result of the first call as part of the first parameter. Since it's compatible like that. So you can merge multiple designs with different merge configs.


eriese:
Conceptually, I'm half on board: 

The pro: the i18n prebuild uses a lot of magic, and this doesn't add to the pile of magic, so that's good. 
The con: another step for designers to remember. it's not a very big con because we'll add it to the new-design templates, but it's not as though we can get rid of prebuild altogether so I'm not opposed to a bit more magic

A suggestion for if this is the implementation: The merge config isn't currently optional, and I think it should be. I'm guessing that config is there to allow us to keep the files smaller, and I can see that if there are no `keeps` and `drops` it does a straightforward overwrite, which I think is what people will expect. I would imagine that in many cases everything will be a keep.

joostdecock:
Started working on this. Changed the title from proposal to WIP to indicate that.

@eriese made the `options` param optional in `mergeIi18n()` as you suggested. It will keep everything in that case.

packages/core/src/utils.mjs
```diff
@@ -704,6 +732,25 @@ export function __macroName(name) {
...
+/**
+ * Returns true if we want to keep the translation
+ * Called by mergeI18n
+ *
+ * @private
+ * @param {string} key - The translation key
+ * @param {object} options - The options (for this particular section of the translation file)
+ * @return {bool} result - Whether or not to keep the translation
+ */
+function __keepTranslation(key, options) {
```
eriese, line 744: default options to `{}`

packages/core/src/utils.mjs
```diff
@@ -441,6 +441,34 @@ export function lineIntersectsCurve(start, end, from, cp1, cp2, to) {
   else return intersections
 }
 
+/**
+ * Helper method to merge translation files from different designs
+ *
+ * @param {object} translations - One or more translation objects
+ * @return {object} result - A merged object of translations
+ */
+export function mergeI18n(designs, options) {
```
eriese, line 450: default `options` to `{}`

eriese (changes requested):
I'm good with this once the checks pass and those defaults are added

PR #4256, opened 2023-06-10
Partial language change 'Measurements Set(s)'

SeaZeeZee:
PR that partially changes people and person to measurements sets and measurements set.


joostdecock (approved):

joostdecock:
The org docs have been rewritten, and I'm not blocking migration for dev docs.
As such, this is completed. Thanks for the effort.

PR #4305, opened 2023-06-17
WIP (org) Implement blog and showcase pages

eriese:
wanted to get this up ASAP because I wanted to make sure that the implementation makes sense, @joostdecock. It's a full copy and paste of the v2 pages, with the exception that it now loads from Sanity instead of Strapi. I'm sure if we want to change the look and feel later, that's fine, but I think that these pages simply loading properly satisfies the requirements for v3-beta.

I think I'm maybe supposed to do something different with the side nav, but I'm also not sold on that side nav being there at all, so just let me know

WRT to Sanity: Despite the fact that `site-content` is a public dataset, I cannot seem to get it to return anything in a query without an auth token (you can see this issue if you open [this CDN query, copied directly from Sanity Vision, where it was working](https://hl5bw8cj.api.sanity.io/v2023-06-17/data/query/site-content?query=*%5B_type%20%3D%3D%20%24type%5D&%24type=%22blogen%22) in an incognito session. [this query for the whole dataset](https://hl5bw8cj.api.sanity.io/v2023-06-17/data/query/site-content?query=*) does work, though) This is mildly annoying because their CDN won't cache requests that have an auth token, and I can't really figure out why it's happening, but for now I'm passing the token via `process.env.SANITY_TOKEN`. Given that we will need the token to access user data anyway, I'm thinking it turns out splitting into two datasets isn't as worthwhile as it initially seemed.


eriese:
Ok, I think I figured out why the queries don't work without authentication: 

[According to the docs](https://www.sanity.io/docs/ids)
> The default, fixed access control rules give unauthenticated users read access to documents under the root path only, which means that it is not possible to make documents under a sub-path (i.e. containing a . in the ID) publicly available. This currently cannot be changed unless you have custom access controls enabled on your plan.

Our blog posts have `.`s in their ids, so they're not publicly searchable

joostdecock:
> *Our blog posts have .s in their ids, so they're not publicly searchable*

@eriese Nice catch. That was a real footgun on their side.

The custom access control the mention only comes with the pricey plans, so we're going to have to work with the system here, rather than against it.
So, I've delete the `site-content` dataset, re-created it, and then re-imported all strapi content, but this time using `--` instead of `.` in the IDs.

I will merge this as-as now (I'll fix the linter stuff later) because with this being in WIP and your work on the measurements view, I think we'll have too many blockers if we keep all of this open.

PR #4320, opened 2023-06-20
feature (workbench) implement measurements view in the workbench

eriese:
Goals of this view

- [x] allow users to change the measurements on the pattern without changing the measurements of the saved set
  - [x] reuse measurement input components from sets pages (I did this, and I also changed the very handy fraction helpers to take up less space)
- [x] allow users to choose a different set to draft the pattern for
  - [x] reuse the measurements set picker
  - [x] change url but keep same settings?


eriese:
> I can push what I was doing to a branch when I get home tonight so you can (if you like) put two and two together. Or, if you prefer, you can focus on the manual inputs as I have already made headway on the selections.

Yes please, push it and I'll merge it. The good news is that it sounds like we focused in different places. Manual inputs are implemented in this branch, but picking a new set hasn't been modified at all

joostdecock:
@eriese alright done. It's this commit: https://github.com/freesewing/freesewing/commit/d1769348b308c4be44ad2a90b8d71d78579eb348

Please forgive me, this is very unpolished, I pushed it as soon as I could without doing anything else.
Good to know:

- I rearranged things in `org/pages/new` but did not (yet) remove any of the old pages
- I copied the main workbend `index.mjs` to `new.mjs` because I wanted to avoid clashing with other changes

eriese:
Great. Thanks

eriese:
@joostdecock is the joost branch missing a commit? I'm having a lot of errors trying to use the new design page

eriese:
never mind. I figured it out

joostdecock:
Sorry, I was working in lab, not in org (because it's faster)

joostdecock:
@eriese I was cleaning things up a bit and pushed this commit which also loads the measurements if you click on a set: https://github.com/freesewing/freesewing/commit/0e85bf5659d972fe20d0dd3e97bcab81c0e8f654

I'll leave it to you now ;)

eriese:
Ok @joostdecock , I think it's ready to be merged

sites/shared/components/sets/measie-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+const heightClasses = {
+  2: 'h-12',
+  4: 'h-10',
+  8: 'h-8',
+  16: 'h-6',
+  32: 'h-4',
+}
+const fractionClasses =
+  'w-full border-2 border-solid border-base-100 hover:border-secondary bg-secondary rounded bg-opacity-50 hover:bg-opacity-100'
+const FractionButtons = ({ t, fraction }) => (
```
joostdecock, line 29: Surprised to see there's no empty line before this method. Do you not have the prettier pre-commit hook setup?

sites/shared/components/sets/measie-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+  <div className="flex flex-row mt-1 content-center items-center justify-around">
+    <span className="text-xs inline-block pr-2">{t('fractions')}</span>
+    <div className="grow max-w-2xl flex items-baseline">
+      {Array.from({ length: 31 }, (_null, i) => {
+        let denom = 32
+        let num = i + 1
+
+        for (var n = 4; n > 0; n--) {
```
joostdecock, line 37: Use of `var` here makes me suspect this was copy/pasted? 
If that's the case, that does not need to be a problem. But we need to do due diligence to ensure the original license does not pose restrictions on derived works that we would not be able to honor.

Also, use `const` here instead of `var`.

sites/shared/components/sets/measie-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+          let fac = Math.pow(2, n)
```
joostdecock, line 38: Use `const` here instead of `let`

sites/shared/components/sets/measie-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        return (
+          <span className="group w-[3.125%] relative" key={i}>
+            <button
+              className={`${heightClasses[denom]} ${fractionClasses}`}
+              title={`${num}/${denom}"`}
+              onClick={() => fraction(num, denom)}
+            />
+            <span className="group-hover:visible invisible text-xs text-center absolute left-0 -bottom-6">{`${num}/${denom}"`}</span>
```
joostdecock, line 53: Please use the inch symbol `″` instead of the straight quotes character `"`

sites/shared/components/sets/measie-input.mjs
joostdecock: In any case, please leave an empty line between methods

sites/shared/components/sets/measie-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+          </span>
+        )
+      })}
+    </div>
+  </div>
+)
+export const MeasieRow = (props) => {
```
joostdecock, line 60: Please add an empty line prior to this method

sites/shared/components/sets/measie-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+  const { t, m, mset } = props
+
+  const isSet = typeof mset.measies?.[m] === 'undefined' ? false : true
```
joostdecock, line 63: Same outcome and IMHO one of those instances where terser is also easier to read:

```js
const isSet = typeof mset.measies?.[m] !== 'undefined'
```

sites/shared/components/sets/measie-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+  const isValid = (newVal) => (typeof newVal === 'undefined' ? isValValid(val) : isValValid(newVal))
+
+  const [val, setVal] = useState(mset.measies?.[m] / factor || '')
+  const [valid, setValid] = useState(isValid(mset.measies?.[m] / factor || ''))
+
+  // Update onChange
+  const update = (evt) => {
+    setVal(evt.target.value)
+
+    let useVal = isDegree
```
joostdecock, line 113: Looks like this should be a `const` not a `let`

sites/shared/components/sets/measie-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+      <MeasieInput {...props} />
+    </Collapse>
+  )
+}
+
+export const MeasieInput = ({
+  t,
+  m,
+  mset,
+  startLoading,
```
joostdecock, line 91: I think it makes more sense to let this default to `() => null` rather than pass in `emptyFunc` later?

joostdecock: Some for `stopLoading` obvs

sites/shared/components/workbench/views/measies/editor.mjs
```diff
@@ -0,0 +1,27 @@
+import { MeasieInput, ns as inputNs } from 'shared/components/sets/measie-input.mjs'
+import { useTranslation } from 'next-i18next'
+
+export const ns = ['wbmeasies', ...inputNs]
+const emptyFunc = () => {}
```
joostdecock, line 5: See remark about making `startLoading` and `stopLoading` default to this in `MeasieInput`

sites/shared/components/workbench/views/measies/editor.mjs
```diff
@@ -0,0 +1,27 @@
...
+
+  const onUpdate = (m, newVal) => {
+    update.settings(['measurements', m], newVal)
+  }
+
+  return (
+    <div>
+      {Design.patternConfig.measurements.map((m) => (
+        <MeasieInput
+          {...{ t, m, mset, onUpdate, startLoading: emptyFunc, stopLoading: emptyFunc }}
```
joostdecock, line 19: See remark about  `startLoading` and `stopLoading` in `MeasieInput`

sites/shared/components/workbench/views/measies/index.mjs
```diff
@@ -42,13 +48,14 @@ export const MeasiesView = ({ design, Design, missingMeasurements, settings, upd
           <h5>We have all required measurements to create this pattern.</h5>
```
joostdecock, line 48: We can fix it later, but this needs translation

joostdecock (changes requested):

sites/shared/components/sets/measie-input.mjs
eriese: not copy/pasted. just a slip

sites/shared/components/sets/measie-input.mjs
eriese: as a user of imperial measurements, I'm not sure this is a correct distinction among symbols, and I don't see it being used anywhere else in our interface, including our `formatImperial` utility

PR #4321, opened 2023-06-20
refactor (sites) DRY headers, sticky side navs

eriese:
Just dealing with some UI things that were bugging me with nav and scrolling

- All sites now use a shared `LayoutWrapper`
    -  tracks scroll and controls hiding and showing the `Header` nav in a smoother way (it's necessary to do it in the layout wrapper so that the rest of the layout accommodates the movement of the header)
    - applies necessary top margin for presence/absence of header nav, so that it does not have to be handled in each layout
 - All sites now use a shared `HeaderWrapper`
     -   hiding and showing behavior is standardized
     -  position at different sizes is standarized
- Aside Nav and Workbench Header both use `position: sticky` to remain in view during scrolling. Side navs are stretched to full height, even if that exceeds the screen height
- Workbench Header Nav arrow bounces 5 times, then stops. Bounces infintely when the nav is hovered or open. Bounces 5 more times after the nav is closed. I just found it made me kind of seasick


sites/shared/components/workbench/header.mjs
```diff
@@ -59,7 +59,10 @@ const NavIcons = ({ setView, setDense, dense, view }) => {
         extraClasses="text-success bg-neutral hover:bg-success hover:text-neutral"
       >
         {dense ? (
-          <RightIcon className={`${iconSize} animate-bounce-right`} stroke={4} />
+          <RightIcon
+            className={`${iconSize} group-hover:animate-[bounceright_1s_infinite] animate-[bounceright_1s_5]`}
```
joostdecock, line 63: This is better. But we should probably stick this animation behind `motion-safe:` to be even better

joostdecock (approved):

PR #4327, opened 2023-06-21
feat(core): not methods to rotate/scale a snippet

joostdecock:


joostdecock:
will deal with this in #4320 instead

PR #4329, opened 2023-06-21
DRAFT: fix(bella): Prevent negative width front waist dart

BenJamesBen:
This fix sets the width of Bella's front waist dart to zero if the calculated width is a negative value and omits drawing the dart. (The dart points remain intact for inheriting designs (Noble, Bee) that might need them.)

### DIFF
--- a/designs/bella/src/front-side-dart.mjs
+++ b/designs/bella/src/front-side-dart.mjs
@@ -18,6 +18,7 @@ export const frontSideDart = {
     macro,
     utils,
     measurements,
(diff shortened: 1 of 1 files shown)


joostdecock:
This is a draft PR, but has not seen updates in 4 months or so.
@BenJamesBen what's the way forward here?

joostdecock:
Two months later, still in draft. I'm going to close this, but if someone wants to pick this up, that's fine of course.

PR #4330, opened 2023-06-21
(shared) Add curated sets back to set picker

eriese:
Also DRY up set candidate components and make the missing measies announcement smaller


eriese:
@joostdecock nudge

sites/shared/components/sets/set-candidate.mjs
```diff
@@ -35,32 +43,31 @@ export const SetLacksMeasies = ({ set, design, t, href, clickHandler }) => {
...
+  const setMeasies = set.measies && Object.keys(set.measies)
   // Quick check for required measurements
-  if (!set.measies || Object.keys(set.measies).length < requiredMeasies.length)
-    return <SetLacksMeasies {...setProps} />
+  if (!set.measies || setMeasies.length < requiredMeasies.length) hasMeasies = false
 
   // Proper check for required measurements
-  for (const m of requiredMeasies) {
-    if (!Object.keys(set.measies).includes(m)) return <SetLacksMeasies {...setProps} />
+  if (hasMeasies) {
```
joostdecock, line 62: Please use the `hasRequiredMeasurements` method in `sites/shared/utils.mjs` here.

sites/shared/components/workbench/views/measies/index.mjs
```diff
@@ -29,19 +30,21 @@ export const MeasiesView = ({ design, Design, settings, update, missingMeasureme
     <div className="m-6">
       <h1 className="max-w-6xl m-auto text-center">{t('measurements')}</h1>
       {missingMeasurements ? (
-        <Popout note>
+        <Popout note compact dense noP>
           <h5>{t('weLackSomeMeasies')}:</h5>
```
joostdecock, line 34: This (knowing what measurements are missing) is an often-requested feature by users, which was added.

And now you removed it, I guess because you want things to be more compact?
I think we need to strike a balance between the vast majority of users who will have 1, maybe 2 or at most a handful of sets and thus it's more important to give them good info than to be compact.
And then those people who have 12 sets because they've been using the site for years.

Either we set a cut-off point (more than x sets => compact display) or we hide more info when control = 5.

But I'm not in favor for removing something wholesale that is a requested feature by our users.

joostdecock (changes requested):

sites/shared/components/sets/set-candidate.mjs
joostdecock: Note: regarding me remarks below. This method will also return the missing measurements, which is what people want to see.

sites/shared/components/workbench/views/measies/index.mjs
eriese: didn't remove it, I put it in a well-labeled collapse element inside the same box.

Yes, I wanted the screen real estate back, but also, with our new flow every user will be redirected to the measurements view and be given this warning saying they are missing every measurement in the design every time they make a new pattern, and I thought that felt bad. (And that for patterns with a goodly number of measurements, it pushed all set choices below the fold).

I guess it was lazy because I didn't want to figure out a way to tell the difference between "I haven't entered any measurements" and "somehow I'm missing a few", though that answer is now obvious to me: check whether there's a measurements property in the settings, so I can do that instead.

I do wonder whether the request was definitely to list which measurements aren't in the settings or whether it was to specify which measurements each set is missing instead of just saying it doesn't have them

joostdecock: > *I do wonder whether the request was definitely to list which measurements aren't in the settings or whether it was to specify which measurements each set is missing instead of just saying it doesn't have them*

The latter. But there too, I think we should add a cut-off point. For example, if you're missing up to three measurements, we'll list them as: This misses A, B, and C.

If it's more, we'll just say: This misses several or something.

Obviously we need to strike a balance here. I'm not opposing changed, I'd just like to give the people what they want which is when they have a set of measurements but a pattern expects one that they haven't yet taken, call it out.
When they have no sets, or they are missing 8, obviously there's no need to list them all.

I also just sort of threw this together. For people who have many sets, or for our own curated set, a select (or drop-down) is probably preferable.

eriese: ok, so what's the requested change right now? I'm leaning towards

- don't warn about missing measurements if none have been set
- leave in the collapse element showing the missing measurements, but have it open by default as a stop gap on more complicated truncation?
- someday make this whole page tidier

PR #4331, opened 2023-06-21
feature (org) add pagination to blog and showcase pages

eriese:
Adds pagination to index pages that present sanity content. These pages no longer request sanity data server-side, which given the payload size I think is probably fine, and will be nice for not having to regenerate static pages when we publish new content.


eriese:
@joostdecock we're back to prebuilding the index pages, but this time paginated.

Two questions:
1. Is there anything I need to do with these pages for sitemap/rss purposes?
2. not sure it actually works, but for prebuilding all the post pages, do we want to try to do a single call to sanity and then find the right post ourselves, rather than dozens of individual queries at build time?

joostdecock:
@eriese 

1) you can add them to `sites/org/hooks/use-navigatiion.mjs` which is what breadcrumbs, sitemap and so on use to know the site structure 

2) One request might be slightly faster, but we'll probably have to deal with pagination/cursor. If it works, I'd ignore it for now

sites/org/components/sanity/utils.mjs
```diff
@@ -23,10 +24,19 @@ export const sanityLoader = ({ query, language, type, slug, order }) => {
     query += ` | order(${order})`
   }
 
-  return sanityClient.fetch(query)
+  query += filters
+
+  if (cache[query]) return cache[query]
+
+  console.warn('making a sanity request', query)
```
joostdecock, line 31: This seems to be a leftover debugging statement

sites/org/components/sanity/utils.mjs
```diff
@@ -23,10 +24,19 @@ export const sanityLoader = ({ query, language, type, slug, order }) => {
...
+  const result = await sanityClient.fetch(query)
+  cache[query] = result
+  return result
 }
 
 export const sanityImage = (image, dataset = 'site-content') => {
```
joostdecock, line 37: I don't think 1 method to load all Sanity images is all that great.
t's a minor thing, but I'd prefer to name this method based on the dataset it is loading the image from.
We will probably end up with a different way to retrieve site images vs user images, so calling it `sanityImage` hides too much complexity under the hood.

I'd prefer to have `sanitySiteImage` (and `sanityUserImage` in the future) because I suspect the way to load these will be very different, and frontend developers should not just assume these two are equal.

sites/org/next.config.mjs
```diff
@@ -6,6 +6,20 @@ import { jargon } from './jargon.mjs'
 
 let config = configBuilder({ site: 'org', jargon })
 config.i18n = i18nConfig.i18n
+config.redirects = async () => {
```
joostdecock, line 9: Can we use [a rewrite](https://nextjs.org/docs/pages/api-reference/next-config-js/rewrites) for this.
I'd prefer to keep the URL clean for page 1.

sites/org/pages/blog/page/[page].mjs
```diff
@@ -72,42 +73,59 @@ const Preview = ({ post, t }) => (
...
         locale,
         // title: 'Freesewing Blog',
-        path: ['blog'],
+        path: ['blog', 'page', params.page],
       },
     },
   }
 }
+
+const locales = ['', 'de', 'es', 'fr', 'nl']
```
Comment (joostdecock) [inline sites/org/pages/blog/page/[page].mjs:117]: Please load the locales from `site/site.config.mjs` and filter out `en` rather than maintain a list of locales here too.

joostdecock (changes requested):

sites/org/components/sanity/utils.mjs
eriese: added additional named methods below. It's true, I'm not clear on how to load the userImages, so it's a guess

eriese:
loading sanity content in the prebuild step is throwing me for a silly little loop, so that might not happen today, but in the mean time here's everything else fixed

PR #4341, opened 2023-06-24
(lab) UI tweaks from contributor call feedback

eriese:
During the contributor call, we looked over the ui and are proposing these tweaks:

- up and down carets on `Collapse` instead of no indicator when closed and an x when open
- `Collapse` bottom bar does not re-iterate collapse title, does display an up caret
- Larger toggles on closed MenuItems
- toggles and reset buttons are inside the collapse top bar element when the collapse is closed
- unusable reset buttons on open collapses should be disabled rather than hidden so that buttons don't shift position based on state
- patterns rendered in SVG should still be pan-zoomable
- test options should have visible radio buttons to indicate their function. No beaker icon


joostdecock:
I will merge this now to keep the peace, but I will revisit some of these because they're not all wanted or requested but I'm too busy to untangle this PR.

eriese:
You could tell me which ones to undo. As you say, if folks want something,
they've got to get it approved

joostdecock:
There's always the chance that I just hate chance. So let's use them as-is for a while and we'll see how it feels.
We're in alpha, might as well make use of that :)

eriese:
Fair enough. I do think you'll end you liking these things, and general
consensus among call participants was that these changes looked better and
made the functionality of ui elements easier to intuit

PR #4342, opened 2023-06-24
feature (lab) Modal Menus for Mobile

eriese:
Fully reworked after actually testing on mobile

This PR implements a context provider that allows components in the ui to register their contents as "menus" or "actions" that should be handled differently on smaller screens. Those components hide on small screens, and the `MobileMenubar` shows:

Each item registered as a "menu" will have a button that, when clicked, opens a modal displaying the registered content. Currently, the workbench nav and view menus are being registered as menus.

Each item registered as an "action" is shown in its entirety, and its styling is handled in its own css using media queries. Currently, the pattern zoom buttons are being registered as a single action.

Because everything gets consolidated into a single component for mobile, we're able to avoid the issue in my previous implementation where I was trying to position elements from different containers near one another in a cohesive way, and each menu had to handle its own styling and positioning. Now we have much cleaner sticky behavior on menu buttons, and changing the look, placement, or functionality of this feature can happen mostly in one file

The workbench nav is on the bottom, and its icon reflects the current view. I'm not totally sold on that vs a consistent button. I find it especially likely to be a problem on the save and export views
Screenshot_20230629_123401_Chrome

View menus open in modal. The close button is in a fixed position, but there's enough padding at the bottom of the modal to be able to be scroll to a place where it's not covering anything
Screenshot_20230629_123413_Chrome

Other exciting things: 

- I got rid of the `SizeMe` component, and svg sizing is now handled by html
- views that have a pattern and a menu now have a shared component they can use to handle their layout
- on view change, it scrolls back to the top, but if you had scrolled to hide the menu it preserves that state


joostdecock:
@eriese is this an actual WIP, or just an outdated title?

eriese:
I think I marked it WIP because it wanted some clean up, but I think that should be fixed now. We'll see what deepscan has to say. I guess the question is, is it what you had in mind, or near enough there?

eriese:
@joostdecock I changed everything about it and I think it's both better now and easier to redesign if you wanted something else

joostdecock:
Nice work @eriese :+1:

PR #4417, opened 2023-07-08
Parts renamed to stacks in layout reference docs

ibek:
After troubleshooting on discord we found a solution to own object layout in v3. It renamed parts to stacks in this object and some other pieces seem to require width and height so I added them to the example too.


joostdecock (approved):

joostdecock (approved):

joostdecock:
@all-contributors please add @ibek for docs

PR #4460, opened 2023-07-13
Added updated Aaron line drawing

nataliasayang:


artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
+import React from "react";
```
joostdecock, line 1: No need to import React when using JSX through Webpack, which is how we'll load this.

I realize that's hard to know but I'm adding these comments to explain :)

artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
+import React from "react";
+
+function Icon() {
```
joostdecock, line 3: We only use named exports in v3. We also want to allow people to override the styling, so the signature should be as follows:

export const Aaron = () => {
  // svg here
}

artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
+import React from "react";
+
+function Icon() {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="592.68"
```
joostdecock, line 7: Don't set height and width on the SVG so it will adapt to the parent container

artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
...
+function Icon() {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="592.68"
+      height="435.138"
+      version="1.1"
+      viewBox="0 0 156.813 115.13"
+    >
+      <g transform="translate(-124.226 -143.605)">
```
joostdecock, line 12: Please apply this transform.

artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
...
+        <path
+          fill="none"
+          stroke="#000"
```
joostdecock, line 15: The only color you can use is `currentColor`. So replace `#000` with `currentColor.

artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
...
+          strokeMiterlimit="10"
+          strokeWidth="0.265"
+          d="M159.604 258.598a215.4 215.4 0 0033.565-2.397v-74.798c-.5-.209-1.167-.518-1.918-.958-2.905-1.706-4.482-3.876-4.795-4.315-1.49-2.1-1.984-4.059-2.397-5.755-.635-2.616-.592-4.815-.479-9.11.019-.685.077-2.664.48-7.672.264-3.284.526-6.004.719-7.911l-8.63-1.918s-2.546 13.29-2.718 13.911c-.09.323-.116.567-.28.84-.228.38-.532.59-.699.698-6.899 4.08-21.715 3.685-27.376 0a2.16 2.16 0 01-.699-.699c-.161-.272-.19-.519-.28-.839-.172-.621-2.718-13.911-2.718-13.911l-8.63 1.918c.193 1.907.455 4.627.72 7.91.401 5.007.46 6.986.478 7.674.114 4.294.156 6.493-.479 9.11-.412 1.695-.907 3.65-2.397 5.754-.312.44-1.889 2.609-4.794 4.315-.752.442-1.418.75-1.918.958v74.798a215.376 215.376 0 0033.565 2.397z"
+          className="st2"
```
joostdecock, line 19: This className should go

artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
...
+      width="592.68"
+      height="435.138"
+      version="1.1"
+      viewBox="0 0 156.813 115.13"
+    >
+      <g transform="translate(-124.226 -143.605)">
+        <path
+          fill="none"
+          stroke="#000"
+          strokeMiterlimit="10"
```
joostdecock, line 16: This `strokleMiterlimit` should go

artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
...
+          strokeWidth="0.265"
+          d="M159.604 258.598a215.4 215.4 0 0033.565-2.397v-74.798c-.5-.209-1.167-.518-1.918-.958-2.905-1.706-4.482-3.876-4.795-4.315-1.49-2.1-1.984-4.059-2.397-5.755-.635-2.616-.592-4.815-.479-9.11.019-.685.077-2.664.48-7.672.264-3.284.526-6.004.719-7.911l-8.63-1.918s-2.546 13.29-2.718 13.911c-.09.323-.116.567-.28.84-.228.38-.532.59-.699.698-6.899 4.08-21.715 3.685-27.376 0a2.16 2.16 0 01-.699-.699c-.161-.272-.19-.519-.28-.839-.172-.621-2.718-13.911-2.718-13.911l-8.63 1.918c.193 1.907.455 4.627.72 7.91.401 5.007.46 6.986.478 7.674.114 4.294.156 6.493-.479 9.11-.412 1.695-.907 3.65-2.397 5.754-.312.44-1.889 2.609-4.794 4.315-.752.442-1.418.75-1.918.958v74.798a215.376 215.376 0 0033.565 2.397z"
+          className="st2"
+        ></path>
+        <path
+          fill="none"
+          stroke="#000"
```
joostdecock, line 23: Use `currentColor` as color

artwork/linedrawings/aaron.jsx
```diff
@@ -0,0 +1,182 @@
...
+          strokeWidth="0.265"
+          d="M176.148 143.764c-6.087 5.963-28.882 6.141-34.769 0"
+          className="st2"
+        ></path>
+      </g>
+    </svg>
+  );
+}
+
+export default Icon;
```
joostdecock, line 182: only use named export

joostdecock (approved):
I am going to use this as a starting point to lay the groundwork for the other linedrawings as JSX components.

We'll do something similar like we do for our icons, where we wrap them to handle the shared SVG stuff. I'll implement that and then let you know where to look for inspiration :)

Issue #4071, opened 2023-05-17
:open_umbrella: FreeSewing.org | v3 issues/todo


joostdecock:
This is an umbrella issue to capture all work that needs to happen on the v3 version of **FreeSewing.org** prior to a v3 release.

If you find something missing, you can comment on this issue and we'll add it.

### Postponed
- [ ] Add linedrawings in artwork as React components so they can be used in the pattern overview pages
- [ ] The new options pack thing
- [ ] Add QR code to PDF for long-ass URL
- [ ] Extend draft log view to handle more common errors
   - [ ] *The most common error that I see/make is when I try to use a Point/Path that doesn't exist because of a typo or slightly incorrect camelCase in the name I used.* (@BenJamesBen )
   - [ ] Passing a non-Point to something that expects a Point (like passing coordinated to `path.move()`)
- [ ] Scratch codes when enabling MFA
- [ ] Adding the several new patterns that are sitting in a PR

## Resolved
- [x] Support for bookmarking:
  - [x]  sets
  - [x] patterns
  - [x] docs
- [x] Curated sets picker & bookmarked set picker
- [x] Add cover art, plus links to design documentation and showcase to Design pages (no cover art in v3)
- [x] Improve paperless view in dark mode (and perhaps other themes, as the draft themeing is somewhat different from the rest of the site)
- [x] Extend designs software config file with techniques based on work done in #880 and add to filtering in design list
- [x] The recent changes to the header have caused the offset of anchor links to be incorrect. 
- [x] Design filtering
- [x] Hide set checkbox when there are no sets
- [x] Implement newsletter signup pages
- [x] The macro work Wouter did in #3998 
- [x] Add the waist to armpit measurement and adapt various patterns to use it
- [x] Make annotations store IDs in store
- [x] Make 'complete' stuff more automatic (not shown when complete is off, with option to override)
- [x] Make title plugin create a map of partname/nr so we can show the nr in the included parts list
- [x] Implement translation method in plugin-i18n (for people who just want to pass in data)
- [x] Implement OIDC/login with Google
- [x] Implement OIDC/login with GitHub
- [x] Inline docs don't have custom components enabled / or they are broken (fixed by @eriese in #4864 )
- [x] Accessible colors in all themes
- [x] Implement changes in #4255 for all designs and adapt UI to handle translation of designs
- [x] text labels on navigation header are not shown on smaller window sizes (reported by @BenJamesBen wnd fixed by @joostdecock )
[...]

PR #4480, opened 2023-07-16
(WIP) Build Time and Bundle Size Improvements

eriese:
This needs polish and comments, which I will add in the morning. Just wanted to provide some visibility into what I'm working on.

Doesn't work in dev.


sites/org/pages/[postType]/page/[page].mjs
```diff
@@ -0,0 +1,206 @@
+// Dependencies
```
Comment (joostdecock) [inline sites/org/pages/[postType]/page/[page].mjs:1]: Here too, the shared code seems to add more friction.
Like for one thing, I don't think we need pagination for the blog.

sites/org/pages/docs/[...slug].mjs
```diff
@@ -1,191 +1,7 @@
-// Used in static paths
-import { mdxPaths } from 'site/prebuild/mdx-paths.en.mjs'
-// Dependencies
+import { DocsPage, ns as pageNs } from 'shared/components/docs/docs-page.mjs'
```
Comment (joostdecock) [inline sites/org/pages/docs/[...slug].mjs:1]: Same remark. Too abstract. Let pages be pages

sites/org/pages/docs/[...slug].mjs
```diff
@@ -219,9 +38,7 @@ export async function getStaticProps({ locale, params }) {
  * To learn more, see: https://nextjs.org/docs/basic-features/data-fetching
  */
 export async function getStaticPaths() {
-  const somePaths = mdxPaths
-    .filter((path) => path.split('/').length < 5)
-    .filter((path) => path !== 'docs')
+  const somePaths = mdxPaths.filter((path) => path.split('/').length < 5 && path !== 'docs')
```
Comment (joostdecock) [inline sites/org/pages/docs/[...slug].mjs:41]: I feel all docs should be pre-built.

sites/org/pages/docs/[...slug].mjs
```diff
@@ -1,191 +1,7 @@
...
-import Head from 'next/head'
-import { PageWrapper, ns } from 'shared/components/wrappers/page.mjs'
-import { Spinner } from 'shared/components/spinner.mjs'
-import { components } from 'shared/components/mdx/index.mjs'
-import { MdxWrapper } from 'shared/components/wrappers/mdx.mjs'
-import { Toc } from 'shared/components/mdx/toc.mjs'
-import { DocsLayout } from 'site/components/layouts/docs.mjs'
-
-/*
- * PLEASE READ THIS BEFORE YOU TRY TO REFACTOR THIS PAGE
```
Comment (joostdecock) [inline sites/org/pages/docs/[...slug].mjs:17]: Did you read this?
Because you're reverted the way we load MDX to the v2 way, which means markdown changes are no longer picked up for fast-refresh.

We're essentially back to where we were 3 months ago when it comes to loading MDX.
I don't understand why you would unilaterally implement such a significant change that essentially rolls back a bunch of work that's been done.

joostdecock (changes requested):
There's good stuff in here, but too many things that are just not wanted.
Like the way MDX is loaded and the abstractions of pages across different sites and content types.

Might be better to discuss these sort of things as they are rather significant changes that go against work that's been done before.

sites/backend/scripts/prebuild-i18n.mjs
```diff
@@ -6,7 +6,7 @@ import yaml from 'js-yaml'
 /*
  * List of supported languages
  */
-const locales = ['en', 'es', 'de', 'fr', 'nl']
+// const locales = ['en', 'es', 'de', 'fr', 'nl']
```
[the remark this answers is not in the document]
eriese: I was trying to reduce the number of places where we have to manually add languages. I believe I did this because with this hard coded array, Ukrainian is not being prebuilt and then the site build fails, but I can double check

sites/dev/pages/[...slug].mjs
```diff
@@ -1,100 +1,7 @@
-// Used in static paths
+import { DocsPage, ns as pageNs } from 'shared/components/docs/docs-page.mjs'
```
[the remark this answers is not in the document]
Comment (eriese) [reply sites/dev/pages/[...slug].mjs:1]: Of course I feel the opposite way. This code was basically a duplicate of the one in org, so when I went in to update it, I felt it was easier to maintain consistency if changes only happened in one place. It's true that this is the reason dev won't build now, because dev doesn't have a docs layout, but it's my opinion that that layout should be shared too.

That having been said, I will try to do less abstraction. I may need help understanding where that line is for you

sites/org/components/posts/page-wrapper.mjs
```diff
@@ -0,0 +1,74 @@
+import { PageLink } from 'shared/components/page-link.mjs'
```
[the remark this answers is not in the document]
eriese: It should be trivial to add `'newsletter'` to the array of `postTypes ` and then we have newsletter pages too. I just didn't implement it because newsletters aren't accounted for in the nav so I figured we didn't serve them this way.

I do understand that `postType` is not as easy to reason about as `blog` or `showcase`. I'm guessing when you say "shared components", though, you don't mean to let me import the entire page component and also the server-side functions?

sites/org/hooks/use-navigation.mjs
```diff
@@ -169,43 +171,50 @@ const sitePages = (t = false, control = 99) => {
...
   const { account } = useAccount()
 
-  const siteNav = {
-    ...pbn[locale],
-    ...sitePages(t, ignoreControl ? undefined : account.control),
-  }
-  for (const [_path, _data] of extra) {
-    objUpdate(siteNav, _path, _data)
-  }
+  const [navigation, setNavigation] = useState({
```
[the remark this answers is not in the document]
eriese: We could move it to its own context, but that felt wasteful to me. What I found was that we were calling `useNavigation` in lots of places where we wanted the same (massive) structure that almost never changes, and it's not cached so it was being calculated every time. You definitely helped with that when you refactored `ReadMore`, but there were still plenty of places where it made sense to me to say "we already made that data, let me access the existing one". In most cases, the `NavigationContext` was already also being used, usually to get, rather than set navigation state, and it was already calling `useNavigation` internally

sites/org/pages/[postType]/page/[page].mjs
Comment (eriese) [reply sites/org/pages/[postType]/page/[page].mjs:1]: Why not? I think uninterrupted flow is always nice, and it costs very little to have it there

sites/org/pages/docs/[...slug].mjs
Comment (eriese) [reply sites/org/pages/docs/[...slug].mjs:1]: Yeah, I'm definitely going to need help understanding what the correct level of abstraction is. All docs pages should look and work the same, I feel?

sites/org/pages/docs/[...slug].mjs
Comment (eriese) [reply sites/org/pages/docs/[...slug].mjs:41]: They are. I just combined your two existing filter calls into one filter call

sites/org/pages/docs/[...slug].mjs
Comment (eriese) [reply sites/org/pages/docs/[...slug].mjs:17]: I did, but I focused more on the don't DRY it up part than the don't load from filesystem part. I went back and forth on dynamic loading vs filesystem loading. All the docs files need to be compiled and bundled anyway because we load them dynamically in the workbench, so there's not a big build time saving from doing it this way instead, but in terms of bundle size and load time, as well as compile time on dev, it's major. 

I wonder if we might consider using a [watch plugin](https://github.com/hashicorp/next-remote-watch) to fix the hot-reload issue.

joostdecock:
> I wonder if we might consider using a [watch plugin](https://github.com/hashicorp/next-remote-watch) to fix the hot-reload issue.

My concern is that you're over-engineering things that we're not asking for and your response is to over-engineer it more. Which makes me feel like we're not on the same page, perhaps not even on the same chapter.

After v3 is out and we've all had a rest, we can discuss optimization and I'm happy to listen to why you feel the work I did to migrate MDX loading to webpack needs to be undone.

Until such time, I will focus on [what we agreed needs to be done](https://github.com/freesewing/freesewing/issues/4071) and will refrain from getting involved in side-quests. Which is in line with [what I announced tot he community earlier this month](https://discord.com/channels/698854858052075530/757634186398924910/1119631364396941322).

Apologies is that prioritization does not align with yours, but I feel it's what's required at this time.

PR #4546, opened 2023-07-23
Refactor translation file structure to fix filename issues with Crowdin

joostdecock:
I feel like I've spent most of these last couple of weeks fighting Crowdin, but we're not out of the woods yet.
The problem is that the namespace structure we use for React components is causing issues.
For example:

```
sites/shared/popout.en.yaml
```

This file will be put back by crowding as:
```
sites/shared/popout.en.nl.yaml
sites/shared/popout.en.fr.yaml
...
```

Which is not great. So I've restructured things so that the files are all `en.yaml` like this:

```
sites/shared/popout/en.yaml
```

This file will be put back by crowding as:
```
sites/shared/popout/nl.yaml
sites/shared/popout/fr.yaml
...
```

Which is what we want.


joostdecock:
Merging this without waiting so crowdin can sync over night (it takes time because it keeps getting rate-limited by github)

PR #4599, opened 2023-07-28
fix (shared) use node conditional exports instead of sitebuildconfigure

eriese:
If my understanding of  https://github.com/freesewing/freesewing/commit/009e1c1cc93c76bc860970a62249e8f55d8334b3 is correct, the goal is to change the entry points  on all our packages to `src/index.mjs`  when (pre)building sites *on vercel* but leave them pointing to `dist.index.mjs` the rest of the time.

Instead of rewriting `package.json`s, we can use [node --conditions](https://nodejs.org/api/cli.html#-c-condition---conditionscondition) to switch to a different, always-defined entry point.

This PR implements the following `exports` property in the `package.json`s
```json
"exports": {
    ".": {
      "internal": "./src/index.mjs",
      "default": "./dist/index.mjs"
    }
  },
  ```

this allows us to not have to run `reconfigure` at all on site builds (it shouldn't be needed. everything should be up to date), and to run site prebuild scripts using `--conditions=internal --experimental-json-modules ./prebuild.mjs`. The same script can now be used for both local and ci builds. We also don't have to build dependencies before testing, so I removed that from nx

### DIFF
--- a/config/scripts.yaml
+++ b/config/scripts.yaml
@@ -13,7 +13,7 @@ _types:
   design:
     prettier: "npx prettier --write 'src/*.mjs' 'tests/*.mjs'"
     test: &test 'npx mocha tests/*.test.mjs'
-    testci: &testci 'npx mocha tests/*.test.mjs --reporter ../../tests/reporters/terse.js'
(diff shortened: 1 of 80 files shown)


joostdecock (approved):
Nice. I like this a lot 👌

eriese:
Thought you might! Probably means we should update `kickstart` to not bother with `buildall`

PR #4638, opened 2023-08-02
Ensuring light and dark theme pass ADA requirements for contrast

SeaZeeZee:
This PR ensures that the light and dark themes contain colors that *can* contrast with other colors with at least a 7, and that elements on the sites have the correct styles applied to use those colors in ways that *will* pass ADA requirements for contrast

It also implements a new method of adding our own custom named colors to our themes, which allows us to have `spectrum` and `rating` colors work in the same way as `primary`, `secondary` and `accent`


joostdecock (commented):
Plenty of changes here for legislation that doesn't apply.

Care to clarify whether there is actually a problem? And if so why I couldn't be addressed by updating a theme?

eriese:
It's true that the ADA wouldn't apply to our sites even if the project were based in the states, but it is considered the bare minimum standard for accessibility and is kind of a shorthand here for easy to find benchmarks. Members of the freesewing community have indicated that better visual accessibility would be helpful to them; and even if none had, having a site with low visual accessibility can undermine trust in our stated goal of inclusivity. 

Updating the theme was the first step, and it can help make it so that the available colors are capable of having accessible levels of contrast, but it won't guarantee that elements have those colors applied in ways that actually provide accessible levels of contrast. It's still possible to have a dark color against another dark color if that's how it's styled. So the next step was actually going through the interface and making sure that elements are styled such that accessible themes will yield accessible interfaces. This required changing the styles applied to a number of components throughout the UI.

A more detailed breakdown of what's happening in this PR:
- update colors in light and dark themes to ones with adequate contrast against themed background colors
- fix places where styling made for low contrast
  - limit use of `btn-ghost` because it's hard to see the hover state
  - change workbench left and top menus to a light background so that themed colors can have adequate contrast against them
- limit number of colors used within a single component to minimize visual overwhelm
- remove un-themed colors in favor of themed colors
- add `spectrum` and `ratings` colors to themes via tailwind
- remove use of `useTheme` to import themed `spectrum` and `ratings` colors in favor of themed class names
- fix re-render issue causing the `HowDoesItWorkAnimation` to be super fast. Slow it down even more than that
- add focus outlines back to text inputs
- change workmark to invert all letter colors on hover (check it out in lgbt theme where there's still a rainbow. it's super cute)

eriese:
@joostdecock gentle nudge

joostdecock:
> @joostdecock gentle nudge

I'm not comfortable to merge this as-is, and too busy right now to tease it apart. So you'll have to wait.

eriese:
Fair enough. Let me know if there's anything I can do to  help rough the teasing

joostdecock:
So this PR. As I mentioned, I'm not super stoked with it. For one thing, the PR template states not to go on and refactor a bunch of stuff without running it by me first, and I put that there for a reason. The reason being that I don't  want to receive these kind of PRs.

I am aware that things are not perfect now. It's alpha after all. But it feels like you're trading one set of problems for another.
I clicked around a bit and took screenshots of things that imho are problematic:

This one is subjective but damn, those colors are not pretty together:
Screenshot from 2023-08-18 18-46-23

The hover on these footer icons makes them dissapear unless you hover them just right (when I point out something like this it's a problem that was not there before, but was introduced by this PR):
Screenshot from 2023-08-18 18-47-29

For a PR that is all about contrast, it's not very good at contrast:
Screenshot from 2023-08-18 18-49-29

I don't know what that border below the icon is about but it shouldn't be there:
Screenshot from 2023-08-18 18-51-40

While specific components can be tweaked, it's not great that the basic *success* button is illegible in the dark theme. Especially because that's sort of the point of this entire PR:
Screenshot from 2023-08-18 18-57-44

This is the main menu and I find it lacking in contrast:
Screenshot from 2023-08-18 18-58-31

Once again, there are issues not too. So it's not bad, it feels more like a lateral move that adds complexity.
And there are objectively good things too. I think the dark theme in particular is nicer, perhaps whoever worked on this prefers dark mode unlike myself who doesn't pay much attention to it. Perhaps one could submit that as a new theme, and then we can get some testers to check it out and potentially make it the new dark theme. That would be a no-brainer to merge too.

So while I do believe there are good things in this PR, it's unfortunate that it's bundled up with a bunch of other stuff and complexity that I am not a fan of in this take-it-or-leave it PR.

So if that's the choice. thank you but I'll leave it.

eriese:
Joost, I'm really sorry that this PR made you feel like you had to take it or leave it. Accessibility is a high priority to me, and as I said I'm glad to tease this apart if you're willing to consider accepting some of it in smaller bites. This PR definitely had some scope creep, so I think it could be separated into 4 concerns that I am invested in:

1) implement light and dark themes that use colors with adequate contrast
2) restyle components to better use theme colors for adequate contrast 
3) (or 2a) restyle components for less visual overwhelm and general aesthetics
4) keep client-side javascript out of theme implementations

Would you be open to discussing which (if any) of those goals you would support?

joostdecock:
I think 1) can happen today, although I would prefer to see a PR with alternatives for a light and/or dark theme. Eg aktlight and/or altdark.

That would be a no-brainer to merge because it would allow us to use the different themes on various devices/scenarios and when we decide one is better get rid of the other one.

For 2) or 3) I think that's an important part of beta testing and thus today too soon to do so. Eg. A lot of this PR was about the rainbow thingie which has been removed in all but the lgbtq theme. So that illustrates that it's probably too soon to start debating the colors of the walls when we're still building the rooms (so to speak).

So yes. But I think we should wait for beta.

As for 4) I don't see how that relates to accessibility. FreeSewing's is a client-side SPA. No client-side Javascript, jo FreeSewing. So that makes me suspect I'm perhaps not understanding the driver for that change.


So to summarize: 

- Other/better themes: Yes please, let's see what you come up with
- Checking that all components look ok (because even the best themes don't guard against bad choices): yes but let's wait for beta (reminder that we're in alpha)

SeaZeeZee:
Could you elaborate on what you mean by components looking ok? Do you mean aesthetically pleasing, or something else?

eriese:
> I think 1) can happen today, although I would prefer to see a PR with alternatives for a light and/or dark theme. Eg aktlight and/or altdark.

Yes, I'm fine to submit them as alternative themes.

> For 2) or 3) I think that's an important part of beta testing and thus today too soon to do so.

Ok, I can see that point. In earlier conversations I got the impression that you wanted the UI looking really good before beta for a big reveal feeling. I'm hearing that that's not feeling like a blocker actually, and I can get on board with setting that as a beta priority rather than an alpha

> A lot of this PR was about the rainbow thingie which has been removed in all but the lgbtq theme. So that illustrates that it's probably too soon to start debating the colors of the walls when we're still building the rooms (so to speak).

Very little about this PR was about the *colors* of the rainbow thingie, but a lot of this PR was about the *implementation* of having a rainbow thingie, which is concern 4. I described it badly, and you're right that it doesn't have to do with accessibility at all, it was just there while I was working with things that used it.

What I mean by client-side javascript in theme implementations is the way that `spectrum` and `ratings` are exported from each theme file and then used by the `useTheme` hook to provide color names directly to components that use them. That pattern, to me, undermines the utility of a theme library.  It's a workaround for the fact that daisyui doesn't allow us to configure additional named colors, but I prefer [my workaround](https://github.com/SeaZeeZee/freesewing/blob/3e54d1f14dca2bfa01b8b1ac82c63f9c3b9270cf/sites/shared/config/tailwind.config.mjs#L3), which  puts them into tailwind the way daisyui does under the hood

The benefits as I see them are:

- more convenient control over our extra colors additional themed colors because we can add as many as we want without having to remember additional exports
- we can use theme-agnostic semantic classnames, in line with how all other theming works on the site
- we can easily use theme inheritance for these colors
- bundle size improvement because we don't have to bundle theme data in with our client-side javascript
- memory use improvement because all that data can't be garbage collected while it might be used
- components that use those classes can just use them without additional hooks and imports
- fewer components will treat a theme change as a state change.

PR #4764, opened 2023-08-18
Fix import errors for ModalThemePicker and ModalLocalePicker

compilin:
On the current develop branch, trying to run the lab results in the following error:
```
- error ./components/layouts/lab.mjs:1:0
Module not found: Can't resolve 'shared/components/theme-picker/index.mjs'
> 1 | import { ThemePicker } from 'shared/components/theme-picker/index.mjs'
  2 | import { LocalePicker } from 'shared/components/locale-picker/index.mjs'
```

Because those two modules have been moved and renamed. So here's a fix that points those imports to the right place, in every file I found that referenced them

### DIFF
--- a/packages/new-design/shared/lab/components/header.mjs
+++ b/packages/new-design/shared/lab/components/header.mjs
@@ -1,7 +1,7 @@
 import { useState, useEffect } from 'react'
 import Link from 'next/link'
-import { ThemePicker } from 'shared/components/theme-picker.js'
-import { LocalePicker } from 'shared/components/locale-picker.js'
+import { ModalThemePicker } from 'shared/components/modal/theme-picker.mjs'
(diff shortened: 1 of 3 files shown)


eriese:
Thanks so much for submitting! Looking at these files I realize they're way behind where they should be in general, so I just want to check: with these fixes, does the lab compile and run as expected, or are there new errors when you try to navigate things?

joostdecock (approved):

joostdecock:
The lab is very much neglected/broken right now and until things stabilize for the .org site we should probably to the minimum to get it to work.

So while this will almost certainly no fix all issues, it's defo better. So I'm merging it as-is.

joostdecock:
@all-contributors please add @compilin for code

compilin:
The lab does work with that fix. The signup page loads without issue. Haven't checked whether the changes on the new-design package also work though

PR #4801, opened 2023-08-23
feat(org): Work in measurments sets, api keys, bookmarks, and account pages

joostdecock:
This is (the start of) work to make things a bit more consistent rather the 'merely working'.

For one thing, uses inputs consistently and ensures there is always documentation for what is expected.


sites/shared/components/account/sets.mjs
```diff
@@ -110,495 +109,486 @@ export const NewSet = ({
...
+  const [filter, setFilter] = useState(false)
+  const [edit, setEdit] = useState(false)
+  const [mset, setMset] = useState()
+  // Set fields for editing
+  const [name, setName] = useState(mset?.name)
+  const [image, setImage] = useState(mset?.image)
+  const [isPublic, setIsPublic] = useState(mset?.public ? true : false)
+  const [imperial, setImperial] = useState(mset?.imperial ? true : false)
+  const [notes, setNotes] = useState(mset?.notes || '')
+  const [measies, setMeasies] = useState({})
```
eriese, line 200: personally would prefer to save the original set in the state as, like, `origSet`, and a deep copy to manipulate in a single state or maybe even ref object `mset`. then just to write an onChange callback that sets each of these properties on the `mset` so you don't have to store and mess with them all separately. Then in the `save` callback, you loop through all the properties to compare `origSet` to `mset` to send the changes

sites/shared/components/account/sets.mjs
```diff
@@ -110,495 +109,486 @@ export const NewSet = ({
...
-    >
-      <EditMeasurementsSet
-        {...{ t, mset, account, backend, toast, refresh, setModal, startLoading, stopLoading }}
-      />
-    </Collapse>
-  )
-}
-
 // Component for the account/sets page
 export const Sets = ({ title = true }) => {
```
eriese, line 579: if these components don't rely on one another, I think it'd make sense for them to be in separate files because this file is overwhelmingly large

sites/shared/components/inputs.mjs
```diff
@@ -0,0 +1,424 @@
...
+}) => {
+  const { setModal } = useContext(ModalContext)
+
+  return (
+    <div className="form-control w-full mt-2">
+      <label className={`label pb-0 ${docs ? 'hover:cursor-help' : ''}`}>
+        <span className="label-text text-lg font-bold mb-0">{label}</span>
+        {docs ? (
+          <span className="label-text-alt">
+            <button
```
eriese, line 61: this is where that `cursor-help` should go. doesn't need to be `hover` because the cursor only shows on hover anyway

sites/shared/hooks/use-apikey-docs.mjs
```diff
@@ -0,0 +1,30 @@
...
+        import(
+          /* webpackInclude: /docs\/site\/apikeys\/[a-z]+\/[a-z][a-z]\.md$/ */ `markdown/docs/site/apikeys//${option}/${locale}.md`
+        ),
+      [locale]
+    )
+
+  const docs = {}
+
+  for (const option of ['name', 'expiry', 'level']) {
+    const { MDX, frontmatter } = useDynamicMdx(loader(option))
```
eriese, line 20: `useDynamicMdx` may be a a bad name for that hook. It is meant for use with mdx that we're dynamically loading *on page load*, so it's technically dynamic, but it's not wrapped in `next/dynamic` because it wasn't really necessary for the implementation. 

These docs hooks really want to lazy-load this markdown, so for that I recommend taking a look at [the dynamic org docs implementation](https://github.com/freesewing/freesewing/blob/f0721700fdcf477d42440db2fc5d8d43e10c1d2c/sites/shared/components/dynamic-docs/org.mjs#L25), though maybe we should unify the approach

eriese (commented):
Some thoughts, no demands

sites/org/pages/account/apikeys/[id].mjs
```diff
@@ -0,0 +1,88 @@
...
+    },
+  }
+}
+
+/*
+ * getStaticPaths() is used to specify for which routes (think URLs)
+ * this page should be used to generate the result.
+ * To learn more, see: https://nextjs.org/docs/basic-features/data-fetching
+ */
+export const getStaticPaths = async () => ({ paths: [], fallback: false })
```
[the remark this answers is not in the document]
Comment (joostdecock) [reply sites/org/pages/account/apikeys/[id].mjs:88]: I was like *sure* and removed it in 8da064398e1966451a631c1034f7649823ada6c3

Then, the next day, this:
image

Which made me cranky because I am trying to ship this thing and instead we're spending our time on these side-quests.

Did I need to get cranky? Of course not. But I'm spread very thin so until v3 is out, I will just say **no side-quests** whenever I come across something like this. 
It means that it may be a good or even great idea, but it's not a blocker so let's just not until v3 is out the door.

sites/shared/hooks/use-bookmark-docs.mjs
```diff
@@ -0,0 +1,30 @@
...
+import { useDynamicMdx } from 'shared/hooks/use-dynamic-mdx.mjs'
+// Components
+import { MdxWrapper } from 'shared/components/wrappers/mdx.mjs'
+
+export const useBookmarkDocs = (locale) => {
+  const loader = (option) =>
+    useCallback(
+      () =>
+        import(
+          /* webpackInclude: /docs\/site\/bookmarks\/[a-z]+\/[a-z][a-z]\.md$/ */ `markdown/docs/site/bookmarks/${option}/${locale}.md`
```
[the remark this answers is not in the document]
joostdecock: I have changed it to `orgmarkdown` as I guess it's better to be explicit.
I've kept the magic comment though because it doesn't hurt and I think it's better to have this here so people looking at the code appreciate it exists.

sites/shared/components/account/sets.mjs
joostdecock: no side-quests

sites/shared/components/account/sets.mjs
joostdecock: no side-quests

sites/shared/components/inputs.mjs
joostdecock: I tried it out to see how it would feel, and I quickly started disliking it. So I threw it out.

sites/shared/components/inputs.mjs
```diff
@@ -0,0 +1,424 @@
...
+              }
+            >
+              <DocsIcon />
+            </button>
+          </span>
+        ) : null}
+      </label>
+      {children}
+      {labelBL || labelBR ? (
+        <label className="label">
```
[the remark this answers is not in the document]
joostdecock: I have re-worked the FormControl component so that you can pass an `id` prop to the various inputs.
If you do so, the FormComponent will link the label with the input with id/for.
If you do not, it will render a div instead of a label.

While I did all the work to make it possible, I have not (yet) passed an ID to all inputs. But it's at least possible now.

sites/shared/hooks/use-apikey-docs.mjs
joostdecock: My focus in these changes was on presenting a more consistent experience to the user when inputting data.
So how we load the MDX was not something I was into. I just want to make sure that ever input has contextual (modal) help.

As per your suggestion, I have removed these various hooks and instead just use the dynamic docs component.
However, it seems that the custom components are not working. Which is odd because I'm pretty certain they did work.

Anyway, this might require another look but for now I'm leaving it as is.

sites/org/pages/account/bookmarks/[id].mjs
```diff
@@ -0,0 +1,87 @@
...
+    },
+  }
+}
+
+/*
+ * getStaticPaths() is used to specify for which routes (think URLs)
+ * this page should be used to generate the result.
+ * To learn more, see: https://nextjs.org/docs/basic-features/data-fetching
+ */
+export const getStaticPaths = async () => ({ paths: [], fallback: false })
```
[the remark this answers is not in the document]
Comment (joostdecock) [reply sites/org/pages/account/bookmarks/[id].mjs:87]: no side-quests

sites/shared/hooks/use-set-docs.mjs
```diff
@@ -0,0 +1,30 @@
...
+import { useDynamicMdx } from 'shared/hooks/use-dynamic-mdx.mjs'
+// Components
+import { MdxWrapper } from 'shared/components/wrappers/mdx.mjs'
+
+export const useSetDocs = (locale) => {
+  const loader = (option) =>
+    useCallback(
+      () =>
+        import(
+          /* webpackInclude: /docs\/site\/sets\/[a-z]+\/[a-z][a-z]\.md$/ */ `markdown/docs/site/sets//${option}/${locale}.md`
```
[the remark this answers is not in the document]
joostdecock: See above

sites/shared/hooks/use-measurement-docs.mjs
```diff
@@ -0,0 +1,32 @@
...
+// Components
+import { MdxWrapper } from 'shared/components/wrappers/mdx.mjs'
+import { measurements } from 'config/measurements.mjs'
+
+export const useMeasurementDocs = (locale) => {
+  const loader = (m) =>
+    useCallback(
+      () =>
+        import(
+          /* webpackInclude: /docs\/measurements\/[a-z]+\/[a-z][a-z]\.md$/ */ `markdown/docs/measurements/${m}/${locale}.md`
```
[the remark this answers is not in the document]
joostdecock: Same here

sites/shared/hooks/use-apikey-docs.mjs
eriese: understood. I can take a look if that would help

sites/org/pages/account/apikeys/[id].mjs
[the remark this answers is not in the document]
Comment (eriese) [reply sites/org/pages/account/apikeys/[id].mjs:88]: agreed, not important at this juncture

(when we approach it in future, we'll replace `getStaticProps` with `getServerSideProps` in addition to removing `getStaticPaths`)

PR #4849, opened 2023-08-29
fix(core): fixed buggy bounds computation in cutting and printing layout when parts are rotated & fix(dev): fixed invalid PageLink imports

compilin:
A bug in the matrixTransform utils led to rotations being applied around (0, 0) if the X coordinate of the center was 0, leading to the cutting and printing layouts incorrectly calculating the outer bounds of the pattern.
The transformation parsing was also implemented a second time in applyTransformToPoint, in which the rotation center wasn't considered at all, so I replaced that second redundant implementation with a call to the first one.

I also stumbled upon the dev site not being buildable because of some import issues so I put that fix in there too.

### DIFF
--- a/packages/core/src/utils.mjs
+++ b/packages/core/src/utils.mjs
@@ -803,7 +803,7 @@ function matrixTransform(transformationType, matrix, values) {
       const centerY = values[2]
 
       // if there's a rotation center, we need to move the origin to that center
-      if (centerX) {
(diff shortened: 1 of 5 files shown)


joostdecock (approved):

joostdecock:
Much like. Thank you.

@all-contributors please add @compilin for code

PR #5022, opened 2023-09-23
feat: Uma

### LINKED ISSUE #5002: [uma]: Prepare for v3 beta

joostdecock:
Closes #5002


BenJamesBen:
My understanding of the situation was that previously there were issues regarding Ursula and Unice, but that the issues had since been resolved between the pattern designers and that Unice was not going to be removed from the repository. If this is the case, then it appears that Unice might have been accidentally removed from the repository, via this PR's commit https://github.com/freesewing/freesewing/pull/5022/commits/45e76dd95827be35db1b7400a134786a79fdc73e.

I've added this topic to the September 30, 2023 community call https://github.com/freesewing/freesewing/discussions/5027.

PR #5165, opened 2023-10-16
gozer

woutervdub:


designs/gozer/src/ghost.mjs
```diff
@@ -0,0 +1,96 @@
...
+    points.topCp1 = points.top.shift(180, size * c)
+    points.topCp2 = points.top.shift(0, size * c)
+    points.leftCp1 = points.left.shift(270, size * c)
+    points.leftCp2 = points.left.shift(90, size * c)
+    points.bottomCp1 = points.bottom.shift(0, size * c)
+    points.bottomCp2 = points.bottom.shift(180, size * c)
+    points.rightCp1 = points.right.shift(90, size * c)
+    points.rightCp2 = points.right.shift(270, size * c)
+
+    points.eyeLefttop = points.eyeLeft.shift(90, size * eyeSize)
```
joostdecock, line 27: Regardless of the remark about circles above, I suspect this might be a mistake.

You are multiplying `size * eyeSize` here, but I think just `eyeSize` is what's intended.

If nothing else, when I fire up the lab and run this pattern, I don't see anything. That's because the eyes are (a lot) bigger than the entire body, so the pattern gets so large that the lines get vanishingly small.

designs/gozer/src/ghost.mjs
```diff
@@ -0,0 +1,96 @@
...
+    const c = 0.55191502449351
+    const eyeSize = measurements.head * 0.0416
+    const size =
+      measurements.hpsToWaistBack + measurements.waistToFloor + measurements.head / Math.PI
+
+    points.middle = new Point(0, 0)
+    points.eyeLine = points.middle.shift(270, measurements.head / Math.PI / 2)
+    points.eyeLeft = points.eyeLine.shift(180, measurements.head * 0.13)
+    points.eyeRight = points.eyeLine.shift(0, measurements.head * 0.13)
+    points.top = new Point(0, -1 * size)
```
joostdecock, line 14: I am trying to see a reason why you're drafting this by hand when this will do the same and is more than twice as fast to draft:

```js
    points.middle.addCircle(size)
    points.eyeLeft.addCircle(eyeSize)
    points.eyeRight.addCircle(eyeSize)
```

Is there a specific reason I'm missing, or was `Point.addcircle()` just not on your radar?

joostdecock (changes requested):

joostdecock:
To further clarify my comments @woutervdub this is what the design looks like as-is now:

Screenshot 2023-10-17 at 09-28-14 Gozer

If you inspect the elements of the SVG, you can see that the seam outline is but a small part of the pattern:

Screenshot 2023-10-17 at 09 29 45

It is the eye(s) that's taking up most of the space:

Screenshot 2023-10-17 at 09 30 00

This is what it looks like when I use circles as in my comment:

Screenshot 2023-10-17 at 09-27-49 Gozer

It also drafts significantly faster. With Cubic beziers:

```
 73 passing (74ms)
 ```
 
 with circles:
 
 ```
 73 passing (34ms)
 ```
 
 So unless I'm missing something, I'd prefer to see use use circles to draft this.

File: designs/gozer/src/ghost.mjs (hunk 1 above)
Comment (woutervdub) [reply designs/gozer/src/ghost.mjs:27]: You're right, and I'm an idiot.

File: designs/gozer/src/ghost.mjs (hunk 2 above)
Comment (woutervdub) [reply designs/gozer/src/ghost.mjs:14]: I was totally unaware of that method. 

Thanks!

Comment (woutervdub) [reply designs/gozer/src/ghost.mjs:14]: Still keeping the self-drawing circle for the fabric itself, so I can do the `sa` normally

**Comment by woutervdub** [comment]
Comments fixed, and added dimensions.

File: designs/gozer/src/ghost.mjs (hunk 2 above)
Comment (joostdecock) [reply designs/gozer/src/ghost.mjs:15]: Why not:
```js
if (sa) points.eyeLeft.addCircle(eyeSize + sa, 'fabric sa')
```

designs/gozer/src/ghost.mjs
```diff
@@ -0,0 +1,75 @@
...
+    points.leftCp2 = points.left.shift(90, size * cbqc)
+    points.bottomCp1 = points.bottom.shift(0, size * cbqc)
+    points.bottomCp2 = points.bottom.shift(180, size * cbqc)
+    points.rightCp1 = points.right.shift(90, size * cbqc)
+    points.rightCp2 = points.right.shift(270, size * cbqc)
+
+    points.eyeLeft.addCircle(eyeSize)
+    points.eyeRight.addCircle(eyeSize)
+
+    paths.seam = new Path()
```
joostdecock, line 31: You are adding this twice, once as a circle, once as a path.
All you need is:
```js
points.middle = new Point(0, 0).addCircle(size, 'fabric')
```

designs/gozer/src/ghost.mjs
woutervdub: Well, I tried `if (sa) points.middle.addCircle(eyeSize + sa).attr('class','fabric sa')` and that didn't work. So I just went back to the old way.

designs/gozer/src/ghost.mjs
woutervdub: Figured that out now. RTFM

joostdecock (approved):

PR #5201, opened 2023-10-18
new design throws error after being created

anna-puk:
after running 'yarn new design', the files/folder are created and yarn lab runs, but the /designs/test3 page gives a server error


joostdecock:
See #5207 instead

PR #5292, opened 2023-10-26
updated path to use {logoPath} instead of string

### LINKED ISSUE #3965: [feature]: Keep FreeSewing's logo DRY

The logo plugin defines the pathstring of the FreeSewing logo.
This is lengthy string, and we're keeping it in more than one place.

So it should be adapted to:

```js
import { logoString } from '../../sites/shared/components/logos/freesewing.mjs'
```

So that we can keep the code DRY.

Note that we might need to apply a transform as this path string is probably built for a different viewBox.

### Additional context

_No response_

sohamtembhurne:
Fixes #3965 
Updated path to use {logoPath} from `../../../sites/shared/components/logos/freesewing.mjs` instead of lengthy string. But there are other instances of the lengthy strings as well. Let me know if I should update them as well.


plugins/plugin-annotations/src/logo.mjs
```diff
@@ -1,8 +1,10 @@
...
+      `<g id="logo" transform="scale(${scale}) translate(-23 -36)"><path class="logo" stroke="none" fill="currentColor" d=${logoPath} /></g>`,
```
joostdecock, line 8: Seems like you forgot to quote the path string

joostdecock (changes requested):
If there's other locations, then sure.
We need to check the viewBox though. Because the string you replaced is not the same as this one. So we may need to apply a transform.

sohamtembhurne:
> If there's other locations, then sure. We need to check the viewBox though. Because the string you replaced is not the same as this one. So we may need to apply a transform.

I analyzed these svgs. There are two different logo strings being used. First the one that I replaced in `logo.mjs` and another the one present in `../../../sites/shared/components/logos/freesewing.mjs`. There are two ways we can encounter this - 
1. To transform the path `logoPath` that I imported from `freesewing.mjs`. This can be done by scaling it 2 times and translating -2px left. 
2. We can create another string in `freesewing.mjs`, consisting of the path that I replaced. Since this path is used in around 8 files. 

In my opinion, it will be better to create a new path (new path refers to the one I replaced) variable.

plugins/plugin-annotations/src/logo.mjs
sohamtembhurne: If you look closely, this is using template literals - `<g> ...        ${logoPath}         ...</g>`, so it doesn't need quotes as logoPath is itself a string in `../../../sites/shared/components/logos/freesewing.mjs`

joostdecock: Yes, but this will inject the string without quotes. So your code will generate this:

```svg
<path class="logo" stroke="none" fill="currentColor" d=M 35.222,0 C 34.233,0... />
```

But what we need is valid SVG output, like this:
```
<path class="logo" stroke="none" fill="currentColor" d="M 35.222,0 C 34.233,0..." />
```

So, while it is true that from a pure Javascript point of view, it is not required to put quotes here, it would result in invalid XML/SVG syntax. That's why that quotes are required.

joostdecock:
> > If there's other locations, then sure. We need to check the viewBox though. Because the string you replaced is not the same as this one. So we may need to apply a transform.
> 
> I analyzed these svgs. There are two different logo strings being used. First the one that I replaced in `logo.mjs` and another the one present in `../../../sites/shared/components/logos/freesewing.mjs`. There are two ways we can encounter this -
> 
>     1. To transform the path `logoPath` that I imported from `freesewing.mjs`. This can be done by scaling it 2 times and translating -2px left.
> 
>     2. We can create another string in `freesewing.mjs`, consisting of the path that I replaced. Since this path is used in around 8 files.
> 
> 
> In my opinion, it will be better to create a new path (new path refers to the one I replaced) variable.

The goal is to have 1 single path string for the logo. And then apply transforms based on the use case.
If you'd like to tackle that, then go ahead.

plugins/plugin-annotations/src/logo.mjs
sohamtembhurne: I have made the required changes.

sohamtembhurne:
> > > If there's other locations, then sure. We need to check the viewBox though. Because the string you replaced is not the same as this one. So we may need to apply a transform.
> > 
> > 
> > I analyzed these svgs. There are two different logo strings being used. First the one that I replaced in `logo.mjs` and another the one present in `../../../sites/shared/components/logos/freesewing.mjs`. There are two ways we can encounter this -
> > ```
> > 1. To transform the path `logoPath` that I imported from `freesewing.mjs`. This can be done by scaling it 2 times and translating -2px left.
> > 
> > 2. We can create another string in `freesewing.mjs`, consisting of the path that I replaced. Since this path is used in around 8 files.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In my opinion, it will be better to create a new path (new path refers to the one I replaced) variable.
> 
> The goal is to have 1 single path string for the logo. And then apply transforms based on the use case. If you'd like to tackle that, then go ahead.

I have transformed the logo based on the use case.

joostdecock (approved):

joostdecock:
@all-contributors please add @sohamtembhurne for code

PR #5312, opened 2023-10-28
Onyx - new armpitEase option

Thrunic:
Testing has shown that placing the sleeves perfectly centered around the shoulder and armpit causes pinching on tight garments when the arms are held downwards.

As such, I'm adding an armpitEase option to move the sleeve lower by an adjustable amount.

PS: I had wrongly measured my waistToArmpit measurement (too small), which caused me to not run into this issue until now since that measurement error led to ease being introduced there.


markdown/org/docs/designs/onyx/options/armpitEase/en.md
```diff
@@ -0,0 +1,11 @@
+---
+title: "Armpit ease"
+---
+
+![The armpit ease option on Onyx](armpitease.svg)
```
joostdecock, line 5: This seems to refer to a non-existing svg file?

markdown/org/docs/designs/onyx/options/armpitEase/en.md
joostdecock, line 1: You need to lowercase all option names. URLs should always be lowercase.

joostdecock (changes requested):

markdown/org/docs/designs/onyx/options/armpitEase/en.md
Thrunic: Fixed. Sorry about that.

markdown/org/docs/designs/onyx/options/armpitEase/en.md
Thrunic: That's probably caused by me erroneously using a camelCase directory name. Hopefully fixing that has fixed that problem.

Thrunic:
The PR should be good now.

markdown/org/docs/designs/onyx/options/armpitEase/en.md
joostdecock: What I mean is: This SVG file is not part of this PR. And since it's a new option, it's not in the repo either.
So if you have it locally, you should add it to the PR.

joostdecock: This problem is actually a lot more widespread.
44 options of Onyx reference an SVG file:
```
joost@machine:~/git/freesewing/markdown/org/docs/designs/onyx/options$ grep -R svg | grep en.md | wc -l
44
```

Yet none of them actually provide said file:
```
joost@machine:~/git/freesewing/markdown/org/docs/designs/onyx/options$ ls * | grep svg | wc -l
0
```

So, what is your intent here?
- If you intent to provide SVG files to illustrate the option, please add them
- If you did not intent to do that, then don't include non-existing images in the documentation

Thrunic: I thought that the file was autogenerated? I'll see if I can remember how they're generated and run the build script that does that, or just remove the .svg links if that's not the case.

Thrunic: The script is 'node ./scripts/optiondocsimages.mjs', but it tries to looks in the old patterns directory, which fails since it's the designs directory now.

If this functionality is old or deprecated, I'll go ahead and remove the .svg links. In my opinion, the functionality in the v3 lab covers what these .svg files did, and better.

Thrunic:
The .svg links have been removed, and a few other minor issues fixed. I think we've good to go now.

markdown/org/docs/designs/onyx/options/armpitEase/en.md
joostdecock: > the functionality in the v3 lab covers what these .svg files did, and better

Yes, that's why these images have been removed in all docs.
The script is still there, but it's no longer in use.

joostdecock (approved):

PR #5423, opened 2023-11-05
Shelly - Support for 'expand' setting.

Thrunic:
The PR adds support to the 'expand' setting for the Shelly design, causing the front, back, and neckband pieces to generate as full pieces instead of on-the-fold when the expand setting is set to true.

### DIFF
--- a/designs/shelly/src/back.mjs
+++ b/designs/shelly/src/back.mjs
@@ -13,6 +13,7 @@ function draftBack({
   macro,
   snippets,
   Snippet,
+  expand,
 }) {
   const raglanAngle = store.get('raglanAngle')
   const neckRadius = store.get('neckRadius')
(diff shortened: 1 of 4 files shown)


Thrunic:
If there's a simpler way to do this expand a pattern piece, please let me know.

Thrunic:
I'm looking at the existing designs, and they don't use expand to mirror the on-the-fold parts, and expand is enabled by default. I'm turning the PR into a draft, because I'm not sure it's a good idea if expand is the default, and I don't know how to change the default to 'no'.

Thrunic:
I'm closing this PR because it's been explained to me that 'expand' isn't intended for this purpose, and that the functionality is in a yet-to-be-implemented 'cut'.

Issue #5435, opened 2023-11-07
The `draftForHighBust` option in Brian does not (seem to) apply


joostdecock:
We need to investigate whether it's not injecting the (optional) `highBust` measurement correctly (which would make it a frontend bug), or that it's somehow a bug in the design itself.

PR #5441, opened 2023-11-08
fix(core): Corectly load conditional plugins

joostdecock:
The parameters passed to the conditional load method were incorrect causing a conditional plugin to never get loaded.


joostdecock:
This fixes #5435

PR #5443, opened 2023-11-08
Box - Update to v3 standards

Thrunic:
I've updated the basic box added to new designs to meet v3 standards. In particular:

- Complete check is removed (done implicitly now for the title and logo.
- Paperless check is removed (done implicitly now for dimensions).
- .attr('class', 'fabric sa') is now .addClass('fabric sa') (it's easier to read).

### DIFF
--- a/config/templates/design/src/box.mjs
+++ b/config/templates/design/src/box.mjs
@@ -1,17 +1,4 @@
-function draftBox({
-  options,
-  Point,
-  Path,
-  points,
-  paths,
-  Snippet,
-  snippets,
-  complete,
-  sa,
-  paperless,
-  macro,
-  part,
-}) {
+function draftBox({ options, Point, Path, points, paths, Snippet, snippets, sa, macro, part }) {
   const w = 500 * options.size
   points.topLeft = new Point(0, 0)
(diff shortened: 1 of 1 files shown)


config/templates/design/src/box.mjs
```diff
@@ -27,33 +14,29 @@ function draftBox({
...
-    snippets.logo = new Snippet('logo', points.logo)
-    points.text = points.logo
-      .shift(-90, w / 8)
-      .attr('data-text', 'hello')
-      .attr('data-text-class', 'center')
+  points.logo = points.topLeft.shiftFractionTowards(points.bottomRight, 0.5)
+  snippets.logo = new Snippet('logo', points.logo)
+  points.text = points.logo
+    .shift(-90, w / 8)
+    .attr('data-text', 'hello')
```
joostdecock, line 21: Since we're improving things, could you change this into:

```js
.addText('hello', 'center')
```

joostdecock (changes requested):

config/templates/design/src/box.mjs
Thrunic: Done.

joostdecock (approved):

Issue #5448, opened 2023-11-10
[bug]: `rmcutonfold` macro does not remove the `store.cutlist.setCutOnFold`


woutervdub:
### What seems to be the problem? 🤔

Removing the cutonfold macro does not remove the store.cutlist entries that correspond to it

### Design / Plugin / Package 🧐

cutonfold.mjs

### Are you a FreeSewing patron? 😃

Yes, I am a tier-2 patron ❤️

joostdecock:
Resolved in #5449

PR #5449, opened 2023-11-10
Fix bug #5448

woutervdub:


joostdecock:
Thanks for this @woutervdub I've added some more edge-cases that would fall victim to similar problems.
Could you check from your side that this (still) resolves your issue? :pray:

woutervdub:
@joostdecock I'm getting errors that `partName` is undefined. But the `store` at the point of when the error is thrown, does contain this: `activePart: "noble.frontPoints"`, so I'm not sure what is going on.

joostdecock:
@woutervdub Sorry for the delay. Can you check again now?

woutervdub:
> @woutervdub Sorry for the delay. Can you check again now?

@joostdecock, you're using `removeGrain()` and `removeGrainline()`, while they should be the same thing. I don't know which term you prefer, so I didn't make any changes. My vote would be for `removeGrainline()`.

curonfold.mjs (38) `store.cutlist.removeGrainline()`
grainline.mjs (36) `store.cutlist.removeGrainline()`
cutlist.mjs (103) `function removeGrain(store)`

joostdecock:
> @joostdecock, you're using `removeGrain()` and `removeGrainline()`, while they should be the same thing. I don't know which term you prefer, so I didn't make any changes. My vote would be for `removeGrainline()`.
> 
> curonfold.mjs (38) `store.cutlist.removeGrainline()` grainline.mjs (36) `store.cutlist.removeGrainline()` cutlist.mjs (103) `function removeGrain(store)`
> 
> Once this is resolved, all works fine for `tristan`.

Good catch. I think `removeGrainline` is a bit misleading as it seems to imply it would remove the grain line.
It does not. All it does is remove the data about the fabric gain in the store, which is a value in degrees. There is no line involved, or being removed.

So I've changed it to `removeGrain`, since it's also `setGrain` (and not `setGrainLine`).

PR #5557, opened 2023-11-27
Fix links on the documentation front page. Closes #5543

### LINKED ISSUE #5543: [bug]: Broken link
The link is broken to https://freesewing.org/docs/guide . I think I was trying to get to it through the Documentation page.

[User 54234](https://next.freesewing.org/users/54234)

ArcticCoder:
Fix link to "Getting started with FreeSewing.org" from the front page of the documentation. Closes #5543

I also noticed that all of the other links on the non-English versions of the page were directing to English-language documentation, which I don't believe was intended, so I fixed it as part of fixing the links on that page.


joostdecock (approved):

joostdecock:
@ArcticCoder Thanks for this. There's no need (not even any point) in updating the non-English files. The translation software will handle them.

Client-side routes should handle making sure that `/docs` goes to `/de/docs` (for example) if the user has selected German.

PR #5573, opened 2023-11-30
Fix links to HPS after HPS doc was moved

### LINKED ISSUE #5568: [bug]: Localize HPS URL not found
When filling my mesurments, I found that the URL linking to "localize your HPS point" redirects you to a 404 error:
Example:
In https://freesewing.org/docs/measurements/hpstowaistback, the link localize your HPS point goes to https://freesewing.org/docs/measurements/hps

[User 54703](https://next.freesewing.org/users/54703)

ArcticCoder:
This was missed when HPS doc was moved. Closes #5568


joostdecock:
Thanks a lot @ArcticCoder 🙏

PR #5590, opened 2023-12-04
Migrate use-persisted-state to use-local-storage-state

### LINKED ISSUE #5125: [feature]: Migrated `use-persisted-state` to `use-local-storage-state`

We currently use this, but it's unmaintained: https://github.com/donavon/use-persisted-state
It's also spawning annoying errors because it does not support React 18 as a peer dependency.

We should migrate to this: https://github.com/astoilkov/use-local-storage-state

It's typically used in the hooks in `sites/shared/hooks`

### Additional context

This is a good candidate for people looking to contribute :)

ghost:
Replace `use-persisted-state` with `use-local-storage-state`. Removed `sites/shared/hooks/use-storage.mjs`, as it seems to be entirely unused. If desired I will ofc restore it and migrate it as well. Closes #5125 .


joostdecock:
Thank you @ArcticCoder :pray:

PR #5602, opened 2023-12-06
Skully documentation update

woutervdub:


designs/skully/src/uppermouth.mjs
```diff
@@ -125,13 +125,13 @@ export const uppermouth = {
     paths.front1 = new Path()
       .move(points.pointM1)
       .curve_(points.pointM1Cp1, points.point1)
-      .setText(complete ? '(16)' : '', textAttribute)
+      .setText(complete ? 'mouthTop' + '(16)' : '', textAttribute)
```
joostdecock, line 128: This will break translation, since the text will be added as a single `mouthTop(16)` string.

Instead we should add the text in 2 steps:

```js
.addText(complete ? 'skully:mouthTop' : '')
.addText(complete ? '(16)' : '')
```

And then add `mouthTop` to `i18n/en.json` in the `s` section.

That should ensure it gets translated

designs/skully/src/uppermouth.mjs
```diff
@@ -125,13 +125,13 @@ export const uppermouth = {
...
       .addClass('hidden')
 
     paths.front2 = new Path()
       .move(points.mPoint1)
       .curve_(points.mPointM1Cp1, points.pointM1)
-      .setText(complete ? '(16)' : '', textAttribute)
+      .setText(complete ? 'mouthTop' + '(16)' : '', textAttribute)
```
joostdecock, line 134: Same translation remark here

joostdecock (changes requested):
This is on my phone, so I didn't actually run it, but I left some remarks because I suspect this will break translation

designs/skully/src/uppermouth.mjs
woutervdub: Thanks! I'll make adjustments accordingly. There are more places in the code where this should be changed.

joostdecock (approved):
Awesome 🎉

PR #5652, opened 2023-12-15
Fixes #5569 (`path.split()`)

### LINKED ISSUE #5569: [bug]: `path.split()` missing parts in one of the resulting segments

When a part is `split()` using a `point` that was obtained with a `intersects()` method, sometimes the `path` is split into two parts where one part is missing most parts in the `ops` array. This only occurs in very limited cases.

Attached are two JSON files with objects containing the path (path), the intersecting path (intersectingPath), the intersecting point (intersectPoint), and the result (results). One (`correct.json`) is where the `split()` results in a correct array of segments, and the other (`wrong.json`) where the second segment is missing most of the parts at the end. Included are also two screen captures of the problem, which correspond with the JSON files.

May of may not be useful: The path used to be split was generated using the `offset()` method. The path was then extended on both ends with a `line()` to make sure there is always a intersecting point.

Code segment:
```javascript
      paths.front = ExtendPath(paths.front.offset(gussetWidth)).hide()
      paths.waistTemp = new Path().move(points.frontWaistband).line(points.centerWaistband)
      points.frontWaistband = paths.front.intersects(
        new Path().move(points.frontWaistband).line(points.centerWaistband)
      )[0]
      console.log({
        path: paths.front,
        intersectingPath: paths.waistTemp,
        intersectPoint: points.frontWaistband,
        results: paths.front.split(points.frontWaistband),
      })
```

[correct.json](https://github.com/freesewing/freesewing/files/13502921/correct.json)
[wrong.json](https://github.com/freesewing/freesewing/files/13502922/wrong.json)

wrong
correct

packages/core

Yes, I am a tier-2 patron ❤️

### Additional context

_No response_

woutervdub:


packages/core/src/path.mjs
```diff
@@ -802,76 +802,65 @@ Path.prototype.split = function (point) {
...
     if (path.ops[1].type === 'line') {
-      if (path.ops[0].to.sitsRoughlyOn(point)) {
-        secondHalf.push(new Path().__withLog(this.log).move(path.ops[0].to).line(path.ops[1].to))
-      } else if (path.ops[1].to.sitsRoughlyOn(point)) {
-        firstHalf.push(new Path().__withLog(this.log).move(path.ops[0].to).line(path.ops[1].to))
-      } else if (pointOnLine(path.ops[0].to, path.ops[1].to, point)) {
+      if (pointOnLine(path.ops[0].to, path.ops[1].to, point)) {
         firstHalf = divided.slice(0, pi)
-        firstHalf.push(new Path().__withLog(this.log).move(path.ops[0].to).line(point))
+        firstHalf.push(new Path().__withLog(this.log).move(path.ops[0].to).line(point.copy()))
```
joostdecock, line 817: This `.copy()` at the end seems unneeded

packages/core/src/path.mjs
```diff
@@ -802,76 +802,65 @@ Path.prototype.split = function (point) {
...
         pi++
         secondHalf = divided.slice(pi)
-        secondHalf.unshift(new Path().__withLog(this.log).move(point).line(path.ops[1].to))
+        secondHalf.unshift(new Path().__withLog(this.log).move(point.copy()).line(path.ops[1].to))
```
joostdecock, line 820: Here too, this `copy()` does not seem to serve any purpose

packages/core/src/path.mjs
```diff
@@ -802,76 +802,65 @@ Path.prototype.split = function (point) {
...
+        break
       }
     } else if (path.ops[1].type === 'curve') {
-      if (path.ops[0].to.sitsRoughlyOn(point)) {
-        secondHalf.push(
-          new Path()
-            .__withLog(this.log)
-            .move(path.ops[0].to)
-            .curve(path.ops[1].cp1, path.ops[1].cp2, path.ops[1].to)
+      let t = pointOnCurve(path.ops[0].to, path.ops[1].cp1, path.ops[1].cp2, path.ops[1].to, point)
```
joostdecock, line 824: We now no longer handle the edge-case where we are splitting on the end of a curve. However, there's a unit test for that and it passes. Plus I don't remember why this was added, so let's just go for it :)

joostdecock (changes requested):
If you can remove those unneeded `copy()` calls I think this can go in

joostdecock (approved):

PR #5731, opened 2023-12-30
Add disabled pattern delete button when no patterns selected

ArcticCoder:
Adds a grey/disabled delete button to `account/patterns` when no patterns are selected. Motivated by multiple users (me included) at least initially glossing over the fact that this is how pattern deletion is done.

### DIFF
--- a/sites/shared/components/account/patterns.mjs
+++ b/sites/shared/components/account/patterns.mjs
@@ -585,7 +585,11 @@ export const Patterns = () => {
         <button className="btn btn-error" onClick={removeSelectedPatterns}>
           <TrashIcon /> {selCount} {t('patterns')}
         </button>
-      ) : null}
+      ) : (
+        <button className="btn" onClick={null}>
+          <TrashIcon /> {selCount} {t('patterns')}
+        </button>
+      )}
       <TableWrapper>
         <table className="table table-auto">
           <thead className="border border-base-300 border-b-2 border-t-0 border-x-0">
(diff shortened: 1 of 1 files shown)


joostdecock (commented):
No need to duplicate the button markup here. You can use the `disabled` attribute. Something like:

```js
<button
  onClick=...
  className=...
  disabled={selCoumt < 1}
>
...
</bultton>
```

joostdecock (changes requested):
Keep in mind that disabled should evaluate to `true` to disable the button

ArcticCoder:
Doih :facepalm: :sweat_smile:  Thanks :D

joostdecock (approved):

joostdecock:
Great work again @ArcticCoder thanks a bunch, and happy new year 🎉

joostdecock:
Just realized that this makes the UI somewhat inconsistent as there's other pages that use the same list of items with the *select one to make the delete button appear* approach:

- https://freesewing.org/account/bookmarks
- https://freesewing.org/account/sets
- https://freesewing.org/account/apikeys

Any chance you'd like to update those too @ArcticCoder ?

ArcticCoder:
> Just realized that this makes the UI somewhat inconsistent as there's other pages that use the same list of items with the _select one to make the delete button appear_ approach:
> 
>     * https://freesewing.org/account/bookmarks
> 
>     * https://freesewing.org/account/sets
> 
>     * https://freesewing.org/account/apikeys
> 
> 
> Any chance you'd like to update those too @ArcticCoder ?

Sure thing! ☺️

PR #5741, opened 2024-01-02
Unify user facing delete buttons to be disabled when nothing selected

VSinerva:
Fixes inconsistent behavior of delete buttons on `/account/*` pages after `account/patterns` was changed. Now all pages have a visible but disabled delete button when 0 elements are selected. As per [Joost.](https://github.com/freesewing/freesewing/pull/5731#issuecomment-1873926677)

The button for bookmarks has a special condition to keep the behavior consistent with the other pages (i.e. on the pages where the table is hidden when 0 items exist, the remove button is also hidden). This also solves a UI spacing issue, though that could easily be solved another way if preferred :D

The following files also contain the same pattern, but they seem to be curator/admin controls (if I'm not mistaken) so I'm leaving it up to you if you want them changed or not:
`sites/shared/components/curated-sets.mjs`
`sites/shared/components/submissions/index.mjs`

I also lack permissions on the production site (and had trouble spinning up my own backend) so wouldn't have been able to test these changes. The rest I tested by hooking up to the production backend.


joostdecock (approved):

joostdecock:
@all-contributors please add @VSinerva for code

Issue #5753, opened 2024-01-03
[bug]: Cutlist info does not get overwritten in the title macro


joostdecock:
### What seems to be the problem? 🤔

When you create a part `from` another, and give it a title, that new `title` macro will overwrite the title from the inherited part (which is what we want).
However, it will not overwrite the cutlist info, resulting in these floating labels as seen in this screenshot shared by Discord user `munic9459`:

![](https://cdn.discordapp.com/attachments/1191972523152244806/1191972523567489124/Screenshot_2024-01-02_at_7.40.28_PM.png?ex=65a761eb&is=6594eceb&hm=6c21fc5c08bd466b0267c8094a750e52668244c8238ed4528fdf310d082cd47f&)

This is a bug in the annotations plugin.

### Are you a FreeSewing patron? 😃

Yes, I am a tier-8 patron ❤️  💙 💜

joostdecock:
Fixed in #5773

PR #5760, opened 2024-01-04
Fix for #5753

woutervdub:
Add code to remove points previously defined for the `cutlist`


plugins/plugin-annotations/src/title.mjs
```diff
@@ -112,6 +112,13 @@ const title = function (config, { Point, points, scale, locale, store, part, log
    * Title: cutlist
    */
   if (mc.cutlist) {
+    /*
+     * Remove any cut instructions already defined
```
joostdecock, line 116: This feels a bit hackish 🤔 

It seems to me that the problem is in line 135:

```mjs
const id = `${ids.cutlist}_${material}_${c}`
```

It's creating an arbitrary ID there, rather than using `store.generateMacroIds()`.

A better fix would be to add all of these to the `store.generateMacroIds()` call, so we can keep track of these point IDs.

joostdecock: If I'm not being very clear, I can take a stab at this in the weekend.

joostdecock (changes requested):

plugins/plugin-annotations/src/title.mjs
woutervdub: > This feels a bit hackish 🤔
That's what I told you ;-)

woutervdub: @joostdecock The `title` macro already stores those `id`s appropriately. You can retrieve them using `store.getMacroIds(mc.id)`. But in an inherited part, you don't know what material was used, and as such, you cannot know if there are any to be removed, how many there are to be removed, and what their name is. 
I can turn the whole shebang into an array in the store, and use that to remove them in any subsequent (higher level) part when called. But I'm not sure that's a more elegant solution. (See in this latest commit)

joostdecock:
@woutervdub Since this is in your fork, I didn't wade into this, but setup a feature branch for my alternative approach in #5773 

I'd appreciate your input (I'll ask for a review)

woutervdub:
Cancelled

PR #5773, opened 2024-01-06
feat: Support notes in title macro

joostdecock:
This started out as a fix for #5753 after I felt the approach in #5760 was not the best way to handle this.

The problem lies in the way the cutlist data is added to the pattern by the title plugin. The whole cutlist thing was added by a less experienced contributor, and they were a little too focussed on what they were trying to accomplish, rather than keeping an eye on the big picture.

In this case, a bunch of points get added to the part to anchor all of the text for the cutlist. This causes problems when inheriting parts because points are added per material, and now we don't know what materials were used and how to remove these points.

I appreciate @woutervdub effort to resolve this in #5760 but as I mentioned above, I don't love the approach.
Rather than finding a clever way to remove these points, I am of the opinion that they should never have been added in the first place.

Specifically, all we are doing is adding one or more lines of text. The idea that you need multiple points for this is an incorrect assumption, and therein lies the root of this problem.

So, this refactor adds support to the title macro for 'notes'. These notes will be added below the title. The cutlist info is now just prepended to the notes. So it stops being something 'special' (which it never should have been) and instead just behaves like other text that is placed on the part by the title macro.

The only flanking measure required to make this happen was to add support for nested arrays in handling text.
This way, you can pass a nested array to concatenate strings in such a way that each part will still be individually translated.

And as all text in FreeSewing, to force a linebreak, you just use "\n" as The Dark Lord intended.

### DIFF
--- a/config/changelog.yaml
+++ b/config/changelog.yaml
@@ -1,6 +1,18 @@
 Unreleased:
   Added:
 
+  Added:
+    plugin-annotations:
+      - The `title` macro now takes a `notes` and `classes.notes` as its config, allowing you to add notes
+      - The `classes.cutlist` config is removed from the title plugin, cutlist info is now included as notes
+    plugin-i18n:
+      - This plugin now supports translation of nested arrays of strings, giving you more flexibility to concatenate translated parts of strings
+    react-components:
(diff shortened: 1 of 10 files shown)


woutervdub (approved):
Looks good to me. I tested it in one of my branches, and it works fine.

PR #5825, opened 2024-01-14
Various fixes

joostdecock:


markdown/dev/tutorials/pattern-design/part3/flag/en.md
```diff
@@ -337,4 +342,3 @@ function draftBib({
 ```
 </Example>
 
-Now the first thing you should know is: **core does not care**. We are
```
BenJamesBen, line 340: FYI, this last line has since been deleted from the file, via e9c75f1 in PR #5836.

PR #5862, opened 2024-01-21
chore(core): Remove options distance debug logging

### LINKED ISSUE #3062: [chore]: (core) Remove DISTANCE_DEBUG messages in pattern.mjs

The `DISTANCE_DEBUG` debug messaging needs to be removed from `package/core/src/pattern.mjs` sometime in the future, either approximately May 2023 or shortly after v3 is released or some other arbitrary time.

Extensive debug messaging controlled by a `DISTANCE_DEBUG` flag was added to `packages/core/src/pattern.mjs` via #3058. This messaging was added to help debug the new part distances functionality introduced in v3. However, the debug messages make the code difficult to read. In the 2022-11-12 Contributor Call, we agreed to remove the debug messages after a period of time when they are no longer needed, after we are sure that the distances functionality is correct and won't need changing.

packages/core

No, I am not 😞

### Additional context

_No response_

BenJamesBen:
In the 2022-11-12 Contributor Call, we agreed to remove the options distance debug logging code once v3 was released. This PR performs that chore and closes #3062.

Options distance information is still logged, but instead of individual messages for each part, things are now logged as a concise summary.

After:
Screenshot 2024-01-20 at 7 45 42 PM

Before:
Screenshot 2024-01-20 at 7 45 29 PM

### DIFF
--- a/packages/core/src/pattern/pattern-config.mjs
+++ b/packages/core/src/pattern/pattern-config.mjs
@@ -64,9 +64,6 @@ export function PatternConfig(pattern) {
   })
 }
 
-/** @type {Boolean} change me to true to get full debugging of the resolution process */
-const DISTANCE_DEBUG = false
-
 ////////////////////
 // PUBLIC METHODS //
 ////////////////////
@@ -101,12 +98,14 @@ PatternConfig.prototype.addPart = function (part) {
 
 /** Log the final report on part inheritance order */
 PatternConfig.prototype.logPartDistances = function () {
+  const priorities = []
   for (const partName in this.parts) {
-    let qualifier = DISTANCE_DEBUG ? 'final' : ''
-    this.store.log.debug(
-      `⚪️  \`${partName}\` ${qualifier} options priority is __${this.__mutated.partDistance[partName]}__`
-    )
+    const p = this.__mutated.partDistance[partName]
+    if (p in priorities) priorities[p] += ', ' + partName
+    else priorities[p] = partName
   }
+  for (let p = 1; p < priorities.length; p++)
+    this.store.log.debug(`⚪️  Options priority __${p}__ :\`${priorities[p]}\``)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/core/src/pattern/pattern-config.mjs
```diff
@@ -101,12 +98,14 @@ PatternConfig.prototype.addPart = function (part) {
...
-    let qualifier = DISTANCE_DEBUG ? 'final' : ''
-    this.store.log.debug(
-      `⚪️  \`${partName}\` ${qualifier} options priority is __${this.__mutated.partDistance[partName]}__`
-    )
+    const p = this.__mutated.partDistance[partName]
+    if (p in priorities) priorities[p] += ', ' + partName
+    else priorities[p] = partName
   }
+  for (let p = 1; p < priorities.length; p++)
+    this.store.log.debug(`⚪️  Options priority __${p}__ :\`${priorities[p]}\``)
```
joostdecock, line 108: Since this is a core change, I'll be a lot more strict 😉 

There's a bunch of bugs that this code invites that are mainly due to the use of an array on line 101.
If the array is sparse, it won't work as expected. Also using `in` on an array seems like a bad idea.

Allow me to propose an alternative:

```mjs
// Use an object here
const priorities = {}
// Using for ... in on an object is frowned upon by purists ;)
for (const partName of (Object.keys(this.parts)) {
  const p = this.__mutated.partDistance[partName]
  // Make sure the property exists in the object, and use a Set to avoid duplicates
  if (typeof priorities[p] === 'undefined') priorities[p] = new Set()
  priorities[p].add(partName)
}

// now iterate over the priorities
for (const p of Object.keys(priorities)) {
  /* 
    * Log the list, we'll turn the set into an array
    * then map() it to add the backticks around the part name 
    * (giving it that code markup) 
    * and finally join() them with a comma and space
    */
  this.store.log.debug(`⚪️ Options priority __${p}__ : ${[...priorities[p]].map(p => '`'+p+'`').join(', ')}`)
}
```

That seems a lot more robust to me.
Seems, because I did not test it.

joostdecock (changes requested):

joostdecock (approved):

Issue #5863, opened 2024-01-21
[bug]: Jargon plugin documentation is misleading


LunaticMuch:
### What seems to be the problem? 🤔

The documentation of the plugin is confusing and wrong:
- The plugin is a rehype plugin bug the example shows remark
- The example repo is empty
- The link to Gatsy example is broken (it does not exist anymore apparently)

### Design / Plugin / Package 🧐

packages/rehype-jargon

### Are you a FreeSewing patron? 😃

No, I am not 😞

BenJamesBen:
This bug was fixed by PR #5864. Closing.

PR #5864, opened 2024-01-21
FIX: jargon manual is not quite correct

LunaticMuch:
Address #5863


joostdecock (approved):

joostdecock:
@LunaticMuch thank you for this. The README was not update after out v3 release. Nice catch.

joostdecock:
@all-contributors please add @LunaticMuch for docs

PR #5899, opened 2024-01-26
chore(markdown): Add Svg.render() documentation

### LINKED ISSUE #3193: [docs]: Svg docs are missing render() and possibly other properties

https://freesewing.dev/reference/api/svg

No, I am not 😞

### Suggested content

The public `Svg.render()` method and the `Svg.body` property are missing from the API documentation. It is unclear whether the omissions are intentional. If not, then the documentation should be added.

### Additional context

_No response_

BenJamesBen:
Also a small grammar fix in `Svg.asRenderProps()`.

(This fixes half of Issue #3193. The other half will be addressed in a separate PR.)

Edit: Added:

The other half of #3193 was subsequently merged via #5900. As a result, this PR now closes #3193.

### DIFF
--- a/markdown/dev/reference/api/svg/render/en.md
+++ b/markdown/dev/reference/api/svg/render/en.md
@@ -11,19 +11,11 @@ The `Svg.render()` method will render a drafted
 string svg.render()
 ```
 
-## Svg.render() example
+<Warning>
 
-```
-import { Aaron } from "@freesewing/aaron"
+This method is mostly internal and should not be used directly.
+Instead, it is intended that the
+[Pattern.render()](reference/api/pattern/render)
+method be used to render patterns as SVG.
 
-// Load some public test measurements from the FreeSewing backend
-const measurements = (
-  await (
-    await fetch("https://backend3.freesewing.org/curated-sets/1.json")
-  ).json()
-).measurements
-
-const pattern = new Aaron({ measurements })
-
-const svg = new Svg(pattern).render()
-```
+</Warning>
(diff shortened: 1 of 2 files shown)


markdown/dev/reference/api/svg/render/en.md
```diff
@@ -2,4 +2,28 @@
...
+// Load some public test measurements from the FreeSewing backend
+const measurements = (
+  await (
+    await fetch("https://backend3.freesewing.org/curated-sets/1.json")
+  ).json()
+).measurements
+
+const pattern = new Aaron({ measurements })
+
+const svg = new Svg(pattern).render()
```
joostdecock, line 28: This is mostly an internal method.
We expose the *proper* way to render a pattern via the `pattern.render()` method, like this:

```mjs
const pattern = new Aaron({ measurements })
const svg = pattern.render()
```

My concern is that documenting things this way will lead people who are looking to render a pattern to SVG to this page, and this is not how to do that.

Calling svg.render() will bypass all lifecycle hooks, and not do any layout. I doubt it's what anything would want.

I appreciate the completionist approach, but this might actually cause more confusion.

joostdecock (commented):
If you want to document this, fine. But don't add an example, and add a warning that this is mostly internal and if you're using this you're probably doing it wrong.

markdown/dev/reference/api/svg/render/en.md
BenJamesBen: As suggested, I removed the example and added the warning.

BenJamesBen:
As suggested, I removed the example and added the warning.

joostdecock (approved):

PR #5909, opened 2024-01-27
chore(markdown): Add missing tags to Custom Tags guide

### LINKED ISSUE #3312: [docs]: Undocumented Custom Tags and attributes

https://freesewing.dev/guides/markdown/custom-components

No, I am not 😞

### Suggested content

In the MDX code, there exists some undocumented custom tags:
- `<Link>` which works like the other popout tags like <Related>.
- `<Tab>` and `<Tabs>`
- `<Examples>` (deprecated?)
- `<pre>`, ``, and `<table>`
- `<Method>` and `<StatusCode>` (which seem to be used for REST API docs)

There is also an undocumented `hideable` attribute that works with the popout tags.

It is unclear whether these undocumented items should be documented for use or whether they were intended to remain undocumented.

### Additional context

_No response_

BenJamesBen:
A major edit of the Custom Tags guide.

Fixes #3312.

### DIFF
--- a/markdown/dev/guides/markdown/custom-components/en.md
+++ b/markdown/dev/guides/markdown/custom-components/en.md
@@ -13,40 +13,286 @@ typically require a lot more complexity.
 Below is a list of custom tags that we support in our Markdown-based
 documentation, both for freesewing.dev as freesewing.org.
 
-## Comment
+## Summary and Availability
 
-Use a **Comment** when you want to illustrate something is a personal opinion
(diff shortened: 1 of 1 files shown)


joostdecock (approved):

joostdecock:
This must have taken a good bit of reading the website(s) code @BenJamesBen
I very much appreciate your efforts, but I'm also happy to hear from you when/where you feel things don't make sense.

PR #5913, opened 2024-01-28
DRAFT: fix(jaeger): Correct undercollar to under-collar-and-stand

### LINKED ISSUE #3820: [bug]: Undercollar on Jaeger is incorrect

From discord:

There's only one collar stand. That's correct.
The problem is that the undercollar covers both collar and collar stand. But that's not how it's drawn. Which is a mistake/bug.
There's no collar stand. Just one piece that is collar + collarstand (but this is incorrect on the pattern).
It should also be one piece rather than joined in the middle with a seam.

Sp the undercollar should be the size of collar + collarstand, and should be cut in 1 piece.

designs/jaeger

### Additional context

_No response_

BenJamesBen:
Fixes #3820.

Screenshot 2024-01-28 at 12 24 44 AM

Screenshot 2024-01-28 at 12 25 13 AM

### DIFF
--- a/designs/jaeger/i18n/en.json
+++ b/designs/jaeger/i18n/en.json
@@ -17,7 +17,7 @@
     "pocketLining": "Pocket lining",
     "side": "Side",
     "topSleeve": "Top sleeve",
-    "underCollar": "Under collar",
+    "underCollar": "Under-collar-and-stand",
     "underSleeve": "Under sleeve"
   },
   "s": {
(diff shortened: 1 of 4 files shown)


BenJamesBen:
I'm not sure whether the seam allowances at the bottom and top make sense.

Screenshot 2024-01-28 at 12 28 28 AM

BenJamesBen:
I mis-read the problem and did not make the under-collar-and-stand one piece. Closing this PR. I'll re-file a new one once things are ready.

PR #5916, opened 2024-01-29
fix(jaeger): Fix undercollar to correct shape

### LINKED ISSUE #3820: [bug]: Undercollar on Jaeger is incorrect

BenJamesBen:
Fixes #3820.

- I am questioning whether the seam allowance is correct for the under-collar-and-stand. The previous version of the part had seam allowance only on the 2 short sides. I re-added full seam allowance around all 4 sides. I asked on Discord but didn't get a reply.
- I also rewrote the sewing instructions, but I had to guess at some of the procedures and the order of operation.
- I also renamed the part to "under-collar-and-stand". The name is a bit awkward, but I think it will help customers better understand what the part is and where it goes.
- I did not add any **expand** functionality to this part. I think that could be done later.

(I added the arrows on the below screenshot so I could use it to ask whether the seam allowance is correct for the top and bottom sides.)
Screenshot 2024-01-28 at 1 03 23 AM

Screenshot 2024-01-28 at 4 15 49 PM

### DIFF
--- a/designs/jaeger/i18n/en.json
+++ b/designs/jaeger/i18n/en.json
@@ -17,7 +17,7 @@
     "pocketLining": "Pocket lining",
     "side": "Side",
     "topSleeve": "Top sleeve",
-    "underCollar": "Under collar",
+    "underCollar": "Under-collar-and-stand",
     "underSleeve": "Under sleeve"
   },
   "s": {
--- a/designs/jaeger/i18n/en.json
+++ b/designs/jaeger/i18n/en.json
@@ -17,7 +17,7 @@
     "pocketLining": "Pocket lining",
     "side": "Side",
     "topSleeve": "Top sleeve",
-    "underCollar": "Under-collar-and-stand",
+    "underCollar": "Under collar",
     "underSleeve": "Under sleeve"
   },
   "s": {
--- a/designs/jaeger/src/undercollar.mjs
+++ b/designs/jaeger/src/undercollar.mjs
@@ -1,4 +1,4 @@
-import { collarStand } from './collarstand.mjs'
+import { collar } from './collar.mjs'
 
 /*
  * This collar would benefit from a redesign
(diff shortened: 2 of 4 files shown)


joostdecock:
> *The previous version of the part had seam allowance only on the 2 short sides. I re-added full seam allowance around all 4 sides.*

Issue #3820 is about the undercollar not including the collarstand space. There is no reason to add SA to all seams, as there should not be any.
> *The previous version of the part had seam allowance only on the 2 short sides. I re-added full seam allowance around all 4 sides.*

I am not a fan of this choice. Because *undercollar* is a perfectly fine name. Renaming it to avoid confusion for people who are unfamiliar with collar construction will only confuse those who know what an undercollar is.

In summary, I think it's great to fix the collar so it includes the collarstand height, but there should be no changes to the SA, and I also feel no need to change the name.
What we do need is somebody making a bunch of collars because the last time I did this was years ago, and it's obvious that we should have better documentation/instructions for this.

As such, I would propose to stick to fixing what's broken now, and perhaps try to find someone who would be willing to go through the steps and make sure everything is ok, plus write better docs.

woutervdub (commented):
I have not made `carlton`, nor `carlita`, and as such, I cannot comment on this particular change.

BenJamesBen:
Updated PR, ready for review and merge.
- I corrected the seam allowance back to the original (SA on the sides only).
- I reverted the part's name back to "undercollar".
- I reverted my changes to the Jaeger instructions, except for some minor corrections. Someone more familiar with Jaeger will need to take on the task of improving the instructions, eventually.
- I did add a note specifying that the undercollar is the shape of the combined collar and collarstand, since confusion about this had come up before in Discord.
- Because neither of the Jaeger collar or collarstand parts have `expand` functionality, I didn't add it to the undercollar.

The updated undercollar, compared to the current `develop` version:
Screenshot 2024-02-06 at 9 39 38 AM
Screenshot 2024-02-06 at 9 38 52 AM

joostdecock (approved):

Issue #5873, opened 2024-01-22
[feature]: Write a linter rule to prevent cased directory names


BenJamesBen:
### What is it that you would like to see happen? 🤔

When FreeSewing showcase or documentation directory names are mixed case, duplicate directories and their contents can get created. The original mixed-case directory is created, along with an all-lowercase directory. And, because we translate into other languages, each time this issue occurs up to 6 directories can be affected. See #5715 and #5689.

When the issue occurs it is difficult to fix because you cannot simply delete the extra directories from the git repository. Crowdin will keep re-creating the extra directories and putting them back into the repository. Instead, the extra directories need to be deleted in both Crowdin and the repository simultaneously.

The suggestion is to write a linter rule to prevent this issue.

### Are you a FreeSewing patron? 😃

Yes, I am a tier-2 patron ❤️

### Additional context

For people on macOS systems (that use a case-insensitive filesystem by default), the issue results in filename collisions and warning messages during repository clone operations.

Issue #5715, opened 2023-12-30
[bug]: Case mixup/duplicate directories and files


BenJamesBen:
### What seems to be the problem? 🤔

There are 4 duplicate directories and files with identical names except for uppercase/lowercase letters. The incorrect, duplicate directories should be deleted. 

```sh
% git clone https://github.com/freesewing/freesewing.git
Cloning into 'freesewing'...
remote: Enumerating objects: 1079506, done.
remote: Counting objects: 100% (451/451), done.
remote: Compressing objects: 100% (191/191), done.
remote: Total 1079506 (delta 276), reused 415 (delta 257), pack-reused 1079055
Receiving objects: 100% (1079506/1079506), 536.64 MiB | 15.07 MiB/s, done.
Resolving deltas: 100% (823369/823369), done.
Updating files: 100% (16978/16978), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'markdown/org/docs/designs/onyx/options/zipperPosition/de.md'
  'markdown/org/docs/designs/onyx/options/zipperposition/de.md'
  'markdown/org/docs/designs/onyx/options/zipperPosition/en.md'
  'markdown/org/docs/designs/onyx/options/zipperposition/en.md'
  'markdown/org/docs/designs/onyx/options/zipperPosition/es.md'
  'markdown/org/docs/designs/onyx/options/zipperposition/es.md'
  'markdown/org/docs/designs/onyx/options/zipperPosition/nl.md'
  'markdown/org/docs/designs/onyx/options/zipperposition/nl.md'
  'markdown/org/docs/designs/otis/options/sleeveType/de.md'
  'markdown/org/docs/designs/otis/options/sleevetype/de.md'
  'markdown/org/docs/designs/otis/options/sleeveType/en.md'
  'markdown/org/docs/designs/otis/options/sleevetype/en.md'
  'markdown/org/docs/designs/otis/options/sleeveType/es.md'
  'markdown/org/docs/designs/otis/options/sleevetype/es.md'
  'markdown/org/docs/designs/otis/options/sleeveType/nl.md'
  'markdown/org/docs/designs/otis/options/sleevetype/nl.md'
  'markdown/org/docs/designs/otis/options/snapPlacket/de.md'
  'markdown/org/docs/designs/otis/options/snapplacket/de.md'
  'markdown/org/docs/designs/otis/options/snapPlacket/en.md'
[...]

Issue #5689, opened 2023-12-22
[bug]: Extra brazenedminstrel-double-ws showcase directory


BenJamesBen:
### What seems to be the problem? 🤔

There are two directories in the repository differing only slightly in capitalization:
```
markdown/org/showcase/brazenedMinstrel-double-ws/
markdown/org/showcase/brazenedminstrel-double-ws/
```

It seems as if one of the directories might have been an accidentally created duplicate. If so, it might be that we would want to remove the extra, unneeded directory.

I'm not sure which directory is actually used by our Showcase CMS. There are reasons why each directory seem to be the correct one, adding to my uncertainty:
- The capitalized `brazenedMinstrel-double-ws` directory contains an `en.md` file which the other directory is missing. So, it seems as if it might be the correct one.
- All other directories in `markdown/org/showcase/` are uncapitalized, suggesting an intended naming convention. So, it seems that the uncapitalized `brazenedminstrel-double-ws` directory might be the correct one.

### Are you a FreeSewing patron? 😃

No, I am not 😞

### Additional context

I noticed this issue when a name collision occurred during `git clone` on a macOS system (which has a case-insensitive filesystem):
```sh
Cloning into 'freesewing'...
remote: Enumerating objects: 1074324, done.
remote: Counting objects: 100% (5741/5741), done.
remote: Compressing objects: 100% (2216/2216), done.
remote: Total 1074324 (delta 4006), reused 5017 (delta 3453), pack-reused 1068583
Receiving objects: 100% (1074324/1074324), 536.03 MiB | 15.17 MiB/s, done.
Resolving deltas: 100% (818590/818590), done.
Updating files: 100% (16955/16955), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'markdown/org/showcase/brazenedMinstrel-double-ws/de.md'
  'markdown/org/showcase/brazenedminstrel-double-ws/de.md'
  'markdown/org/showcase/brazenedMinstrel-double-ws/es.md'
  'markdown/org/showcase/brazenedminstrel-double-ws/es.md'
  'markdown/org/showcase/brazenedMinstrel-double-ws/fr.md'
  'markdown/org/showcase/brazenedminstrel-double-ws/fr.md'
  'markdown/org/showcase/brazenedMinstrel-double-ws/nl.md'
  'markdown/org/showcase/brazenedminstrel-double-ws/nl.md'
[...]

PR #5970, opened 2024-02-05
fix(markdown): Remove cased folder names

BenJamesBen:
(The extra files/directories will also need to be deleted from Crowdin or else they will continue to be propagated back to the repository.)

See #5715 and #5689 for previous times the issue arose. See #5873 for the enhancement request to prevent the issue in the future.

I was temporarily wedged because of this issue.
It was because the contents of these two files are different:
`markdown/org/docs/designs/otis/options/snapPlacket/nl.md` `markdown/org/docs/designs/otis/options/snapplacket/nl.md`

In the local repository on my case-insensitive filesystem, git told me that the `snapPlacket` file was modified. Because of this, it wouldn't allow me to switch branches without `--force` and it also caused problems when trying to rebase to `develop`. Now, I could easily fix the `snapPlacket` file. However, doing so would result in the other `snapplacket` version being modified, and I would be stuck with the same git switch and rebase issues.

I was able to work around the issue by mounting a flash drive formatted with a case-sensitive filesystem, cloning another copy of the repository on it, and working from that repository.


joostdecock (approved):

joostdecock:
Oh yes, this is the absolute worst to deal with if you're on a non-case-sensitive filesystem.

PR #5982, opened 2024-02-07
Fix bugs with test view and conditionally hidden options

### LINKED ISSUE #5975: [bug]: Test view broken for designs that conditionally hide options

The test view is blocked by errors for the following designs:

- Bent
- Breanna
- Carlita
- Carlton
- Simon(e)

Some of these have a common causes, but there doesn't seem to be a single cause for all. This happens in my local dev environment as well as the production site.

designs/bent, designs/breanna, designs/carlita, designs/simon, designs/simone

### Additional context

_No response_

VSinerva:
Closes #5975

The following patterns were crashing the test view due to the implementation of conditional options:

- Bent
- Breanna
- Carlita
- Carlton
- Simon(e)

 The test view was also not properly changing to hide/show options as needed, so this was also fixed.

### DIFF
--- a/designs/bent/src/sleeve.mjs
+++ b/designs/bent/src/sleeve.mjs
@@ -192,7 +192,7 @@ export const sleeve = {
     sleeveBend: { deg: 10, min: 0, max: 20, menu: 'fit' },
     draftForHighBust: {
       bool: false,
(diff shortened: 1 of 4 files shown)


joostdecock (approved):

joostdecock:
Nice work @VSinerva thanks a lot 🙏

PR #5989, opened 2024-02-08
fix(carlita, carlton, jaeger): In cutting doc, add comment with non-nett canvas method

### LINKED ISSUE #5457: [docs]: Carlton cutting instructions remarks
Hi!
I'm making the Carlton coat right now and I have some remarks on the instructions on how to custom draft the canvas pieces.

First of all, the suggestion to cut the canvases net.
Whilst I get that the support should not be worked into the seam allowance, I would much rather suggest to cut all canvases with seam allowance, rather than net, to proceed clip the excess off till the seam allowance of the main fabric piece. This way, ease can be worked into the garment, for a much better shape which is supported by the canvas till the seam.
Whilst padstitching the front shoulder to the chest canvas (diagonally) I worked 0,5 cm into the surface, which is not supported as I have it all cut net.I cut the chest canvas and shoulder piece net, then padstitched it, and with one half I worked 0,5 cm

Chest canvas
It is suggested that the lapel part should be discarded. Yet, the lapel ought to be padstitched for maximum shape. Padstitching to what…? Here too I’d like to emphasize it would be better to leave the seam allowance on for the time being. It could be clipped later.

Thank you for your time and effort! To illustrate the pad stitching of the front shoulder, have this image.

[User 28034](https://next.freesewing.org/users/28034)

BenJamesBen:
Closes #5457.

### DIFF
--- a/markdown/org/docs/designs/carlita/cutting/en.md
+++ b/markdown/org/docs/designs/carlita/cutting/en.md
@@ -60,6 +60,21 @@ If your main fabric is quite lightweight and flimsy you may need to interface al
 
 <Note>
 
-The canvas parts need to be cut nett (without seam allowances).
+The canvas parts need to be cut _nett_ (without seam allowances).
 
 </Note>
+
+<Comment by="Benjamin F.">
+
+Instead of cutting the canvas parts nett, there is an alternate
+method where seam allowance is included on the canvas parts.
+The idea is that because the canvas parts have seam allowance,
+when they are attached to the main fabric pieces this can allow
+ease to be worked into the garment for a better shape.
+The seam allowance is then trimmed from the canvas at a later point
+in the garment construction process.
+
+However, because this is a more advanced tailoring technique,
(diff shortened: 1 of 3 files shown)


SeaZeeZee (commented):
Can you explain what you mean by ease being worked into the garment? How does that relate to the canvas pieces specifically? Linings and stiffening fabrics are generally not eased into the main fabric as far as I'm aware.

joostdecock (approved):

BenJamesBen:
> Can you explain what you mean by ease being worked into the garment? How does that relate to the canvas pieces specifically? Linings and stiffening fabrics are generally not eased into the main fabric as far as I'm aware.

I don't know the answer to this. I simply copied the opinion and wording from #5457's author. (If I had to guess, I assume that it means the main fabric pieces being stretched and shaped, along with its backing canvas, before being sewn into place. The pieces being shaped instead of being flat would provide garment ease in certain areas.)

PR #5990, opened 2024-02-08
fix(markdown): Path doc updates for optional parameters

BenJamesBen:


markdown/dev/reference/api/path/addtext/en.md
```diff
@@ -7,7 +7,7 @@ The `Path.addText()` method adds text to the path.
 ## Signature
 
 ```js
-Path path.addText(string text, string className)
+Path path.addText(string text[, string className])
```
joostdecock, line 10: While that `[ brackets ]` notation is rather common in man files and so on, it's not how we tend to document things in the FreeSewing docs.

Parameters that are optional typically have a default value. So I'm in favor of showing that:
```mjs
Path path.addText(string text, string className = '')
```

This makes it obvious the parameter is optional to people who are used to JavaScript, as opposed to people who are used to read man pages.

markdown/dev/reference/api/path/attr/en.md
```diff
@@ -5,16 +5,16 @@ title: Path.attr()
...
 instead, thereby overwriting the value of the attribute.
 
 ## Signature
 
 ```js
 Path path.attr(
   string name,
-  mixed value,
-  bool overwrite = false
+  mixed value[,
```
joostdecock, line 16: Same remark. No need for the square brackets as the optional value has a default assigned.

markdown/dev/reference/api/path/join/en.md
```diff
@@ -8,9 +8,12 @@ The `Path.join()` method joins this path with another path.
 ## Signature
 
 ```js
-Path path.join(path other)
+Path path.join(path other[, bool closed=false])
```
joostdecock, line 11: Same remark. No need for the square brackets as the optional value has a default assigned.

markdown/dev/reference/api/path/settext/en.md
```diff
@@ -7,7 +7,7 @@ The `Path.setText()` method sets text on the path.
 ## Signature
 
 ```js
-Path path.setText(string text, string className)
+Path path.setText(string text[, string className])
```
joostdecock, line 10: Same remark. No need for the square brackets as the optional value has a default assigned.

joostdecock (changes requested):

BenJamesBen:
Changed the optional parameters from man page format to JavaScript format, in all files.

joostdecock (approved):

PR #5991, opened 2024-02-08
fix(core): Add missing anchor property to Stack() constructor

BenJamesBen:
I noticed that the `anchor` property gets added only when/after `Stack.home()` was called.

(No docs updates are needed. The `anchor` property is already listed in the `Stack` Core API doc.)

### DIFF
--- a/packages/core/src/stack.mjs
+++ b/packages/core/src/stack.mjs
@@ -15,6 +15,7 @@ export function Stack(name = null) {
   this.bottomRight = false
   this.width = false
   this.height = false
+  this.anchor = new Point(0, 0)
 
   return this
 }


joostdecock:
Merging this as-is for now, but I seem to remember that stacks are not properly getting anchored, and I suspect that's something we need to look into.
Then again, it's a bit niche, so not urgent.

joostdecock (approved):

BenJamesBen:
> Merging this as-is for now, but I seem to remember that stacks are not properly getting anchored, and I suspect that's something we need to look into. Then again, it's a bit niche, so not urgent.

I filed Issue #6000 for this.

PR #5992, opened 2024-02-08
fix(markdown): Add missing list property to Attributes and Defs docs

BenJamesBen:
It was unclear to me whether the omission of the `list` property from the Attributes and Defs docs was intentional or not. Because `list` is an Enumerable property, I thought that perhaps the omission was accidental.

But, if the omission was intentional, please just close this PR.

### DIFF
--- a/markdown/dev/reference/api/attributes/en.md
+++ b/markdown/dev/reference/api/attributes/en.md
@@ -12,6 +12,12 @@ Attributes new Attributes()
 
 The Attributes constructor takes no arguments.
 
+## Properties
+
+An Attributes object comes with the following property:
+
+- `list` : The of individual attributes.
+
 ## Methods
 An Attributes object exposes the following methods:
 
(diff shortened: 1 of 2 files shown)


markdown/dev/reference/api/attributes/en.md
```diff
@@ -12,6 +12,12 @@ Attributes new Attributes()
 
 The Attributes constructor takes no arguments.
 
+## Properties
+
+An Attributes object comes with the following property:
+
+- `list` : The of individual attributes.
```
SeaZeeZee, line 19: Is there a word missing here?

joostdecock: There is. List holds the internal object in which attributes are stored. So if we're documenting it, we should fix the description.

joostdecock (changes requested):

BenJamesBen:
Corrected the `list` property descriptions.

joostdecock (approved):

PR #5995, opened 2024-02-09
fix(markdown): In Part doc, add properties and methods

BenJamesBen:
Added properties and methods info to the Part doc.

(The methods are manually added because `<ReadMore />` would catch the config and draft directories.)

### DIFF
--- a/markdown/dev/reference/api/part/en.md
+++ b/markdown/dev/reference/api/part/en.md
@@ -7,6 +7,27 @@ Parts truly are the building blocks of FreeSewing as they not only provide
 the configuration, but also a `draft()` method that does the actual work
 of drafting a parametric design.
 
+## Properties
+
+A Part object comes with the following properties:
+
+- `attributes` : An [Attributes](/reference/api/attributes) instance holding
+  the part's attributes
+- `hidden` : When this is `true` the part will be hidden (excluded it from the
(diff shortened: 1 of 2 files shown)


markdown/dev/reference/api/part/en.md
```diff
@@ -7,6 +7,27 @@ Parts truly are the building blocks of FreeSewing as they not only provide
 the configuration, but also a `draft()` method that does the actual work
 of drafting a parametric design.
 
+## Properties
+
+A Part object comes with the following properties:
+
+- `attributes` : An [Attributes](/reference/api/attributes) instance holding
+  the part's attributes
+- `hidden` : When this is `true` the part will be hidden (excluded it from the
```
SeaZeeZee, line 16: I've spaced on how to alter the text here but I believe this should read 'excluding' rather than 'excluded'

SeaZeeZee (changes requested):
Other than the typo it looks great!

markdown/dev/reference/api/part/en.md
joostdecock: Feel free to submit a PR to fix that.

joostdecock (approved):

BenJamesBen:
"Excluded" typos fixed in #6013.

PR #5997, opened 2024-02-09
chore(markdown): Add docs for new Pattern* classes

BenJamesBen:
I'm filing this PR to inquire whether these new Pattern* classes should be documented? I get the impression that perhaps they were instead intended to be non-public (as far as documentation is concerned), and perhaps only `Pattern` is intended to be public?

(This PR currently consists of doc stubs with `<Fixme>` tags, but if we want a more substantial PR before the merge, I can switch it to Draft status, work on it further, and then resubmit it.)

### DIFF
--- a/markdown/dev/reference/api/patternconfig/en.md
+++ b/markdown/dev/reference/api/patternconfig/en.md
@@ -1,0 +1,15 @@
+---
+title: PatternConfig
+---
+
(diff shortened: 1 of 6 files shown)


joostdecock:
No, I don't want these classes documented. Because, they are internal. But also because I am not happy with them, and it's vaguely on my todo list to rewrite this part of the code.

PR #6014, opened 2024-02-11
fix(markdown): In Windows tutorial, change link from .node-version to .nvmrc

BenJamesBen:
Corrected a 404 link. The previous `.node-version` file was removed some time ago.

_(I'm still thinking about a long-term fix and messaging for the acceptable/recommended Node.js versions issue. It would be easier to maintain if we had this documented in just one place, instead of in every tutorial.)_


joostdecock (approved):

joostdecock:
@BenJamesBen I've set the required version in `package.json`: https://github.com/freesewing/freesewing/pull/6015/commits/59adf43911f22893e7549507472ba117f2aabaf4

So the best thing to do is probably to refer to that, since `.nvmrc` is specific to nvm, and there's no obligation to install Node this way.

Documentation for the `engines` field in `package.json` is here: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

PR #6033, opened 2024-02-13
fix(shared): Show missing measurements for sets for new patterns

### LINKED ISSUE #5822: [feature]: Make it clear when some measurements are missing

When only a few measurements are missing, we should list them so it's more clear to users what the problem is, or why they can't use a certain measurements set for a given design.

### Additional context

See #5819 for an example where this would be helpful

BenJamesBen:
This was the best I could come up with, in terms of UI.
Screenshot 2024-02-12 at 5 18 50 PM

Fixes #5822.

### DIFF
--- a/sites/shared/components/account/sets.mjs
+++ b/sites/shared/components/account/sets.mjs
@@ -139,14 +139,31 @@ export const MsetCard = ({
     wrapperProps.style.backgroundPosition = 'bottom right'
 
   let icon = <span></span>
+  let missingMeasies = ''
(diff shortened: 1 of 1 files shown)


joostdecock (approved):

joostdecock:
I've merged this as-is because it's obviously an improvement. However, there seems to be some hard-coded English, so we'd have to fix that at some point.

BenJamesBen:
PR #6050 filed to add translations.

Issue #5874, opened 2024-01-23
[bug]: Inch and eighth symbols are missing from exported PDF


BenJamesBen:
### What seems to be the problem? 🤔

When patterns are exported to PDF, the 1/8, 3/8, 5/8, 7/8, and " (inch) symbols are missing from them.

In the below example, the symbols are missing from the scalebox dimensions on the PDF, while they are correctly present in the SVG.

PDF:
![IMG_0651](https://github.com/freesewing/freesewing/assets/109869956/e128d42b-4ccf-49cd-b82c-59cc0bac96d2)

SVG:
![IMG_0652](https://github.com/freesewing/freesewing/assets/109869956/00e232d3-bf69-4026-9948-32b30abd4b94)

### Are you a FreeSewing patron? 😃

Yes, I am a tier-2 patron ❤️

### Additional context

This issue was reported on Discord:
https://discord.com/channels/698854858052075530/944926520282054666/1199059190707134594
discord://discord.com/channels/698854858052075530/944926520282054666/1199059190707134594

BenJamesBen:
PR #6047 was filed to provide a temporary workaround for this issue.

PR #6047, opened 2024-02-13
fix(plugin-annotations): Use non-special characters in scaleboxes to work around #5874

BenJamesBen:
This is just a workaround, intended to be applied temporarily until a fix is found for #5874. (The better thing to do would be to actually fix the issue, but I don't think I have the ability to do that.)

The workaround basically just uses plain ascii characters instead of the special characters for eighth fractions and the inch symbol.

Screenshot 2024-02-13 at 9 55 28 AM


joostdecock (approved):

joostdecock:
I am starting to think that it was a mistake to move PDF generation browser side.
Because this is a problem that requires us to embed a font in the PDF, and that would be so much easier to handle server-side.

I'll have to think about it a bit. For now, this works.

PR #6076, opened 2024-02-17
fix(shared, plugins): Provide measurement set name to patterns

### LINKED ISSUE #6051: [bug]: Patterns from measurement sets show "ephemeral" in titles

When a pattern is made using a measurement set, instead of showing the name of the measurement set in the `title` macro, the word "ephemeral" is displayed.
Screenshot 2024-02-13 at 7 00 18 PM

_No response_

### Additional context

This issue started when the switch to v3 was made.

The `title` macro gets the name of the measurement set from `store.data.for`. It appears that the v3 frontend might not be saving this information in the store when a measurement set is selected.
In `plugins/plugin-annotations/src/title.mjs`:
```js
    .addText(store.data?.for ? store.data.for : 'ephemeral')
 ```

BenJamesBen:
Closes #6051.

This PR:
- Passes measurement set name to the pattern, so the `title` macro can display it.
- Also adds the measurement set name to the PDF cover sheet.
- Makes some changes to the PDF cover sheet, how things are formatted/presented.

Please note that this PR introduces a new `settings.metadata.setName` setting which is not a part of Core. I felt that I had to do this to be able to provide the set name to the export handler. Otherwise the `title` macro would display "ephemeral" on PDFs. (If there is a different way to do this without introducing the non-official, non-Core setting, this PR can be rewokred to do so.)

Screenshot 2024-02-16 at 5 40 55 PM
Screenshot 2024-02-16 at 5 41 33 PM
Screenshot 2024-02-16 at 5 43 19 PM

### DIFF
--- a/plugins/plugin-annotations/src/title.mjs
+++ b/plugins/plugin-annotations/src/title.mjs
@@ -142,7 +142,7 @@ const title = function (config, { Point, points, scale, locale, store, part, log
 
       `${mc.classes.name} ${mc.align}`
     )
-    .addText(store.data?.for ? store.data.for : 'ephemeral')
+    .addText(store.data?.setName ? store.data.setName : 'ephemeral')
     .addText(')')
     .attr('data-text-transform', transform)
     .attr('data-render-always', 1) // Render even when outside the part bounding box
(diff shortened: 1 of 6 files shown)


joostdecock (approved):

joostdecock:
Nothing wrong with overloading the `settings` object.

PR #6077, opened 2024-02-17
fix(shared): Add set names to curated set buttons

### LINKED ISSUE #5898: [feature]: Display curated measurement set names

To make it easier to discuss patterns made using curated measurement sets, I would like it if each set had its own name or ID number shown to users. This would be a public-facing name or id, not necessarily the actual name of the model.

That way, I could say things like "To reproduce the problem, use Charlie with curated measurement set 'Delta`" or "I tested using curated measurement set # 302". Instead, right now I would have to say "using the 2nd person from the left" or "the guy with the orange shirt" which might be ambiguous as we add more measurement sets.

Yes, I am a tier-2 patron ❤️

### Additional context

I was able to mock up something with the below code change, to show id numbers.

```diff
--- a/sites/shared/components/curated-sets.mjs
+++ b/sites/shared/components/curated-sets.mjs
@@ -76,7 +76,16 @@ const SetLineup = ({ sets = [], href = false, onClick = false }) => (
       if (onClick) props.onClick = () => onClick(set)
       else if (typeof href === 'function') props.href = href(set.id)
 
-      if (onClick) return <button {...props} key={set.id}></button>
+      if (onClick) return (
+        <div>
+          <div>
+            <button {...props} key={set.id}></button>
+          </div>
+          <div>
+            <span class="text-2xl">&nbsp;&nbsp;&nbsp;&nbsp;#{set.id}</span>
+          </div>
+        </div>
+      )
       else if (href) return <Link {...props} key={set.id}></Link>
       else return <div {...props} key={set.id}></div>
     })}
```

Screenshot 2024-01-25 at 9 22 10 AM

BenJamesBen:
Closes #5898.

This was the best I could come up with. Someone better at CSS positioning might be able to do better.
- There is a `text-center` Tailwind class that makes the text alignment looks much better. However, that class doesn't work with `absolute` positioning.
- I had to use `absolute` positioning because otherwise the additional text would alter the vertical alignment of the image models, causing the heights of the models to be incorrect compared to the height marker background.
- So, I used `&nbsp;` characters to try to manually position the text, which is not a great thing to do.

Screenshot 2024-02-16 at 5 55 29 PM

### DIFF
--- a/sites/shared/components/curated-sets.mjs
+++ b/sites/shared/components/curated-sets.mjs
@@ -76,7 +76,20 @@ const SetLineup = ({ sets = [], href = false, onClick = false }) => (
       if (onClick) props.onClick = () => onClick(set)
       else if (typeof href === 'function') props.href = href(set.id)
 
-      if (onClick) return <button {...props} key={set.id}></button>
+      const { i18n } = useTranslation(ns)
+      const lang = i18n.language
+      if (onClick)
+        return (
+          <div>
+            <div>
+              <button {...props} key={set.id}></button>
+            </div>
+            <div className="text-xl absolute">
+              <span>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span>
+              <span>{set[`name${capitalize(lang)}`]}</span>
+            </div>
+          </div>
+        )
       else if (href) return <Link {...props} key={set.id}></Link>
       else return <div {...props} key={set.id}></div>
     })}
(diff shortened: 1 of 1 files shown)


joostdecock:
I appreciate the effort here @BenJamesBen but I agree that spacing with `&nbsp;` is not great.

So I've put up #6080 to implement this. It also implements it for all views, not just for the button view.

Closing in favor of #6080

PR #6113, opened 2024-02-21
DRAFT: fix(markdown): Add details to User Experience doc

BenJamesBen:
After:
Screenshot 2024-02-21 at 9 33 04 AM

Before:
Screenshot 2024-02-21 at 8 53 30 AM


BenJamesBen:
(Filed as a Draft PR, to get feedback before the PR is ready to be submitted.)

BenJamesBen:
Added an edit after taking the screenshot changing "Locale" to "Language" to match the wording used in the UI.

joostdecock:
Note that level 5 does not add features but removes guardrails (like asking 'are you sure' before removing things).

joostdecock (approved):

joostdecock:
We can always revisit this later should people have concerns, but this is a good change. I don't see a need to keep it in draft.

PR #6134, opened 2024-02-23
fix(tests, charlie): Add menu translation keys

BenJamesBen:
This PR includes a change to the design i18n tests to allow design menu translation keys. If `o.<key>.d` is set to `menu`, it means that the item is menu name and not an actual option. This causes the test to check for an actual option to be skipped.

So, to add a translation key for a menu:
```json
   "o": {
    "myMenu": {
      "t": "My Menu",
      "d": "menu"
    },
```

This PR also adds 3 menu translations for Charlie.

### DIFF
--- a/designs/charlie/i18n/en.json
+++ b/designs/charlie/i18n/en.json
@@ -102,13 +102,16 @@
       "d": "Controls how curved the waistband is."
     },
     "backpockets": {
-      "t": "Back Pockets"
+      "t": "Back Pockets",
+      "d": "Back Pockets"
     },
     "frontpockets": {
-      "t": "Front Pockets"
+      "t": "Front Pockets",
+      "d": "Front Pocketsa"
     },
     "fly": {
-      "t": "Fly"
+      "t": "Fly",
+      "d": "Fly"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


BenJamesBen:
I'm thinking that using the string `menu` might cause confusion in Crowdin, if the translators mistakenly think that the word should be translated. it might avoid confusion if we were to use a different string like `M` or `M1` or `MMM`. Something that the translators will just leave as-is.

BenJamesBen:
Reworking this PR to not use the "o" section in the i18n json files.

joostdecock (approved):

joostdecock:
> *Reworking this PR to not use the "o" section in the i18n json files.*

Didn't see this comment earlier but I appreciate the effort

BenJamesBen:
I am still working on this PR, though I've been sidetracked the past few weeks.
I think that the idea is to add a new `m` top-level property to the JSON files for menu translations, with `t` and `d` sub properties (since the UI code already works with `.t` text keys). This will require changing docs, `utils.mjs`, and prebuild scripts, and adding tests.

BenJamesBen:
Closing this PR to work the issue offline. When the code changes are ready, I will file a new PR or re-open this one.

PR #6211, opened 2024-02-27
adding photos on cornelius documentation

PtitLutinn:
I had put pictures with my camera during a cornelius making and I add it on the doc for next people asking for help


joostdecock:
Thank you for improving our documentation @PtitLutinn 

However, before merging this, we need to put these images on Cloudflare (our image hoster).
It's something I can do, but I need to find some time for it.

Merging this as-is would add about 300Mb to our repo, and we need to be mindful that that would impact all contributors, so better not to.

I will keep this open until I or someone else has the time to add these to cloudflare.

PtitLutinn:
Is adding the images to cloudflare something that I can do ?

PR #6221, opened 2024-02-28
Add user71677 Author

bobgeorgethe3rd:
#6210


bobgeorgethe3rd:
Ok can it be checked if this user id matches bjeppson before merging. If it does great if not we've been having some false claims.

bobgeorgethe3rd:
Confirmed that there has been some false claiming. Still unclear on this one so leaving open for now.

joostdecock (approved):

bjeppson:
My username is Buck. Does that answer your question sufficiently?

PR #6227, opened 2024-02-29
fix(org): Correct showcase author claim label

BenJamesBen:
I am not sure, but I think that this code change might fix the incorrect label?

Before:
Screenshot 2024-02-28 at 9 17 19 PM


bobgeorgethe3rd:
I understand the urge to fix this but can we can consider that is useful to sort out the author issues. That's what I have been using it for as it makes easy to quickly sort them and then do them in batches.

joostdecock (approved):

joostdecock:
@bobgeorgethe3rd It's obviously a badly encoded label. So while it's a handy side-effect that this label exists only on author claims, it's hardly a reason not to fix something that is clearly not the intended behaviour.

So instead, I'll add an additional `author claim` label so you can continue to filter based on this.

PR #6230, opened 2024-02-29
fix(markdown): In Jargon guide, add freesewing.org translation process

BenJamesBen:
(I came up with the best procedure I could think of, but perhaps there is a better one. Maybe a procedure that would get the needed translations before the additions are made to the jargon file. Or, perhaps there could be a technical solution splitting up the jargon into individual language JSON files, allowing them to be translated through the regular Crowdin process.)

Screenshot 2024-02-29 at 7 12 53 AM

### DIFF
--- a/markdown/dev/guides/markdown/jargon/en.md
+++ b/markdown/dev/guides/markdown/jargon/en.md
@@ -40,6 +40,36 @@ const jargon = {
 }
 ```
 
+### Adding freesewing.org jargon translations
+
+Because the freesewing.org jargon file contains keys for all supported
+languages, a key/value pair for each supported language must be added
+to the file when adding a new jargon term.
+To do this, you will need to know the jargon term's translation in
+each language, information which you may not have initially.
+
+**Because the jargon file itself does not get translated as part of the
+regular FreeSewing translation process, it will be your responsibility
+to obtain the translations and add them to the file manually.**
+We have developed a process to help you do this:
+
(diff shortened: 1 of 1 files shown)


joostdecock:
@BenJamesBen I appreciate the push to clarify how this works, but I feel it's leading us too far into territory that is outside the scope of a markdown guide.
That is of course entirely my own doing as the current version of the guide includes a section on *How to add jargon* and so it makes sense that you wanted to update that to reflect the recent changes on how we handle jargon.

However, I think we should split this up based on the audience:

- How to use jargon: This should be in the guide, and needs to be documented as writing docs is something plenty of people contribute to.
- What it takes to become a jargon term: This I think can still be included in the guide. Essentially saying: You need to create a specific documentation page holding the definition
- How to add jargon: This is something that hardly any contributor does (*) and falls outside the scope of this markdown guide.

(*) as far as I know, all the jargon terms that exist are both added and written by myself

To further demystify this process, I've put up a PR #6242 that adds an overview component that list the terms available, shows what they look like when used, and links to the documentation page that is backing them.

If you agree, I would propose to remove documentation on how to *add* jargon terms, and focus on how to use them, and link to the ones who exist. (and perhaps explain that any doc page can be made a term).

joostdecock:
@BenJamesBen I have removed the **How to add jargon** section from the markdown guide, as I feel it's out of scope.
Instead, I've added a link to the (new) page where you can find all the terms one can use.

As such, I'm closing this. Thanks for bringing this up.

PR #6240, opened 2024-03-01
fix(uma): fix back exposure if expand is off

### LINKED ISSUE #6239: [bug]: Uma Expand option removes low back exposure special handling

When I generate Uma with back exposure lower than 25%, the shape of the uma.back part gets special handling that flares out the buttock area:

https://github.com/freesewing/freesewing/blob/6b8f8bd6e576f5f8be4601b9dc037d142e9c1267/designs/uma/src/base.mjs#L386-L390

This flare out only happens when expand is enabled, else it doesn't have any effect:

Bildschirmfoto vom 2024-03-01 10-16-15
Bildschirmfoto vom 2024-03-01 10-16-06

Both images were created with identical settings, I only toggled expand.

designs/uma

### Additional context

_No response_

HaasJona:
Fixes #6239

The main fix is the reuse of the existing path in designs/uma/src/back.mjs instead of building an (inferior) copy, the rest is minor code style improvement or removal of useless code (in front.mjs)

### DIFF
--- a/designs/uma/src/back.mjs
+++ b/designs/uma/src/back.mjs
@@ -120,14 +120,13 @@ function draftUmaBack({
     // Also Y-flip these central points
     points.cfWaistbandDipBack = points.cfWaistbandDipBack.flipY()
     points.cfBackGusset = points.cfBackGusset.flipY()
-    paths.saBase = new Path()
-      .move(points.cfWaistbandDipBack)
-      .curve_(points.cfWaistbandDipCpBack, points.sideWaistbandBack)
-      .line(points.sideLegBack)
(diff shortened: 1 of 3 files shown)


joostdecock (approved):

HaasJona:
Sorry, forgot a missed hide() call, should re ready now.

joostdecock (approved):

joostdecock:
@all-contributors please add @HaasJona for code

joostdecock:
@HaasJona Thank you so much for this fix :pray: 

You will/should receive an invite to join the FreeSewing organisation on GitHub.
Accepting the invite is optional, but doing so will give you triage permissions on our repository.

Thank you for your contribution, much appreciated :medal_sports:

PR #6332, opened 2024-03-10
Add Umbra, an underwear pattern based on Uma

HaasJona:
Umbra is an underwear design I've been working on. It's based on some of my own patterns and measurements of my favourite briefs. I have sewn some prototypes already.

This is not completely polished, but I'm putting it up for review already to get some feedback from both a development and pattern design perspective.

umbra

image

### Summary from designer notes:

Umbra is a comfortable and flexible underwear briefs design that aims to be inclusive for as many people as possible.

Umbra is based on some experiments, prototypes and measurements of my favourite underwear.

Umbra is especially designed for diverse users. You can easily add several useful modifications, even if some of them are not directly supported by the software.

Some features of Umbra are as follows:

* It can be sewn so that no (potentially itchy) seam allowance touches the skin, which helps people with sensitive skin and those with autism. The crotch and side seams are placed on the inside between the front part layers and any elastics on the waist and leg openings can be put on the outside, if so desired.
* It can be sewn with or without a bulge. You can even draft the design twice, and cut the front lining without bulge and the main fabric front part with one, and you'll be able to create a front pocket with plenty of space.
* Rise, back exposure, front exposure and gusset width are controlled separately, so you can create e.g. very comfortable thongs with enough room in the front, or many other designs.

Umbra is based on code from Uma by Natalia Sayang and Joost De Cock, but has been heavily modified in its construction and default style.
Some key differences are:
* Umbra doesn't have a separate gusset part and instead adds lining to the whole front piece.
* The hip measurements are being taken into account to better determine the waistband size.
* The height of the underwear is based on the usually more accurate Cross Seam measurement instead of the Waist to upper leg measurement.
* The positions of the side and crotch seams can be adjusted freely, even with a bulge selected.
* The back part has a smoother, more comfortable curve and provides greater coverage by default and at maximum.

If you wish to adapt Umbra to a swimwear bottom, it should be easy to do so. Simply use swimwear fabric and, if desired, sew on a folded over piece of fabric with two holes instead of a waistband elastic, to create a drawstring.

### Development notes:

* Umbra's code is based on Uma, although there's not much original, untouched code left. 
* I chose the name Umbra because it starts with an U for underwear, it's similar to Uma to show its heritage, and Umbra is latin for "shadow", which sounds nice and mysterious and fits underwear. However, if you prefer another name, I wouldn't mind much
* I thought about expanding Uma instead, but the differences are very large and integrating all the options into one part would be confusing.
* It might make sense to extract a common underwear block body and extend both this and Uma from this, but as this was my first design, I avoided doing a large rework. Might be a good idea for the future, but will probably create some compatibility questions
* It might also make sense to port some new features or fixes of this to Uma


HaasJona:
Note: Unsure what to do with the Changelog as this is new and has no changes. Copied from Uma for now.

joostdecock:
The changelog is auto-generated grom `config/changelog.yaml` so don't worry about it 🤗

Didn't get around to reviewing this today as I was otherwise occupied, but I'll try to find time for it soon.

Meanwhile, thanks for this PR 🙏

HaasJona:
> The changelog is auto-generated grom `config/changelog.yaml` so don't worry about it 🤗

Good to know. By the way, there is a typo in the auto generated text:

- All FreeSewing **pacakges** are now ESM only.
 - All FreeSewing **pacakges** now use named exports.

HaasJona:
Converting this to a draft for now as I want to improve the back curve calculation and check some measurements again.

designs/umbra/src/front.mjs
```diff
@@ -0,0 +1,226 @@
...
+        .setClass('fabric sa')
+        .unhide()
+    }
+  }
+  store.cutlist.setCut({ cut: 1, from: 'fabric', onFold: !expand })
+  store.cutlist.addCut({ cut: 1, from: 'lining', onFold: !expand })
+  points.title = points.cfWaistbandDipFront
+    .shiftFractionTowards(points.sideLegFront, 0.14)
+    .shiftFractionTowards(points.cfMiddle, 0.25)
+  if (!expand) {
```
joostdecock, line 94: This is perhaps a nitpick, but I find:

```mjs
if (something) {
  // ...
} 
else {
  // ...
}
```

more intutitive than:

```mjs
if (!something) { 
  // ...
} 
else {
  // ...
}
```

designs/umbra/src/front.mjs
```diff
@@ -0,0 +1,226 @@
...
+  if (options.frontDip !== 0) {
+    macro('vd', {
+      id: 'outer2',
+      from: points.cfWaistbandDipFront,
+      to: points.sideLegFront,
+      x: points.sideSeat.x,
+    })
+  }
+
+  macro('vd', {
```
joostdecock, line 154: This lacks an `id`

designs/umbra/src/front.mjs
```diff
@@ -0,0 +1,226 @@
...
+    from: points.cfWaistbandDipFront,
+    to: points.sideWaistbandFront,
+    x: -15,
+  })
+
+  macro('vd', {
```
joostdecock, line 160: This lacks an `id`

designs/umbra/src/front.mjs
```diff
@@ -0,0 +1,226 @@
...
+    from: points.cfWaistbandDipFront,
+    to: points.backGussetSplitBulge,
+    x: points.backGussetSplitBulge.x,
+  })
+
+  macro('hd', {
```
joostdecock, line 166: This lacks an `id`

designs/umbra/src/front.mjs
```diff
@@ -0,0 +1,226 @@
...
+    from: points.backGussetSplitBulge,
+    to: points.sideLegFront,
+    y: points.backGussetSplitBulge.y,
+  })
+
+  if (options.bulge >= 2) {
+    macro('hd', {
```
joostdecock, line 173: This lacks an `id`

joostdecock (commented):
On first glance, LGTM :+1:

designs/umbra/src/back.mjs
```diff
@@ -0,0 +1,220 @@
+import { base } from './base.mjs'
+
+export const back = {
+  name: 'umbra.back',
+  from: base,
+  draft: draftUmbraBack,
+}
+
+/*
+ * This drafts the front of Umbra, or rather recycles what's needed from base
```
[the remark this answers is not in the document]
HaasJona: That error was copy-pasted from Uma

designs/umbra/src/front.mjs
HaasJona: You're right

BenJamesBen:
Although Umbra seems to be in the middle of development with things not fully implemented and changing quickly, I did some initial testing in the localhost lab.

- I note that the Expanded Back part has no horizontal dimensions.
- The seam allowance is on the inside of the Back part, when Flip Back is false.
- The Back part title appears in the old, wrong location when Flip Back is false.
- I was able to induce a crash:
```
ERROR | TypeError: undefined is not an object (evaluating ‘points.pocketSeamBottom.x’)
We have also logged this error to the console of your browser so you can inpect it further.
ERROR | Unable to draft part umbra.front (set 0)
WARN | snippets.pocketBottom was set with an anchor parameter that is not a Point
```
```yaml
measurements:
  waist: 730
  seat: 960
  waistToSeat: 200
  crossSeam: 620
  crossSeamFront: 280
  waistToHips: 90
  hips: 880
units: metric
metadata:
  setName: Elanor
sabool: 1
sa: 15.3125
samm: 15.3125
paperless: true
options:
  backDip: 0.089
  backExposure: 0.45
  frontDip: 0.111
  frontExposure: 0.947
  frontReduction: 0.234
  legRise: -0.5
  rise: 0.764
  pockets: inside
  pocketHem: 0.125
```

HaasJona:
Thanks for the review. The first one was already fixed (as you noticed). I've fixed the other issues, although the flip back one was intended as a developer and i don't consider the seam allowance and title position as important when in "developer mode". But fixed anyways.

HaasJona:
So some general comments about the state: I overhauled some of the construction after helpful discussions on discord and some prototype results. I'm mostly okay with the shape now, but if you have any suggestions, feel free to discuss them. The code should be mostly ready, but I'm aware that some of the point names and so on are not really self-explanatory and the scheme they follow isn't perfect.

I'll be sewing another prototype later and I'll cleanup and squash the commits and do a final code pass. Other than that, this should be ready for review.

@joostdecock Some of the changes could be backported to Uma, like some of the new options (flip back, front reduction), the usage of the cross seam and cross-seam-front measurement instead of waist-to-upper-leg (maybe optional, like the legacy armhole option on brian), what do you think about that?

BenJamesBen:
Not necessarily issues but some behaviors I saw while testing:

- Pocket lines (Zipper and Fold lining) are visible when Details are disabled.
- One set of "Zipper" and "Fold lining" text is upside down due to the mirroring. I did some testing to see how the text could be finagled to be rightside up, but I'm not sure if it's worth going through all the trouble to do so.
```diff
+    const mid = (paths.mirroredZipperCut.start().x + paths.mirroredZipperCut.end().x) / 2
+    macro('mirror', {
+      mirror: [new Point(mid, 0), new Point(mid, 100)],
+      paths: ['mirroredZipper'],
+      clone: true,
+    })
+    paths.mirroredZipper = paths.mirroredMirroredZipper.clone()
+    delete paths.mirroredMirroredZipper
```
- The below behavior is also seen with Uma and its predecessors, but it is possible to pull Umbra's parts into impossible geometries.
```yaml
measurements:
  waist: 920
  seat: 1075
  waistToSeat: 280
  crossSeam: 900
  crossSeamFront: 415
  waistToHips: 145
  hips: 960
units: metric
metadata:
  setName: Joost
options:
  pockets: inside
  pocketHem: 0.181
  rise: 0.2
  pocketGap: 0.173
  legRise: 0.8
  frontExposure: 1
  frontDip: 0.15
  backExposure: 1
  backDip: 0.1
  bulge: 11.1
  bulgeFullness: 0.679
  gussetPosition: 0.5
  gussetWidth: 0.03
  splitPosition: 0.259
sabool: 1
sa: 15.3125
samm: 15.3125
```
Screenshot 2024-03-17 at 9 40 11 AM

HaasJona:
The zipper and fold text I'm aware of. Instead of mirroring, it would probably be simpler to just construct the left paths manually in an appropriate way. Maybe I'll look at this later.

About the last case, you're combining several very extreme options. Some of the options (mainly the gusset position, which I added very late) should probably have a more restricted range, on the other hand, without something like #6335 present, this could be restricting for some use cases. I'm not sure if it's worth it to change something here. At least it doesn't crash.

HaasJona:
Current result looks like this:
grafik

Todo: 
* thong string width probably shouldn't be hardcoded to 2cm
* Maybe enforce minimum thickness for waistband part

HaasJona:
These particular settings look like this now:
grafik

(I renamed pocketHem to pocketHeight)

HaasJona:
I'm happy with this now. Any reviews are appreciated.

designs/umbra/src/base.mjs
```diff
@@ -0,0 +1,813 @@
+import { stretchToScale } from '@freesewing/core'
+import { formatPercentage } from '@freesewing/shared/utils.mjs'
```
joostdecock, line 2: This does not seem like a valid import.

designs/umbra/src/front.mjs
```diff
@@ -0,0 +1,232 @@
+import { base } from './base.mjs'
+import { units } from '@freesewing/core'
```
joostdecock, line 2: Not sure why you're using `units` from core here, rather than deconstruct like you did in base.

The difference is that the deconstructured `units` is aware of the units requested by the user, and will format accordingly. The `units` from core does not know, and so you need to tell it what units to use (or use the default, which is metric).

Check how `units` is loaded in `base.mjs`, and you should do the same here, unless there's some specific reason that I'm missing.

designs/umbra/src/base.mjs
joostdecock: Note that the `formatPercentage` method is a one-liner:

```mjs
const formatPercentage = (val) => Math.round(1000 * val) / 10 + '%'
```

You might just want to inline it.

designs/umbra/src/front.mjs
HaasJona: > Not sure why you're using units from core here, rather than deconstruct like you did in base.

Probably because I applied a quickfix from the IDE instead of looking at the proper way.

HaasJona:
I fixed both things and rebased it onto develop.

config/software/designs.json
```diff
@@ -909,6 +909,22 @@
       "curvedSeam"
     ]
   },
+  "umbra": {
+    "code": "Joost De Cock, Jonathan Haas",
+    "description": "A FreeSewing pattern for a basic, highly-customizable underwear pattern",
+    "design": "Joost De Cock, Jonathan Haas",
```
HaasJona, line 915: Do you want to be credited here? I have no problem leaving that in, but I'm not sure you want to be credited here, especially considering that you just took over Uma yourself.

designs/umbra/src/base.mjs
```diff
@@ -0,0 +1,814 @@
...
+
+  paths.back = new Path()
+    .move(points.cfWaistbandDipBack)
+    .curve(points.cfWaistbandDipCp2Back, points.cfWaistbandDipCp1Back, points.sideWaistbandBack)
+    .join(paths.elasticLegBack)
+    .line(points.cfBackGusset)
+    .hide()
+
+  // Determine center position of bulge
+  // TODO: Add measurement or make adjustable?
```
HaasJona, line 375: This is a minor open TODO, I don't think it should be blocking the merge, and I don't see a problem addressing it later,  just wanted to make you aware of it.

config/software/designs.json
joostdecock: I don't mind, but I also don't want people to think they have to credit me just because they are re-using something I wrote/designed.

So feel free to take my name off if you want.

joostdecock (approved):

PR #6375, opened 2024-03-17
DRAFT: fix(core): Add non-cloneable attributes functionality

BenJamesBen:
Even though the mirrored paths' Points are cloned, Uma displays warning messages for 2 points:
Screenshot 2024-03-17 at 6 35 42 AM

To fix this issue, this PR adds non-cloneable attributes (designated by having `noclone` in the key) that are filtered out during a `clone()` operation. It also makes the `plugin-mirror`'s `data-mirrored` attribute non-cloneable. The idea is that cloned Points should be considered to be fresh, new Points that haven't been mirrored before.

(There are other ways this issue could be fixed, like renaming non-cloneable attributes to "get rid" of them while also preserving them. Or, setting `data-mirrored` count back to zero. But, I chose the method that has non-cloneable attributes that simply get omitted.)

(There is another bug in `plugin-mirror` where it uses the non-existent `point.name` in the warning message, resulting the `Point undefined` wording. But, I think I want to handle it in a separate PR.) 

Creating as a Draft PR for review, to see if the idea is okay. Also, I need to add tests for this new functionality.

### DIFF
--- a/markdown/dev/reference/api/attributes/clone/en.md
+++ b/markdown/dev/reference/api/attributes/clone/en.md
@@ -21,3 +21,7 @@ const attr = new Attributes()
 const cloned = attr.clone()
 ```
 
+## Notes
+
+Individual attributes with a `key` containing the string `noclone` will not
+be cloned and will be omitted from the new Attributes object.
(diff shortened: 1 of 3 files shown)


BenJamesBen:
Added tests and moved out of Draft status.

joostdecock:
Let's put this in v4 so we are free to do what we like

BenJamesBen:
I cleared the merge conflict, so this PR should be set to go once we're ready for v4.

PR #6404, opened 2024-03-20
Added links to sew along on Vimeo since the blog post is currently down.

karlnippoldt:


designs/simon/src/front.mjs
```diff
@@ -141,7 +141,7 @@ function simonFront({
    * Annotations
    */
   // Cutlist
-  store.cutlist.setCut({ cut: 1, from: 'farbic' })
+  store.cutlist.setCut({ cut: 1, from: 'main fabric' })
```
joostdecock, line 144: `fabric` is a translation key. That's what needs to be used here, so that it gets properly translated

joostdecock (changes requested):

joostdecock:
@all-contributors please add @karlnippoldt for docs

joostdecock:
Thanks for this @karlnippoldt 🙏 

You will receive an invite to join the FreeSewing organisation on GitHub.
Doing is is optional, but if you accept the invite you will be granted triage rights on this repo.

PR #6407, opened 2024-03-20
fix(jaeger,paco,sandy): Typo

BenJamesBen:
(After seeing the "farbic" typo fix in #6404, I searched for the same typo and found these 3 others.)


joostdecock:
Did you investigate who is the dyslexic idiot who can't spell?

Because I have a strong suspicion I know 😂

PR #6424, opened 2024-03-23
fix(huey): Zipper length calculation, clarify it is a maximum

joostdecock:
This stems from the excellent work by @HaasJona in #6421 in which they requested to double-check the zipper length calculation. Which I didn't get around to doing in the code review, but was able to now.

Turns out there was a slight issue with it, so this resolves it. I've also updated the wording to clarify that what we are calculating as the __maximum__ zipper length. Since zippers typically come is standard lengths, it's typically not possible to find the exact length you need, so it's IMHO important to make it clear to the user they should always purchase a zipper that is as close as possible, yet shorter than the calculated maximum length.


config/changelog.yaml
```diff
@@ -22,6 +22,8 @@ Unreleased:
   Fixed:
     huey:
       - Fixed an issue where ribbing height would distort the pocket under certain conditions. Fixes #6238
+      - Adding missing translation for rubbing option
```
HaasJona, line 25: ribbing not rubbing?

HaasJona (commented):
Looks good except for the typo. Didn't check the length calculation, but since you corrected mine, I assume you got it correct.

config/changelog.yaml
joostdecock: That's what she said :joy:

joostdecock: (I've fixed it)

PR #6484, opened 2024-03-31
Add Tina, an overlap top

HaasJona:
Add Tina, a simple top with front overlap. Inherits code from Teagan.

image

Documentation still missing, but the basic steps are:

* Shoulder seams
* Hem diagonal and neck seam (or use fusing tape)
* Attach sleeves
* Close seams below sleeves and at the side of the body.

It's probably easier to sew than Teagan as you don't need the neck finish, so I set the difficulty to 2. The tricky bit is the sleeves, but you can of course omit the sleeves if desired.

### DIFF
--- a/config/software/designs.json
+++ b/config/software/designs.json
@@ -863,6 +863,24 @@
     ],
     "techniques": []
   },
+  "tina": {
+    "code": "Jonathan Haas",
+    "description": "A FreeSewing pattern for overlap top",
+    "design": [
+      "Jonathan Haas"
+    ],
+    "difficulty": 2,
+    "lab": true,
+    "org": true,
+    "tags": [
+      "tops"
+    ],
+    "techniques": [
+      "curvedSeam",
+      "hem",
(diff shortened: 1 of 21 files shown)


woutervdub:
There seem to be language files included in the `tina` design that are copies of the `bibi` design. These should just be copies of the  `en.md` of `tina` if the correct translation is not part of this PR. For instance: `designs/tina/i18n/nl.json`

HaasJona:
> There seem to be language files included in the `tina` design that are copies of the `bibi` design. These should just be copies of the `en.md` of `tina` if the correct translation is not part of this PR. For instance: `designs/tina/i18n/nl.json`

Yeah, I'll get than cleaned up now that the Bibi changes have been merged and I can tease this branch properly onto develop.

But generally, I think it would be nice if there were a way to inherit option documentation from the parent design automatically (with the option to override it of course). Otherwise this feels like a lot of redundant copy and paste.

PR #6505, opened 2024-04-04
fix(teagan): Hide sleevecap path from brian

HaasJona:
This extra path causes extra-thick rendering of the sleevecap in brian (and maybe problems for laser cutters)


joostdecock (approved):
Nice catch 👍

PR #6512, opened 2024-04-05
add(core): Add Path.rotate(...) function analogue to Point.rotate(...)

HaasJona:
I found a function like this very helpful in my Umbra and Tina designs. Sometimes you can't practically rotate the individual control of a path points because the path itself can be composed of multiple lines/curves that might have been splitted, and so on.


joostdecock (approved):
Nice work @HaasJona  :clap: :clap: :clap:

PR #6527, opened 2024-04-08
fix(core): Path.offset(...) improvements

### LINKED ISSUE #6519: [bug]: path.offset(...) method sometimes throws an error

For example executing the following code will throw an exception:

`new Path().move(new Point(0,0)).curve(new Point(0.1, 0.1),new Point(0.2, 0.2),new Point(0.1, 1.1)).offset(1)`

```
Uncaught TypeError: paths[0] is undefined
    __joinPaths path.mjs:1309
    __pathOffset path.mjs:1431
    offset path.mjs:649
    <anonymous> debugger eval code:1
```

Seems to happen mostly with very short path segments.

These short segments are themselves mostly nonsensical of course, but sometimes occur as a result of a previous split(...) call.

packages/core

### Additional context

_No response_

HaasJona:
Path.offset(...) no longer fails on very short curves or zero length paths. Fixes #6519

Also fixed bbox calculation of another degenerate case (path objects without line or curve, e.g. as a result of cleaning up a very short path)

### DIFF
--- a/packages/core/src/path.mjs
+++ b/packages/core/src/path.mjs
@@ -192,6 +192,11 @@ Path.prototype.bbox = function () {
     if (op.to) current = op.to
   }
 
+  if (bbs.length === 0 && current) {
+    // Degenerate case: Line is a point
+    bbs.push(__lineBoundingBox({ from: current, to: current }))
+  }
+
   return __bbbbox(bbs)
 }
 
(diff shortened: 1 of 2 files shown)


joostdecock (approved):

joostdecock:
Nice work @HaasJona 👍

PR #6531, opened 2024-04-09
Manually input design options outside range

### LINKED ISSUE #6335: [feature]: enter values outside slider limit in text box

I would like to be able to type in a value for a slider that is outside the normal range (ie, range is 0 to 5 but I want -3) to set the slider to that value.  This functionality existed when v3 was in alpha but went away at some point.  This means that when I want values outside the range (with my user experience set to 5) I have to manually edit the url to set the value.

### Additional context

There are situations where the slider values are unreasonably restrictive.  For instance, I want my Aaron to have a low neck and short armscyes, which is basically the opposite of the default parameters.  If I set armhole drop to 0, I still get an armhole that is nearly 8" deep when what I need for the fit I want is 6.5".  I also find myself needing values outside the slider range for front armhole extra cutout (I want it above the provided range for Simon) and shoulder ease (I want it below the provided range for Simon).

VSinerva:
This is a proposed way to close #6335 

The idea is to only allow this at UX level 5, and only through "Edit settings by hand". This helps keep it the niche feature it should be, and avoids errors being generated from typos etc. in the live "Draft" view. (These came up as ideas in the discussion linked above)

All errors are still shown as before. The only difference is that at UX 5, if all errors are `RangeError` in `options`, the settings will be saved regardless. I believe this strikes a balance of not letting inexperienced (or experienced) users make mistakes during their normal workflow, while still giving experienced users increased control *when explicitly asked for.*

There were concerns raised in the discussion of the issue about this feature cluttering automatically generated bug reports with errors. However, I don't think that should be a deal breaker for this feature, because experienced users are already using and saving patterns with parameters out-of-range (by editing the URL). This would still be a 'hidden' niche option, so it's unlikely IMO to see much more usage than the URL-trick already does.

I'm of course open to feedback on the implementation, or on if this should be implemented at all :)


karlnippoldt:
I think this should be level 4 for consistency with how the user experience setting is used elsewhere. Per the documentation, there is no feature difference between level 4 and 5; the only difference is that on level 5 "`“Are you sure?” confirmation guardrails removed.`"

HaasJona:
I'm not sure this is the right approach. I never use the "Edit settings by hand" tab except for copy-pasting settings. The format is weird and the units are hard to read and there's no live preview.

If we allow entering out-of-bounds values at all, I think at least this editor should also allow setting out-of-bounds values:
grafik

It's probably simple, instead of making the textfield red and disallow the input, maybe make it orange and show a warning text on experience level 4 and 5.

However I find that if people find the need to exceed the slider range, that's probably something that needs to be fixed in the design itself in the first place by extending the defined range.

VSinerva:
> I think this should be level 4 for consistency with how the user experience setting is used elsewhere. Per the documentation, there is no feature difference between level 4 and 5; the only difference is that on level 5 "`“Are you sure?” confirmation guardrails removed.`"
> 
> For me, I'm not confident enough in FreeSewing to go without any warnings at all, but there are definitely situations where I'd want to manually input options outside the recommended range.

Based on the documentation for UX levels, I agree. Since Edit by Hand is only available at or above level 4 anyway, I'm gonna update this PR to reflect that. Basically, edit by hand would always allow values outside the given range, but at level 4 it would ask for confirmation, while skipping that at level 5

VSinerva:
> I'm not sure this is the right approach. I never use the "Edit settings by hand" tab except for copy-pasting settings. The format is weird and the units are hard to read and there's no live preview.
> 
> If we allow entering out-of-bounds values at all, I think at least this editor should also allow setting out-of-bounds values: grafik
> 
> It's probably simple, instead of making the textfield red and disallow the input, maybe make it orange and show a warning text on experience level 4 and 5.
> 
> However I find that if people find the need to exceed the slider range, that's probably something that needs to be fixed in the design itself in the first place by extending the defined range.
> 
> That said, I think this patch makes sense and should probably be pulled in, as it would at least help somewhat.

I don't necessarily disagree! The only complication is that currently the textboxes run validation+update on every change, so for level 4 that would need to be changed so it doesn't bombard the user with confirmations. That could probably be worked around by skipping the update step for out-of-range values, and only updating the pattern once the value is finalized+confirmed. I personally am not super interested in working on that rn, and I think there's value in implementing this first in a more limited scope, seeing as there were concerns raised about it :)

joostdecock:
Thanks for the heads-up and sorry for the delay.

I am refactoring the pattern editor to untangle it from the website code. In that refactor, entering values outside the allowed range is already implemented, and this obviously slipped my mind.

The work is in the `wb2` branch if you're curious (although it's very rough around the edges still).

My hope is to make it easier to contribute to the UI code, but also make it easier to customize the editor.

PR #6539, opened 2024-04-10
Unversion autogenerated files

GeniaHarrietBoeing:
I was working on a pattern and noticed that I had to add some autogenerated files to my commits, which I added to .gitignore.
There might be a reason to version these that I am not aware of.


joostdecock:
The reason we added these originally was because we wanted to make it easier for people to work with our monorepo.

However, as the amount of files that are generated in the (pre)build step grows, it tends to become a bit unwieldy, so I think we should indeed re-visit this decision.

That being said, we need to be somewhat careful to ensure we don't break anything. Both the website and the stand-alone development environment will load (some) files directly from GitHub so we need to see that they are still there.

I'm going to add this to my todo list to look into this. I'll keep this open for now.

vladh:
Hmm, is there a use case where these files wouldn't be generated anyway? For example, for me, these files seem to be generated by `yarn lab`. Presumably other kinds of builds could also generate them as needed. Just trying to figure out in which scenario one would need to have these files prebuilt, if any.

HaasJona:
> Just trying to figure out in which scenario one would need to have these files prebuilt, if any.

Other than the lab environment, I think the main thing to check would be the new design environment from this tutorial: https://freesewing.dev/tutorials/pattern-design which is even more bare-bones than the lab (only contains the block bodies and an empty design).

vladh:
> I think the main thing to check would be the new design environment

It looks like this PR poses no problems to the new-design environment, because this environment doesn't include the generated files this PR affects:

```
$ npx @freesewing/new-design
...
$ ls -la sites/lab/pages/new sites/lab/pages/account/patterns sites/shared/prebuild/data/
ls: cannot access 'sites/lab/pages/new': No such file or directory
ls: cannot access 'sites/lab/pages/account/patterns': No such file or directory
ls: cannot access 'sites/shared/prebuild/data/': No such file or directory
```

In fact, there seems to be no `sites/` directory at all:

```
$ ls -la sites
ls: cannot access 'sites': No such file or directory
```

(I did run into #6543, but I think that is unrelated.)

joostdecock:
To be clear, the stand-along development environment (what you get when your run `npx @freesewing/new-design` is different from our monorepo. But it uses code from our monorepo, obviously.

Some of the things it downloads are prebuild artifacts that are now in the repo.
So if we drop those (somethink I and I think everyone is in favor of) we need to make sure we keep the stand-along environment from working, and deliver those files some other way. Either some sort of published archive, or an NPM package.

That's why it's going to take some time to address this.

joostdecock:
FYI: I have started working on #6552 and will probably tackle this within that scope.

joostdecock:
I'm going to close this in favor of #6552 (which is on our roadmap now)

PR #6540, opened 2024-04-10
fix(backend): Check for public set. Fixes #6538

### LINKED ISSUE #6538: [bug]: Bookmarked measurement picker fails listing measurements from other users

When I bookmark a measurement set from another user, these doesn't appear in the bookmarked measurement set list when creating a pattern. Only your own measurement sets appear there (which is pretty useless as these appear under "your own measurement sets", too).

 Debugging reveals, that the measurement set is known, but then the browser tries to fetch it by requesting something like https://backend3.freesewing.org/sets/[id]/jwt, which fails authentication.

In `sites/backend/src/routes/sets.mjs` it looks like only yaml and json values may be downloaded from public sets from other people. But the jwt route is forbidden (`{"result":"error","error":"insufficientAccessLevel"}`).

I see the following options to fix it:
A) Allow public access for the jwt route
B) Make the measurement picker download the data using the yaml or json route instead

What would be the right approach?

sites/backend, sites/org

### Additional context

_No response_

joostdecock:


HaasJona (approved):
Looks good to me

PR #6547, opened 2024-04-12
Replace .setRender() with .hide()

GeniaHarrietBoeing:
The function Path.setRender() was used in the how-to "Add seam allowance and/or hem allowance". 
As far as I can tell this function doesn't exist anymore and isn't documented anywhere and I replaced it with Path.hide().

I am not quite sure if it makes sense to use Path.hide() here or if there is a more resource efficient option to not render the same path twice.


joostdecock (approved):

joostdecock:
Thanks @GeniaHarrietBoeing that was a v2 holdover

joostdecock:
@all-contributor please add @GeniaHarrietBoeing for docs

joostdecock:
@all-contributors please add @GeniaHarrietBoeing for docs

PR #6558, opened 2024-04-15
add(core): Add Path.angleAt(point) method

HaasJona:
Add a path.angleAt(point) function for one of my in progress designs, which returns the tangent of the path at that point (or false if the point isn't close to the line).

Useful for extending a path or adding a smooth control point after a split() operation.


HaasJona:
This might need some added docs and tests but first I wanted to check if this function is good for inclusion in the library.

joostdecock (commented):
It's a nice addition to the core library, but I'm not in love with `relativeOffsetOnCurve()` as a name. Mainly because `offset` is a term we use to draw a path parallel to the original. Whereas here you are using offset to refer to *walking* the Bezier from t=0 to t=1.

This overloading of the same term is likely to confuse people, including myself 6 months from now 😉 

So I'd prefer to see a name that is less likely to confuse people. Something like `localePointOnCurve()` perhaps? I'm open to alternative suggestions too.

HaasJona:
> I'm open to alternative suggestions too.

Maybe `relativeCurvePosition()` or `getCurveRatio()` or `curveParameterFromPoint()`. Just some random additional suggestions, but your proposal is also good.

HaasJona:
I've added some documentation and renamed the method to `curveParameterFromPoint`. I think that describes best what it does, because it neither returns the absolute offset nor the relative percentage-wise position when measured.

joostdecock (approved):
LGTM

PR #6610, opened 2024-04-23
add(bibi): Add Bibi, a FreeSewing pattern for a knit top body block

### LINKED ISSUE #6503: [feature]: Teagan should allow lower ease settings

When sewing it from knit fabric, it should be possible to use zero ease or even negative ease if you want to go for a skintight fit. Even when a t-shirt is usually supposed to be loose, you might benefit from being able to adjust the ease if you're re-using and modifying the pattern for your own designs. See for example this Discord post by Laser: https://discord.com/channels/698854858052075530/787402509543145532/1224200703241617509

Of course this could be also be fixed by something like #6335, or by implementing two slider ranges, one for the recommended values and one for the range that technically works without causing errors but which doesn't necessarily make sense for the original design.

### Additional context

_No response_

HaasJona:
Bibi is a design for a basic shirt/top/dress mostly for knit fabric. It has support for busts (doing a basic knitwear FBA with either fabric gathering at the side seam or an optional dart) and with a highly variable height and sleeve length. It can be sewn on its own, but it is mostly designed as a highly extensible building block for other designs like [my Tina top](https://github.com/freesewing/freesewing/pull/6484) or as a base for other custom designs.

Todo:

- [x] Check and maybe improve shape of the neck opening
- [x] Research and improve shape of straps with sleeveless option
- [x] Add paperless measurements
- [x] Documentation

Fixes #6503


HaasJona:
This is ready for review. I'll mark this as fixing #6503 as Bibi can fullfil that role instead of having to change Teagan.

Fixes #6503

joostdecock (approved):

PR #6623, opened 2024-04-25
Demonstrate alt intersects

anna-puk:
I'd like some feedback on these alternative methods for finding curve intersections (including line-intersects-curve).

For now, I've called my functions curvesIntersectAlt and lineIntersectsCurveAlt.

This method returns an intersection for the example provided in [issue 3367](https://github.com/freesewing/freesewing/issues/3367). (The algorithm is different, so this isn't a big surprise, and no guarantee that everything works.) This is demonstrated in the new `demo` design. (Use the options to shift the line around, or try TestDesign.)

Before I make an actual pull request, I'd like a pre-review, because:

a) the goal of the function is slightly different: a location where two curves almost touch (to within tolerance) is considered an intersection (I do believe 
b) I'm still getting used the syntax, so I probably wrote some parts really awkwardly

Feedback on both these points is very welcome!

For now, I've modified lineIntersectsCurve and curvesIntersect to become wrappers to the new function. Existing designs will therefore use the new algorithm. (Switch to the previous commit to undo this and keep the demo design and its use of the Alt functions.)


anna-puk:
@BenJamesBen and @woutervdub , I'd appreciate your feedback!

BenJamesBen:
I'm starting to take a look at this PR.

I note that I get a crash when `snapTol` is set to a certain value.
```
ERROR | TypeError: undefined is not an object (evaluating ‘this.log.error’)
```

```yaml
options:
  snapTol: 0.029
```

I also note that trying to generate some designs using curated sets, for example the Charlie design with the Elanor set, also now results in errors.

anna-puk:
I found (and fixed) a bug which appeared for certain angles between the line/curve(s) and it looks like this also fixes the crash on certain 'tol' values.

The `charlie`-for-Elanor issue remains (for now).

BenJamesBen:
I tested the updated code.

1. I note that sometimes 2 intersections are incorrectly found.
Screenshot 2024-08-27 at 12 07 07 PM
Screenshot 2024-08-27 at 12 08 57 PM

2. I am still able to cause crashes in the Demo design.
Screenshot 2024-08-27 at 12 14 24 PM
Screenshot 2024-08-27 at 12 15 39 PM
Screenshot 2024-08-27 at 12 15 49 PM

anna-puk:
Thanks for testing, @BenJamesBen!

I'm continuing to work on this; the task is more difficult than I initially thought, both in terms of the algorithm and in terms of figuring out what we actually need.

(For instance, given that there is a tolerance, finding two points for a single intersection isn't 'incorrect' per se, but it's not the desired behavior.)

Issue #6324, opened 2024-03-08
Bad new user experience report


alex-fu27:
My girlfriend had a very bad experience creating an account today and I have learned a lot about UX from that that I would like to share with you.

## TLDR
* Passwordless login, if not explained, seems confusing if you have been taught that a strong password was required for any website to be safe.
* Some parts of the account creation workflow are unclear or even confusing, at least in the German translation.
* The account creation workflow should not drop the user into the documentation, but have a page that reads something like *"Done! If you want, you can set up a password in your account settings, but you can also keep logging in through sign-in links. We will now direct you to the documentation so you can learn how to create sewing patterns!"*

## Story
To begin with, we speak German natively and she has absolutely zero technical background.
I had found freesewing.org a week earlier, and created an account, but set the language to English, so I had ended up on en.freesewing.org which I found out now hosts an older version of the website.

My girlfriend started by entering `freesewing.org` into the address bar on her latpop PC, I had given her the domain name. The site was correctly served in German. It was looking a little bit different than on my PC, which started the confusion.

She then entered her e-mail address into the singup form and had a sign-up link sent to her e-mail address. She found it weird that she didn't have to enter a password and we will come back to that later.
Thinking that this was a confirmation e-mail, she clicked the link on her phone (she doesn't usually do this on her PC because then she has to enter her password because for reasons of perceived security she always logs out of any website before she closes the browser tab).
That made the onboarding flow continue on her phone.

She worked through the wizard and the first thing that comes up is the GPDR confirmation
<img width="512" alt="Screenshot 1" src="https://github.com/freesewing/freesewing/assets/49982580/3ad4813b-059b-4d5d-b347-37785328e81f">
Which she read and thought "of course they need to process my measurements" so she clicked "confirm" and then two checkmarks appeared
<img width="556" alt="Screenshot 2" src="https://github.com/freesewing/freesewing/assets/49982580/9bfecd63-1c48-49b0-affd-21d8ceb90d32">
while the text did not change, which confused me but didn't confuse her because she didn't read the labels because the text didn't change ¯\\\_(ツ)\_/¯

I then left her to join the Discord and ask about the difference between `en.freesewing.org` and `freesewing.org`. 

In the meanwhile, she got this screen
<img width="449" alt="Screenshot 3" src="https://github.com/freesewing/freesewing/assets/49982580/57e72e31-f68f-4937-a4e8-98505126e195">
[...]

PR #6677, opened 2024-05-01
Some improvements in German translation

alex-fu27:
Following #6324, I now finally fond time to look into some of the translation, especially the account pages. I sadly didn't see the Crowdin installation which I don't think was there when I created the ticket. I hope it is possible to merge these.

As there is now a translation team for German, I would however first like to check back with them: There are some terms which are difficult to translate, like the 'measurement sets', so I create this as a draft.


joostdecock:
Thank you for sumitting this PR @alex-fu27 
I could merge it, but doing so would be rather pointless as the changes would be overwritten again by our translation software which is the source of truth for translations.

Please check the docs for translators at: https://freesewing.dev/guides/translation

I appreciate this can be frustrating, but I hope you'll be able to make these improvements in Crowdin so that they actually show up on the site :pray:

PR #6738, opened 2024-05-12
fix(jane): Fix various issues with Jane

joostdecock:
This includes wrong names for MDX components, missing content for documentation of design options, and renaming a poorly named option.


joostdecock:
@SeaZeeZee note that I've renamed your `sleeveBonus` option to `sleeveLength` because the name was misleading.

PR #6747, opened 2024-05-13
[fix] Jane documentation fixes

### LINKED ISSUE #6739: [docs]: Add missing documentation for Jane's design options

http://localhost:8000/docs/designs/jane

### Suggested content

Please add docs for the options in Jane.

### Additional context

Note that I've already created all the folders and files in #6738 you just have to write the descriptions.

SeaZeeZee:
Closes #6739. 

- added options design documentation 
- updated the dates in titles
- changed sleeveLength to sleevelengthbonus because the design has a very specific sleeve length that is specific to the time period and so this option is adding additional length. 
- reformatted the instructions to be more readable 
- added designer note

the designer notes page is giving a 404 and Im not sure why


designs/jane/src/sleeve.mjs
```diff
@@ -3,7 +3,7 @@ export const sleeve = {
   measurements: ['biceps', 'shoulderToElbow'],
   options: {
     bicepsEase: { pct: 18, min: 18, max: 51, menu: 'fit' },
-    sleeveLength: { pct: 80, min: 60, max: 100, menu: 'style' },
+    sleeveLengthBonus: { pct: 80, min: 60, max: 100, menu: 'style' },
```
joostdecock, line 6: It is not a `bonus` when the minimum is 60%. Which is why I changed the option because the name is misleading.

Not sure if you saw [my comment](https://github.com/freesewing/freesewing/pull/6738#issuecomment-2106192386), but this is not an accident.

joostdecock (changes requested):

joostdecock (approved):

PR #6751, opened 2024-05-13
Adds Huey, Hugo, and Lunetius line drawings

nataliasayang:
Adds Huey and Hugo line drawings as React components. As always, if others can improve on these, I hope they will submit their versions to FreeSewing.

image

image


nataliasayang:
Sorry, one more and I'll stop adding to this. Lunetius:

image

This one would benefit from a back view but I don't have one ready right now.

joostdecock (approved):

PR #6763, opened 2024-05-15
Replace input type="number" with input type="text" inputmode="numeric"/"decimal"

### LINKED ISSUE #6542: [feature]: Change input type for number fields

Right now, number fields in FreeSewing use `<input type="number">`. This provides the advantage of a number keyboard when using a software keyboard and (mostly) limiting keys entered to numerical values. However, it has a number of disadvantages, most noticeably, when the user scrolls while the field is in focus, the number value changes in addition to the page scrolling. This can result in unintended changes to the number fields. I've noticed this most on the measurements entry page, but it may be elsewhere as well. There are accessibility issues with this field type as well. Changing these fields to `<input type="text" inputmode="numeric" pattern="[0-9]*">` will address these issues, among others.

See more info on the issue and how the fix was implemented on gov.uk [Why the GOV.UK Design System team changed the input type for numbers](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).

### Additional context

_No response_

HaasJona:
Fixes #6542

Also fixes some other small issues:

* duplicate `item_number` input element in patron Paypal code
* replaced invalid `inputMode="number"` with `inputMode="decimal"`
* replaced `disabled={Number(amount) < 1}` with `disabled={!(Number(amount) > 0)}` (an invalid value would be `NaN`, which wouldn't be less than 1.)

### DIFF
--- a/sites/org/components/github/inputs.mjs
+++ b/sites/org/components/github/inputs.mjs
@@ -16,7 +16,9 @@ export const AuthorInput = ({ author, setAuthor }) => (
     className={`input input-text input-bordered input-lg w-full mb-2 ${
       author ? 'input-success' : 'input-error'
     }`}
-    type="number"
+    type="text"
+    inputMode="numeric"
+    pattern="[0-9]*"
     value={author}
     placeholder="Enter the user ID here"
     onChange={(evt) => setAuthor(evt.target.value)}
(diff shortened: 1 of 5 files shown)


joostdecock (approved):
Nice work 👍

PR #6764, opened 2024-05-15
Set maintainer of my design documentations to me

HaasJona:


joostdecock (approved):
FYI: You only have to/had to do this in the `en.md` files.

Sorry, I probably should have made that more clear.

HaasJona:
> FYI: You only have to/had to do this in the en.md files.

I just used search&replace, filtering that to only the en.md files would have been actually slightly more work.

PR #6776, opened 2024-05-18
 [feature]: Bibi: Add waistband, cuff and neck finish option, improve side seam calculation

### LINKED ISSUE #6775: [feature]: Bibi: Add waistband, cuff and neck finish option

Bibi can be improved a bit as it could give you some more guidance for finishing the waistband, armholes and neck:

- Add a waistband ribbing option that works similar to Sven and automatically reduces the height of the front/back part by the right amount
- Calculate neck hole size and remove seam allowance on neck by default as it isn't needed for knit band or knit binding finish.
- Add option for finishing armholes:
  - simple folded over hem (only available on separate sleeve part)
  - ribbing like Sven (primarily for long arm shirts, only available on separate sleeve part)
  - knit binding/knit band without seam allowance (primarily for armless and cap sleeve shirts, having seam allowance here would be bad as the sleeve isn't tube shaped and you can't fold over the seam allowance properly)
 
I'll also improve the documentation in that area.

### Additional context

_No response_

HaasJona:
Fixes #6775

Before:
Bildschirmfoto vom 2024-05-27 09-32-28

After:
Bildschirmfoto vom 2024-05-27 09-34-50

The side seam now follows the measurements (plus ease) exactly and makes sure there is enough room around the chest, hips and waistband.

### DIFF
--- a/designs/bibi/src/front.mjs
+++ b/designs/bibi/src/front.mjs
@@ -59,12 +59,11 @@ function bibiFront({
   Snippet,
   snippets,
   options,
+  absoluteOptions,
   measurements,
   macro,
   complete,
-  utils,
   part,
-  log,
 }) {
   // Hide Brian paths
   for (const key of Object.keys(paths)) paths[key].hide()
@@ -84,7 +83,7 @@ function bibiFront({
   delete points.cbHem
   delete points.cbNeck
 
-  constructFrontPoints(points, Point, measurements, options)
+  constructFrontPoints(part)
 
   // FBA
   points.originalChest = points.chest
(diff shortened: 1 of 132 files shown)


designs/bibi/src/front.mjs
```diff
@@ -137,15 +137,15 @@ function bibiFront({
...
+  constructFrontHem(part, options.useWaistRibbing ? -store.get('ribbingHeight') : 0)
 
   store.set('frontSideSeamLength', paths.sideSeam.length())
   const frontLength = store.get('frontSideSeamLength')
   const backLength = store.get('backSideSeamLength') ?? 0
   const dartLength = frontLength - backLength
   const constructDart = (path, tip, dartLength) => {
     const length = path.length()
-
+    console.log('dart', length, path)
```
joostdecock, line 148: Please remove this debug output

joostdecock (changes requested):
Just drop that `console.log()` and we're good to go :)

designs/bibi/src/front.mjs
HaasJona: Done.

joostdecock (approved):

PR #6815, opened 2024-05-28
add(shared): Add line drawings for Teagan and Bibi

HaasJona:
Bildschirmfoto vom 2024-05-28 11-01-28
Bildschirmfoto vom 2024-05-28 11-01-18


HaasJona:
Bibi isn't really a classical block. It's more like a simple extensible top with a lot of configuration options. It's only a block in the sense that it doesn't represent a specific garment or style. Maybe we should call it something else, but I'm not sure.

 I'll add instructions for hemming and so on here: #6775

joostdecock (approved):

PR #6818, opened 2024-06-02
Load markdown text and images from the server rather than from GitHub when markdown is dynamically loaded

micheal65536:
Currently the website loads markdown from GitHub for client-side rendering. I am not sure what the reason for this is, however it creates confusion and barriers to being able to fully test local changes to the documentation pages and to being able to run a full local copy of the site (e.g. if you add documentation for a new sewing pattern that is not in upstream, this will not display in the site, or if you modify existing documentation your changes won't actually display). (see Discord: https://discord.com/channels/698854858052075530/698862765053575188/1246570558024323143)

I have changed it to load from the actual server that is being used. I am not sure if there is a reason why it is not done this way. I am not sure if there is a problem with the way that I have done it, or if this approach was already ruled out for some other reason. If so please can you explain, or let me know if you need me to structure the code differently or whatever (I'm not familiar with React). Sorry if this is unwelcome. However, it was very frustrating to me that I am unable to simply add my own design and have it properly appear in my own local copy of the website.

### DIFF
--- a/sites/shared/prebuild/markdown.mjs
+++ b/sites/shared/prebuild/markdown.mjs
@@ -309,6 +309,28 @@ const writeFile = async (filename, exportname, site, content) => {
 }
 
 /*
+ * Copy all the markdown resources + image assets to the public folder
+ */
+const copyMdRawFiles = async (site) => {
+  const src = await path.resolve(process.cwd(), '..', '..', 'markdown', site)
+  const dst = await path.resolve(process.cwd(), '..', site, 'public', 'markdown')
+
+  await new Promise((resolve, reject) => {
+    exec(`mkdir -p ${dst}`, { maxBuffer: 2048 * 1024 }, (error, stdout, stderr) => {
+      if (error) reject(error)
+      resolve(stdout)
+    })
+  })
+
+  await new Promise((resolve, reject) => {
+    exec(`cp -r ${src} ${dst}`, { maxBuffer: 2048 * 1024 }, (error, stdout, stderr) => {
+      if (error) reject(error)
+      resolve(stdout)
+    })
+  })
+}
+
+/*
  * Main method that does what needs doing for the docs
  */
 export const prebuildDocs = async (store) => {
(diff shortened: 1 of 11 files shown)


joostdecock:
Thanks for creating this PR @micheal65536 
I appreciate your feedback, and understand that you find the currently implementation frustrating. 
I don't disagree, and your fix is a great improvement.

The reason we currently load the content from GitHub is because our stand-alone development environment (SDE) is... stand alone. In other words, it exists/gets deployed outside of our monorepo, and thus there is no option to load the content from the local disk.

Loading MDX dynamically comes with its own set of tradeoffs, and at one point we went down this route because it gave us a solution that worked both for the site, and for the SDE.

However, that solution was chosen under time pressure to release v3. So not because it was great, but because it was a quick fix.

I am happy to go with this approach, as I've already started re-working the development environment, so I don't think this should be blocked. Especially since doing development in our monorepo is gaining popularity over using the SDE.

My only remark is that we do not need to copy all markdown, as only docs are loaded dynamically, so we should leave blog, showcase, and newsletter posts out of it. But I'll put that in the review.

sites/shared/prebuild/markdown.mjs
```diff
@@ -308,6 +308,28 @@ const writeFile = async (filename, exportname, site, content) => {
   )
 }
 
+/*
+ * Copy all the markdown resources + image assets to the public folder
+ */
+const copyMdRawFiles = async (site) => {
```
joostdecock, line 314: We do not need to copy **all** markdown.
The only markdown that gets loaded dynamically are `docs`
So we can skip `blog`, `showcase`, and `newsletter` content.

As such, I'd like to update this to only copy the `docs` subfolder:
```mjs
  const src = await path.resolve(process.cwd(), '..', '..', 'markdown', site. 'org', 'docs')
  const dst = await path.resolve(process.cwd(), '..', site, 'public', 'markdown', 'org', 'docs')
```

joostdecock (changes requested):
LGTM, just make sure to only copy the docs content

sites/shared/prebuild/markdown.mjs
micheal65536: I made it copy everything to avoid the risk of someone changing something else in the future and running into the same problem again (e.g. if blog posts were changed to load dynamically), and because I wasn't sure exactly what parts were loaded from where (especially images which to me it looked like/seemed reasonable that they were always loaded remotely). (I'm not familiar with React so I have a hard time understanding what the code is actually doing.)

I don't agree with only copying docs, this sounds like a future issue waiting to happen, but if you want to change it then go ahead, I'm not sure why you need to ask me about it before merging.

PR #6837, opened 2024-06-12
fix(core): Handle path splits on start or end points

### LINKED ISSUE #6816: [bug]: Path.split can sometimes return empty arrays in the result array

Path. split is designed to return two paths `[firstHalf, secondHalf]` but if one of the paths would be zero length or super short, an empty array is returned in place of that path.

E.g. calling `myPath.split(pointNearTheStartOfMyPath)` can return `[[], approximatelyMyPath]`.

The empty array is unexpected. A zero length path or maybe `null` would probably be a better return value, as that would be easier to handle than having to do something like `Array.isArray(splitResult[0])`.

_No response_

### Additional context

_No response_

joostdecock:
In some edge cases, like when splitting a path on or very close to its
start or end point, `Path.split()` will return an array in which one of
the elements is not a Path object, but rather an empty array.

That is rather combersome to check for, so this changes that behaviour
and will return null for such cases.

I've also updated the documentation to clarify this behaviour.

Fixes #6816

This also fixes a bug when a path was being split on its end point.

### DIFF
--- a/packages/core/src/path.mjs
+++ b/packages/core/src/path.mjs
@@ -898,7 +898,12 @@ Path.prototype.split = function (point) {
       break
     }
     if (path.ops[1].type === 'line') {
-      if (pointOnLine(path.ops[0].to, path.ops[1].to, point)) {
+      if (path.ops[1].to.sitsRoughlyOn(point)) {
+        pi++
+        firstHalf = divided.slice(0, pi)
+        secondHalf = divided.slice(pi)
+        break
+      } else if (pointOnLine(path.ops[0].to, path.ops[1].to, point)) {
         firstHalf = divided.slice(0, pi)
         firstHalf.push(new Path().__withLog(this.log).move(path.ops[0].to).line(point))
         pi++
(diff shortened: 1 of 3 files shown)


packages/core/src/path.mjs
HaasJona, line 1: What do you think about changing these defaults to `null`

```
  let firstHalf = []
  let secondHalf = []
```

otherwise I fear that the empty array can sneak through in some cases anyway.

joostdecock: We use a lot of array methods like `.slice()`, `.push()`, or `.unshift()` in this method.
Initializing this as `null` would require us to check at each instance whether it's an array first, so that would complicate matters.

What makes you think an empty array can sneak through? 
Where previously it could skip this check and return the initialized value, it now always returns either the result of `__joinPaths()` or `null`. The initial value is never returned,

What used to be:
```mjs
if (firstHalf.length > 0) firstHalf = __joinPaths(firstHalf)
```

is now:
```mjs
firstHalf = firstHalf.length > 0 && firstHalf[0].ops.length > 1 ? __joinPaths(firstHalf) : null
```

PR #6838, opened 2024-06-12
fix: Bruce is a bottom, not a top

### LINKED ISSUE #6803: [bug]: Tagging issue on Bruce
The Bruce boxer briefs come up when I filter using the tag "Tops". It also does not come up under the tag "Bottoms". I think someone hit the wrong option when tagging the design. 

[User 24039](https://next.freesewing.org/users/24039)

joostdecock:
This fixes Bruce's tagging to show that it's a bottom, not a top.

Fixes #6803

### DIFF
--- a/config/software/designs.json
+++ b/config/software/designs.json
@@ -179,7 +179,7 @@
     "lab": true,
     "org": true,
     "tags": [
-      "tops",
+      "bottoms",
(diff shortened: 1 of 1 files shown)


joostdecock:
Dupe of #6804 Closing

PR #6928, opened 2024-07-02
fix(core): Fix typo in path.clean() method

HaasJona:
Path.clean() does weird things on paths with curves. Found while messing with Bibi design.


joostdecock (approved):
Nice catch!

Issue #6931, opened 2024-07-03
[docs]: `waralee` straps documentation inconsistent and unclear


woutervdub:
### Where can we see the problem? 🤔

https://freesewing.org/docs/designs/waralee/cutting
https://freesewing.org/docs/designs/waralee/instructions

### Additional context

@Wouter.vdub Waralee is a little bit confusing, because the cutting instructions don't mention the strips and the sewing instructions tell you to measure and construct them manually. (It explicitly says "These strings are not included in the pattern." at the top). But then the pattern actually contains the strips.

VSinerva:
Closing as fixed by #6932

PR #6932, opened 2024-07-03
Fix clarity issues with the straps #6931

woutervdub:


markdown/org/docs/designs/waralee/instructions/en.md
```diff
@@ -102,28 +102,18 @@ Now you get to sew the two pants parts together.
...
+If you printed them, just cut them out. If not, the *Messages from the designer* will have a note about their sizes.
 
 <Note>
 
-##### You can lengthen the front ties
-
-Optionally, to make the front ties longer, extend them by your quarter waist measurement.
-
-When generating your pattern, one of the style options is called “Waist Band”. The strings, when folded, should have the same width as you specified for that option, 2.5cm (1") by default.
+##### Tradisionally, the straps are the same size
```
HaasJona, line 113: Typo: Tradisionally

markdown/org/docs/designs/waralee/options/knotinfront/en.md
```diff
@@ -2,5 +2,5 @@
 title: "Knot in Front"
 ---
 
-Normally the Waralee is closed with the tie in the front. This allows you to change
-that to a bow tie in the back.
\ No newline at end of file
+Normally the Waralee is closed with all straps tied in the front. This allows you to change
+that to a set tied in the front and in the back, the traditional way.
```
HaasJona, line 6: If the knot in front option is set to "Knot in Back", the back ties are made longer instead, so both straps will be tied in the back. That contradicts this documentation. With knot in the back selected, both ties probably should be the same length or the option should be split into 3 (both ties at the front, both ties at the back, and the traditional way with shorter same-length straps).

HaasJona (changes requested):

HaasJona (approved):
The markdown documentation for "knotInFront" is now unused and probably can be deleted (unless someone still wants to keep it for compatibility or whatever?) but other than that, looks good to me.

designs/waralee/src/waistband.mjs
```diff
@@ -48,13 +48,13 @@ function waraleeWaistband(
     case 'strapFront':
       partNr = 7
       waistBandLength -= waistBandLengthFront * 2
-      waistBandLength += options.knotInFront ? measurements.waist / 2 : 0
+      waistBandLength += options.knotPlacement == 'front' ? measurements.waist / 2 : 0
```
joostdecock, line 51: `===` instead of `==` please 😊

It's a good habit to use strict comparison to avoid subtle type bugs, even if in this case there's no real risk.

designs/waralee/src/waistband.mjs
```diff
@@ -48,13 +48,13 @@ function waraleeWaistband(
...
       break
 
     case 'strapBack':
       partNr = 8
       waistBandLength -= waistBandLengthBack * 2
-      waistBandLength += options.knotInFront ? 0 : measurements.waist / 2
+      waistBandLength += options.knotPlacement == 'back' ? measurements.waist / 2 : 0
```
joostdecock, line 57: `===` instead of `==` please 😊

It's a good habit to use strict comparison to avoid subtle type bugs, even if in this case there's no real risk.

joostdecock (changes requested):

joostdecock (approved):

PR #7030, opened 2024-08-25
Add/measurement validation

### LINKED ISSUE #6576: [feature]: Warn on impossible physical measurements

The measurement set dialog should warn when the user enter values that are physically impossible and will likely cause issues when drafting a pattern. The scope of this feature request is limited to warning about measurements that are (almost¹) impossible. This is not supposed to warn about measurements that are simply uncommon, scaled up/down for dolls, statues, and so on, or which are simply different from some assumed standard body type.

¹) People with amputations or limb transplants or measurement sets that are used for non-human-shaped dolls or animals might get a warning, but there's not much we can do about it as you might need to fake some measurements or manually adjust the patterns anyway...

I propose the following checks:
### Upper body front vertical measurements
[HPS to waist front](https://freesewing.org/docs/measurements/hpstowaistfront) > [HPS to bust](https://freesewing.org/docs/measurements/hpstobust)
[HPS to waist front](https://freesewing.org/docs/measurements/hpstowaistfront) > [Waist to underbust](https://freesewing.org/docs/measurements/waisttounderbust)
### Arm measurements
[Shoulder to wrist](https://freesewing.org/docs/measurements/shouldertowrist) > [Shoulder to elbow](https://freesewing.org/docs/measurements/shouldertoelbow)
### Lower body vertical measurements
[Waist to floor](https://freesewing.org/docs/measurements/waisttofloor) > [Waist to knee](https://freesewing.org/docs/measurements/waisttoknee) > [Waist to upper leg](https://freesewing.org/docs/measurements/waisttoupperleg) > [Waist to seat](https://freesewing.org/docs/measurements/waisttoseat) > [Waist to hips](https://freesewing.org/docs/measurements/waisttohips)
[Waist to floor](https://freesewing.org/docs/measurements/waisttofloor) > [Inseam](https://freesewing.org/docs/measurements/inseam)
### Measurements that are strictly defined as part of another measurement
[Cross seam](https://freesewing.org/docs/measurements/crossseam) > [Cross seam front](https://freesewing.org/docs/measurements/crossseamfront)
[Seat circumference](https://freesewing.org/docs/measurements/seat) > [Seat back](https://freesewing.org/docs/measurements/seatback)
[High bust](https://freesewing.org/docs/measurements/highbust) > [High bust front](https://freesewing.org/docs/measurements/highbustfront)
[Chest circumference](https://freesewing.org/docs/measurements/chest) > [Bust front](https://freesewing.org/docs/measurements/bustfront) > [Bust span](https://freesewing.org/docs/measurements/bustspan)

(If I forgot something or you think something is incorrect, please leave a comment)

Also none of the length measurements should be zero or negative. Like even if the waist sits (almost) on the underbust or on the hips, you probably still want to enter a small positive measurement to prevent errors in pattern construction.

### Additional context

_No response_

HaasJona:
WIP.

Example messages:

```
Chest circumference should be larger than Bust front.

Back Chest Measurement: High bust - High bust front (= 61.4cm) should be approximately equal to Chest circumference - Bust front (= 64.6cm). Both calculations should measure the back half of your ribcage and should thus be roughly equal.

Front Bust measurements: HPS to waist front (= 47.6cm) should be smaller than HPS to bust + Bust point to underbust + Waist to underbust (= 53.8cm). Measuring from the HPS point to the waist across your front should be equal or smaller (due to e.g. skin folds) than the sum of the individual vertical bust measurements.

Vertical leg measurements: Waist to floor (= 117.4cm) should be smaller than Waist to upper leg + Inseam (= 116cm). The waist-to-upper-leg measurement goes from the waist along the body to a bit below the crotch and the inseam measurement goes vertically from the crotch to the floor. Thus, going directly vertically from the waist to the floor should be a bit smaller than the sum.
```

Fixes #6576

### DIFF
--- a/sites/shared/components/account/en.yaml
+++ b/sites/shared/components/account/en.yaml
@@ -363,6 +363,6 @@ validateMeasurements: Check Measurements
 backChest.t: Back Chest Measurement
 backChest.d: Both calculations should measure the back half of your ribcage and should thus be roughly equal.
 bustFront.t: Front Bust measurements
-bustFront.d: Measuring from the HPS point to the waist across your front should be equal or smaller than the sum of the individual vertical bust measurements.
+bustFront.d: Measuring from the HPS point directly to the waist across your front can't be larger than the sum of the individual vertical bust measurements. However note that waist to underbust is measured at the side of the body instead of at the front, but usually this doesn't make a large difference.
 waistToUpperLeg.t: Vertical leg measurements
-waistToUpperLeg.d: The waist-to-upper-leg measurement goes from the waist along the body to a bit below the crotch and the inseam measurement goes vertically from the crotch to the floor. Thus, going directly vertically from the waist to the floor should be a bit smaller than the sum.
+waistToUpperLeg.d: The waist-to-upper-leg measurement goes from the waist along the body to a bit below the crotch and the inseam measurement goes vertically from the crotch to the floor. Thus, going directly vertically from the waist to the floor should be a bit smaller than the sum of the other measurements.
--- a/sites/shared/components/account/en.yaml
+++ b/sites/shared/components/account/en.yaml
@@ -353,3 +353,16 @@ importSetTip2: Your file can either contain a single measurements set, or an arr
 importing: Importing
 imported: Imported
 importFailed: Import failed
+
+# validation
+shouldBeLarger: "{ lhs } should be larger than { rhs }."
+shouldBeEqual: "{ lhsConstraint } (= { lhsSum }) should be approximately equal to { rhsConstraint } (= { rhsSum })."
+shouldBeSlightlySmaller: "{ lhsConstraint } (= { lhsSum }) should be approximately equal to or smaller than { rhsConstraint } (= { rhsSum })."
+validationSuccess: No problems were found.
+validateMeasurements: Check Measurements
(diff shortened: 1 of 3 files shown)


BenJamesBen:
I note that the Waist to Underbust measurement is taken at the side of the body, whereas HPS to Waist Front, HPS to Bust, and Bust Point to Underbust are taken at the front of the body. I'm not sure if it makes a difference.

In case anyone wanted to know what the validation looks like, In the Account -> Your Measurements Sets -> [Measurements Set] page there is a "Check Measurements" button. Pressing it will cause a popup to appear with either the issues seen or "no problems found".
Screenshot 2024-08-25 at 8 00 46 PM
Screenshot 2024-08-25 at 8 00 28 PM
Screenshot 2024-08-25 at 8 03 42 PM

HaasJona:
That pop-up is mainly because I'm not sure how to create a proper react component for it or what the final design should look like. 

Good point about the measurement. I don't see how that would make a large difference. Both underbust and waist measurements should be parallel to the ground and the body should be at least approximately vertical between these measurements. Can be curved obviously, but nothing extreme like going around a bust or belly. The tolerance can be adjusted obviously.

HaasJona:
Closed in favour of #7379

PR #7035, opened 2024-08-26
fix(docs): Update backend Access Control to v3

BenJamesBen:
Based on the information in [config/roles.mjs](https://github.com/freesewing/freesewing/blob/16e6b75056305052f74d4f3ee12305c38e4adf3b/config/roles.mjs)

Screenshot 2024-08-26 at 10 23 10 AM


BenJamesBen:
The backend documentation was removed during the port to Docusaurus, so this PR is no longer applicable. Closing.

PR #7038, opened 2024-08-27
fix(docs): Add return value for Path.split() with point not on path

BenJamesBen:


HaasJona (approved):

BenJamesBen:
Rebased and cleared merge conflict resulting from the port to Docusaurus. This PR is now good to go again.

PR #7047, opened 2024-08-28
fix(lab): Add curated measurements lineup backdrop image

BenJamesBen:
Before:
Screenshot 2024-08-28 at 3 33 07 PM

After:
Screenshot 2024-08-28 at 3 32 49 PM


BenJamesBen:
The lab site was removed, so this PR is no longer applicable or needed. Closing.

PR #7050, opened 2024-08-30
fix(docs): Add expand setting

### LINKED ISSUE #5923: [docs]: Add expand documentation

_No response_

Yes, I am a tier-2 patron ❤️

### Suggested content

There is a newly-added `expand` setting that needs to be documented. This setting is also exposed in **Part** shorthand and needs to be added to the docs there. Finally, we should write a _Howto_ to provide practical instructions to pattern designers and coders how to use the setting in their source code to conditionally hide parts or present alternate parts.

https://freesewing.dev/reference/api/part/draft
https://freesewing.dev/reference/settings
https://freesewing.dev/howtos/code

### Additional context

_No response_

BenJamesBen:
Fixes #5923

Screenshot 2024-08-30 at 1 22 36 AM

### DIFF
--- a/markdown/dev/reference/api/part/draft/en.md
+++ b/markdown/dev/reference/api/part/draft/en.md
@@ -26,6 +26,7 @@ access the following properties:
 || **_Access to settings_** |
 | `absoluteOptions` | Access to `settings.absoluteOptions` |
 | `complete`        | Access to `settings.complete` |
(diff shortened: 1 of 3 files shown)


BenJamesBen:
I updated the PR to remove conflicts caused by the port to Docusaurus. Things should be ready to go again.

PR #7071, opened 2024-09-10
Prototype for cornered path joins

HaasJona:
As discussed on Discord, this changes the offset, join and close methods so the paths are joined at a continuous angle without cutting off the corners.

Some sample screenshots:
Yuri:
Bildschirmfoto vom 2024-09-10 19-12-55
Hortensia:
Bildschirmfoto vom 2024-09-10 18-55-31
Simon:
Bildschirmfoto vom 2024-09-10 18-54-47
(Some corners are still cut, but this is explicitly coded like that in the pattern)
Teagan:
image
(not sure what's going on in the one corner of the sleeve, probably the design doing something weird, have not investigated yet)

This should not be merged as is and is more meant for testing. Probably would need a way to gradually test and migrate designs (if desired by the creator?) and maybe add an option for the behavior as some people might like the cut corners. This is open for discussion.

### DIFF
--- a/packages/core/src/path.mjs
+++ b/packages/core/src/path.mjs
@@ -12,7 +12,11 @@ import {
   round,
   __addNonEnumProp,
   __asNumber,
+  beamsIntersect,
 } from './utils.mjs'
+import { addLengthGuard } from 'chai/lib/chai/utils/index.js'
+
+const newJoinEnabled = true
 
 //////////////////////////////////////////////
 //               CONSTRUCTOR                //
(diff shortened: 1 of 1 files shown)


BenJamesBen:
When testing the PR in the lab, I'm getting an error when enabling seam allowance on Aaron.
Screenshot 2024-09-10 at 1 15 29 PM

Steps to reproduce:
1. In the localhost lab, create a new Aaron using the Elanor curated measurements set.
2. Enable seam allowance.

BenJamesBen:
A few more unusual corners:
Screenshot 2024-09-10 at 1 25 30 PM
Screenshot 2024-09-10 at 1 31 13 PM
Screenshot 2024-09-10 at 1 33 28 PM
Screenshot 2024-09-10 at 1 35 02 PM

BenJamesBen:
I think that I need some documentation for the new join method, so I know how it is supposed to behave, to better test it.

Currently I am seeing variable behavior, "sometimes join the end of path1 to the start of path2, and other times extend the end of path1 and the start of path2 so they intersect and join the extended lines together". It isn't clear to me under what circumstances each behavior is supposed to occur.

I note that sometimes `path1.join(path2)` produces a different line than `path2.reverse().join(path1.reverse()).reverse()` which seems to be incorrect since I expect the behavior to be consistent. But, without knowing how the new join method is supposed to work, I am not sure-- perhaps that is the expected and correct behavior?
Screenshot 2024-09-10 at 3 37 27 PM

If the new join behavior is implemented as a new `Path` method that depends on changes to path closing behavior, then perhaps the new behavior should be implemented as a new method as well, rather than changing the existing `Path.close()`. Otherwise, I think I would also need some documentation for the new `Path.close()` behavior, to be able to test it.

HaasJona:
I've come to the conclusion that this makes more sense as a kind of offset or seam allowance macro instead of changing/extending the core functions. I'll leave this branch in place for reference, but this approach is convoluted and has too many issues to be used as it is.

PR #7090, opened 2024-09-16
fix(shared) Line drawing border overlap

### LINKED ISSUE #6790: [bug]: Line drawings are sometimes displayed out of their box

Depending on the screen resolution, the line drawing on the design chooser site sometimes escape their box:

grafik

This seems to happen mostly with very wide screens. I'm using 2560x1440.

sites/org

### Additional context

_No response_

bobgeorgethe3rd:
Closes #6790
I had a play around and I believe this solves this issue or at least creates a temporary patch.

### DIFF
--- a/sites/shared/components/designs/design.mjs
+++ b/sites/shared/components/designs/design.mjs
@@ -93,20 +93,20 @@ export const DesignCard = ({ name, lineDrawing = false }) => {
       }
     >
       <div
-        className={`flex flex-col flex-nowrap items-start justify-start gap-2 h-80 w-full
+        className={`flex flex-col flex-nowrap items-start justify-start gap-0 h-80 w-full
         btn btn-ghost border border-neutral p-0 border-b-none
         hover:border hover:border-secondary
         relative`}
         style={bg}
       >
         <h5
-          className={`text-center py-2 px-4 rounded-t-lg m-0 w-full
+          className={`text-center py-1 px-4 rounded-t-lg m-0 w-full
         ${lineDrawing ? '' : 'bg-neutral text-neutral-content bg-opacity-70'}`}
         >
           {t(`designs:${name}.t`)}
         </h5>
         <div className={lineDrawing ? 'p-4 grow w-full' : 'py-8'}>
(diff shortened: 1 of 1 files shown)


sites/shared/components/designs/design.mjs
```diff
@@ -93,20 +93,20 @@ export const DesignCard = ({ name, lineDrawing = false }) => {
...
         <h5
-          className={`text-center py-2 px-4 rounded-t-lg m-0 w-full
+          className={`text-center py-1 px-4 rounded-t-lg m-0 w-full
         ${lineDrawing ? '' : 'bg-neutral text-neutral-content bg-opacity-70'}`}
         >
           {t(`designs:${name}.t`)}
         </h5>
         <div className={lineDrawing ? 'p-4 grow w-full' : 'py-8'}>
-          <LineDrawing className="h-64 max-w-full m-auto my-4 text-base-content" />
+          <LineDrawing className="h-64 max-w-full m-auto max-w-full text-base-content" />
```
HaasJona, line 109: There's `max-w-full` twice in that line. Not sure if that makes sense.

joostdecock: It does not, but also does not cause harm. If we ignore that noop, this boils down to:
- Reducing the gap on the flex column items
- Reducing the vertical padding on the heading
- Removing the vertical margin on the linedrawing

In other words, this removes gap, margin, and padding, so it's strictly removes whitespace. So this is likely a fix for the browser + resolution Bob was using, but it's not a _real_ fix.

As I [explained](https://github.com/freesewing/freesewing/issues/6790#issuecomment-2162809519) the issue is that the bounding box of the SVG is not properly calculated because it lacks a height value.

This requires a proper fix that addresses the matter for all scenarios. So I'm going to close this as only trades one set of problems for another.

Or to put it differently: Fixing this will probably require a more fundamental fix.

joostdecock:
This attempts to avoid the problem by removing whitespace, but does not fix the underlying issue.

I'm going to close this as we need a proper fix, but I appreciate the effort :)

PR #7093, opened 2024-09-16
add(lab) Missing Jane & Opal pages

bobgeorgethe3rd:
These pages are missing and keep getting auto-generated when using `yarn lab`. When added they won't be generated removing confusion for users who  are working in the lab.


joostdecock:
Lab is deprecated and will be removed

PR #7098, opened 2024-09-17
fix(bella): Front bust dart changes: add angle and minimum fabric options, obey side seam length, allow convex curve

### LINKED ISSUE #7082: [bug]: (Bella) Front side seam is longer than back side seam

https://discord.com/channels/698854858052075530/1271241731194486784/1271241731194486784
discord://discord.com/channels/698854858052075530/1271241731194486784/1271241731194486784
From Discord:
> First timer here trying to create the noble block. The sides that will be sewn together are two very different lengths and i can’t figure out why. I’ve already cut it out and stitched it, and it does seem to fit me relatively pretty well if i match up the two bottom seams. 
IMG_6805

> I experimented with Noble a bit, and I note that I can cause a similar side seam length mismatch by increasing hpsToBust and decreasing hpsToWaistBack so the values get closer to each other. For example:
```yaml
measurements:
  highBust: 820
  chest: 830
  underbust: 750
  waist: 730
  waistBack: 370
  bustSpan: 140
  neck: 360
  hpsToBust: 300
  hpsToWaistFront: 360
  hpsToWaistBack: 330
  shoulderToShoulder: 360
  shoulderSlope: 20
```

> I believe that Noble is inheriting the issue from Bella.
Screenshot_2024-08-09_at_1 13 26_PM

> Bella generates the back part first. No problems there. Then, it calculates the front `points.bustDartTop` beginning location of the dart based on how far down points.bust is located. But, it does so without taking into consideration the already-calculated back side seam length. So, the length of front side seam from armhole to dart can start out longer than the back side seam.

> Additionally, if the length of the front side seam from armhole to top of the dart is longer than the back side seam, Bella incorrectly calculates a negative value length for the bottom of the dart to the waist. It then uses the negative length to incorrectly calculate where the side waist `points.sideHem` is located, and this adds additional length to the back side seam below the dart.

I was able to fix the issue by adjusting the location of the dart top to match the side seam length already established by the back part. I also was able to add code to adjust the location of the dart top to make sure there is some side seam length below the dart, to eliminate scenarios where there is no front bust dart.

However, it was pointed out that Bella might be supposed to work only with measurements where the bust point is above the waist:
> Regarding the bust point vs the natural waist: I think we've said before that the bust point is supposed to always be above the waist.
>
> (If it's not, several other measurements also take on a different meaning, and the pattern is unlikely to work.)

If that is the case, then the proposed fix would be invalid and the correct remedy might be to not draft the pattern and instead tell the customer why it isn't possible.

designs/bella

### Additional context

_No response_

BenJamesBen:
Fixes #7082

Previously Bella drew a front bust dart at a hardcoded angle, regardless of whether doing so would create a front side seam longer than the back side seam.

This changes Bella's front bust dart:
1. Adds a `bustDartAngle` option to allow variable angle.
2. Adds a `bustDartMinimum Fabric` option to require a minimum amount of fabric above and below the dart. The default is 5% of the side seam which prevents tiny slivers of fabric above/below the dart.
3. Fixes the bug so the front and back side seams are the same length.
4. Also changes the `bustDartCurve` minimum to allow convex curved darts.
5. Adds constant options to Bee and Noble so this PR's changes to Bella do not affect them. Bee and Noble remain the same as before.
6. Adds/changes documentation for the new/changed options.

I tested the fix with Bella, Noble, and Bee.

Screenshot 2024-09-19 at 6 01 12 PM
Screenshot 2024-09-19 at 6 01 36 PM
Screenshot 2024-09-19 at 6 01 56 PM
Screenshot 2024-09-19 at 6 02 18 PM


BenJamesBen:
Changed status to Draft.

I want to rework this PR to incorporate work from the previous, closed, draft PR #4181 which added a bust dart angle option and top and bottom limits to the bust dart position. I think I also want to add an additional new percent option for how the minimum amount of fabric which should be above/below the dart. And, there will need to be documentation added for the new options.

BenJamesBen:
Removed Draft status.

Modified top comment to reflect new PR.

HaasJona:
This seems to be still marked as draft. Otherwise looks good to me.

joostdecock (dismissed):

joostdecock (approved):

PR #7111, opened 2024-09-20
fix(bent): Base sleeve width on biceps circumference

### LINKED ISSUE #6023: [bug]: Jager sleeves still have problems, though ease is fixed.

Jager (and Bent) sleeves generate larger for smaller bicep measures, and smaller for larger bicep measures.
All else equal, at default ease of 15% Jager, with 20cm set for bicep circumference results in this sleeve
image
with the bicep measure set to 30cm this is the sleeve that is generated 
image
and with a bicep of 40 cm this is the sleeve that is generated 
image
Moving the bicep ease slider now correctly makes it larger for more ease, and smaller for less ease however.

designs/jaeger

### Additional context

I checked and Bent has the same issue.

BenJamesBen:
Fixes #6023

This changes the sleeve width algorithm to be based on biceps circumference. Previously the algorithm used a set fraction of the sleeve opening as the sleeve width.

(I am initially filing this as Draft status. I will post to Discord to get feedback on the PR, and within 7 days I will either move this PR out of Draft status or close it entirely.)

Comparing before/after patterns using curated measurements sets, it seems that this new formula produces roughly the same patterns as before. However, as biceps circumference increases, the sleeve width now increases instead of decreasing.

Screenshot 2024-09-20 at 7 15 41 AM
Screenshot 2024-09-20 at 7 16 00 AM

Screenshot 2024-09-20 at 7 18 47 AM
Screenshot 2024-09-20 at 7 18 10 AM

Screenshot 2024-09-20 at 7 15 18 AM
Screenshot 2024-09-20 at 7 15 02 AM

Screenshot 2024-09-20 at 7 19 24 AM
Screenshot 2024-09-20 at 7 20 06 AM

### DIFF
--- a/designs/bent/src/sleeve.mjs
+++ b/designs/bent/src/sleeve.mjs
@@ -5,7 +5,8 @@ function draftBentSleeve({ Path, paths, points, store, options, part }) {
     let { Point, Path, points, store, options, measurements, utils } = part.shorthand()
     // Sleeve frame
     points.top = new Point(0, 0)
-    points.boxTopRight = points.top.shift(0, (store.get('sleevecapTarget') / 5.8) * tweak)
+    const easedQuarterBiceps = (measurements.biceps / 4) * (1 + options.bicepsEase)
+    points.boxTopRight = points.top.shift(0, easedQuarterBiceps * tweak)
     points.boxTopLeft = points.boxTopRight.flipX()
     points.boxBottom = points.top.shift(
       -90,


BenJamesBen:
Increasing the biceps ease option results in a wider sleeve.

Screenshot 2024-09-20 at 9 11 27 AM
Screenshot 2024-09-20 at 9 11 08 AM

HaasJona:
Looks good to me. On my own measurements, the sleeve gets a little bit wider and shorter with your changes, similar to the Elanor measurement set on your screenshot, but that is probably fine. (I haven't made an actual Bent garment to check what might fit better, though.)

But the issue that the sleeve gets smaller on larger biceps measurement is definitely gone.

Lab-Mode before:
Bildschirmfoto vom 2024-09-24 21-26-10

Lab-Mode after:
Bildschirmfoto vom 2024-09-24 21-26-26

Note that with your patch, changing the biceps measurement in the small range that the test mode uses has pretty much no effect. I assume this is because the limiting factor is the armhole size for me after your patch. I confirmed that substantially increasing the measurement definitely increases the sleeve. Lab-mode for "waist-to-armpit" changes the sleeve shape as expected.

joostdecock (approved):

PR #7122, opened 2024-09-25
Add path-util macros

HaasJona:
Bildschirmfoto vom 2024-09-25 14-17-18
Bildschirmfoto vom 2024-09-25 14-17-34
Bildschirmfoto vom 2024-09-25 14-17-48
(same as the offset macro, just applies `sa` as default offset and the `sa` class to the created path)
Bildschirmfoto vom 2024-09-25 14-19-52


HaasJona:
This is now ready for review. Open questions:

* Is it okay that the macros return the paths directly, or should we generate an id for the path and return a list of id's (like the other macros do)? I find returning paths directly easier to use.
* What unit tests do we need?
* Should this be part of the core plugins and automatically available for designs?
* Is the API okay?

Feedback would be welcome.

Issue #7121, opened 2024-09-25
[bug]: Octoplushy design crashes if changing armlength option (and others)


HaasJona:
### What seems to be the problem? 🤔

Steps to reproduce:

- new Octoplushy design
- In the right bar, click on Design Options → Style → Arm length (no dragging required)

Actual result:

- website turns white, error in Browser console:

```
Uncaught TypeError: options is undefined
    toAbs head.mjs:635
    PctOptionInput inputs.mjs:23
    React 16
    onClick accordion.mjs:54
    React 23
    renderReactElement index.tsx:582
    doRender index.tsx:795
    render index.tsx:824
    hydrate index.tsx:993
    pageBootrap page-bootstrap.ts:22
    <anonymous> next-dev.ts:21
    promise callback* next-dev.ts:20
    NextJS 7
head.mjs:635:9
    toAbs head.mjs:635
    PctOptionInput inputs.mjs:23
    React 16
    onClick accordion.mjs:54
    React 19
    forEach self-hosted:4304
    React 4
    renderReactElement index.tsx:582
    doRender index.tsx:795
    render index.tsx:824
    hydrate index.tsx:993
    AsyncFunctionNext self-hosted:807
    (Async: async)
    pageBootrap page-bootstrap.ts:22
    <anonym> next-dev.ts:21
    (Async: promise callback)
    <anonym> next-dev.ts:20
    NextJS 7
```

Expected result:

- no crash

### Design / Plugin / Package 🧐

designs/octoplushy

### Additional context

Filing this under the Octoplushy component, but I suspect the actual issue might be in the FreeSewing core.

BenJamesBen:
There is a workaround:
1. Change the Design Options->Style->Type to Octopus or Squid or change it back to Octoplushy.
2. Go to Design Options->Style->Arm Length and change the setting.
3. The bug does not occur.

bobgeorgethe3rd:
```
armLength: {
    pct: 200,
    min: 100,
    max: 500,
    menu: 'style',
    toAbs: (val, { options }, mergedOptions) =>
      ((mergedOptions.sizeConstant *
        (options.size === undefined ? mergedOptions.size : options.size) *
        2) /
        3.1415) *
      val *
      (options.type === undefined
        ? 1
        : options.type == 'octopus'
        ? 2
        : options.type == 'squid'
        ? 1.8
        : 1),
  },
  ```
  
[...]

PR #7125, opened 2024-09-26
Fix issue #7121 and two more bugs

woutervdub:


designs/octoplushy/src/head.mjs
```diff
@@ -613,15 +613,16 @@ const options = {
     min: 5,
     max: 500,
     menu: 'style',
+    // eslint-disable-next-line no-unused-vars
```
HaasJona, line 616: You can use `toAbs: (val, _, mergedOptions) => ` if you don't need the `options` variable and then don't need to disable the lint warning.

designs/octoplushy/src/head.mjs
```diff
@@ -613,15 +613,16 @@ const options = {
...
     toAbs: (val, { options }, mergedOptions) =>
       ((mergedOptions.sizeConstant * val * 2) / 3.1415) *
```
HaasJona, line 618: Unrelated to the fix, but could use `Math.PI` here for clarity

designs/octoplushy/src/head.mjs
```diff
@@ -613,15 +613,16 @@ const options = {
...
-      (options.type === undefined
+      (mergedOptions.type === undefined
```
HaasJona, line 619: Under which circumstances is the type undefined? Also wouldn't that be already handled by the `: 1` at the end?

designs/octoplushy/src/head.mjs
```diff
@@ -613,15 +613,16 @@ const options = {
...
         ? 1
-        : options.type == 'octopus'
-        ? 1.7
-        : options.type == 'squid'
-        ? 2
-        : 1),
+        : mergedOptions.type == 'octopus'
```
HaasJona, line 621: Is there a reason to not use `===` here?

BenJamesBen (approved):
I tested the PR in the lab, and it works as expected, with no issues seen.

designs/octoplushy/src/head.mjs
woutervdub: Done

designs/octoplushy/src/head.mjs
woutervdub: Done

designs/octoplushy/src/head.mjs
woutervdub: Don't think so. Removed

designs/octoplushy/src/head.mjs
woutervdub: Yes, that should not happen. Removed

designs/octoplushy/src/head.mjs
HaasJona: Use `_`, not `{}`, or you get another warning. 🙄

woutervdub: I added an ignore line ;-)

designs/octoplushy/src/head.mjs
```diff
@@ -613,15 +613,10 @@ const options = {
...
-      ((mergedOptions.sizeConstant * val * 2) / 3.1415) *
-      (options.type === undefined
-        ? 1
-        : options.type == 'octopus'
-        ? 1.7
-        : options.type == 'squid'
-        ? 2
-        : 1),
+    // eslint-disable-next-line no-empty-pattern
+    toAbs: (val, {}, mergedOptions) =>
```
joostdecock, line 617: No need to destructure the second parameter. You can use `_` like on like 628 so that we're consistent.

joostdecock (changes requested):
One small fix, then we're good to go.

joostdecock (approved):

PR #7128, opened 2024-09-26
Opal: Fixes intersection of parallel lines bug.

### LINKED ISSUE #7123: [bug]: Opal calculates intersection between two parallel vertical beams

When I generate an Opal and increase the **Back bib base angle** option to 90 degrees, the design tries to calculate a beam intersection between these 4 points in back.mjs line 124
```
  points.bibCurveMax = utils.beamsIntersect(
    points.waist,
    points.waist.shift(180 - options.backBibBaseAngle, 9001),
    points.bibHexagonLowerInside,
    points.bibHexagonUpperInside
  )
```

these are the values of the 4 points in my browser console:

```
Object { x: 0, y: -49.5, attributes: {…}, name: "waist", … }
 
Object { x: 1.6534568758687989e-12, y: -9050.5, attributes: {…}, … }
 
Object { x: -108.43749999999997, y: -220, attributes: {…}, name: "bibHexagonLowerInside", … }
 
Object { x: -108.43749999999997, y: -280, attributes: {…}, name: "bibHexagonUpperInside", … }
```

As you can see, these are both (almost) vertical lines, which causes stability problems in the calculation and crash the site (especially after applying https://github.com/freesewing/freesewing/pull/7122, which adds a missing check for intersecting two vertical beams).

_No response_

### Additional context

_No response_

Thrunic:
Fixes a bug where it was possible to attempt to find the intersection of two parallel lines. backBibBaseAngle is now limited to +85 degrees to prevent this.

Closes #7123

### DIFF
--- a/designs/opal/src/back.mjs
+++ b/designs/opal/src/back.mjs
@@ -639,8 +639,8 @@ export const back = {
   after: bib,
   draft: draftBack,
   options: {
-    // What angle the back bib leaves the outseam at. 0 is horizontal, 90 is vertical.
-    backBibBaseAngle: { deg: 25, min: 0, max: 90, menu: 'style' },
+    // What angle the back bib leaves the outseam at. 0 is horizontal, 90 would be vertical.
+    backBibBaseAngle: { deg: 25, min: 0, max: 85, menu: 'style' },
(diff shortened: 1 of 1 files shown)


bobgeorgethe3rd:
Closes #7123

joostdecock (approved):

PR #7129, opened 2024-09-28
chore: Port FreeSewing.dev to docusaurus

joostdecock:
The replaces the NextJS site powering FreeSewing.dev with a Docusaurus setup. It's part of my efforts to simplify FreeSewing's setup so we can focus on our core value proposition.


joostdecock:
Eslint is drunk 🤷‍♂️

PR #7137, opened 2024-10-02
Fix and improve path intersection methods

HaasJona:
This is mostly a general collection of fixes and improvements that I discovered why working on the seam allowance macro.

- Add path.intersectsBeam() method
- Add utils.beamIntersectsLine() method
- Simplify calculation and improve precision on beam intersections. A beam intersection situation between two parallel vertical lines going in opposite direction was not detected properly
- Document return types properly
- beamIntersectsCurve now uses the proper function from Bezier library instead of emulating it by constructing a huge line
- docs: path.intersect... methods never return false, they simply return an empty array in case of no intersection. This behavior is unfortunately not consistent between methods, but let's at least document the current behavior.

Note: This might affect compatibility as the detection of (almost) parallel paths in beam intersections and curve-beam intersections is now more reliable and possibly less/more tolerant. Patterns that relied on the previously incorrect behavior might fail in new ways. See #7128 for an example. I recommend to only push this for a major update. An alternative might be to have two version of the methods for compatibility reasons.  (like maybe two versions of the utils class)

### DIFF
--- a/packages/core/src/utils.mjs
+++ b/packages/core/src/utils.mjs
@@ -94,32 +94,60 @@ export function beamIntersectsY(from, to, y) {
  * @param {Point} a2 - Point 2 of line A
  * @param {Point} b1 - Point 1 of line B
  * @param {Point} b2 - Point 2 of line B
- * @return {Point} intersections - The Point at the intersection
+ * @return {Point|false} intersections - The Point at the intersection or `false` if the lines are parallel
  */
 export function beamsIntersect(a1, a2, b1, b2) {
-  let slopeA = a1.slope(a2)
-  let slopeB = b1.slope(b2)
-  if (slopeA === slopeB) return false // Parallel lines
+  const intersection = beamIntersection(a1, a2, b1, b2)
+  if (!intersection) return false
+  return intersection.p
+}
 
-  // Check for vertical line A
-  if (Math.round(a1.x * 10000) === Math.round(a2.x * 10000))
-    return new Point(a1.x, slopeB * a1.x + (b1.y - slopeB * b1.x))
-  // Check for vertical line B
-  else if (Math.round(b1.x * 10000) === Math.round(b2.x * 10000))
-    return new Point(b1.x, slopeA * b1.x + (a1.y - slopeA * a1.x))
-  else {
-    // Swap points if line A or B goes from right to left
-    if (a1.x > a2.x) a1 = a2.copy()
-    if (b1.x > b2.x) b1 = b2.copy()
-    // Find y intercept
-    let iA = a1.y - slopeA * a1.x
-    let iB = b1.y - slopeB * b1.x
+/**
+ * Finds the intersection of two endless lines (beams)
+ *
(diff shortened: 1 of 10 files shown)


BenJamesBen:
I haven't looked at the code changes yet, but I did run the org development environment to test the PR. For each and every FreeSewing design I generated a sample pattern. No problems were seen.

packages/core/src/utils.mjs
```diff
@@ -94,32 +94,60 @@ export function beamIntersectsY(from, to, y) {
...
+ * @param {Point} b1 - Point 1 of line B
+ * @param {Point} b2 - Point 2 of line B
+ * @return {{p:Point, t: number, u:number}|false} the intersection.
+ * The method will return `false` if the lines are (approximately) parallel or undefined,
+ * e.g., if both points of a line have (approximately) the same coordinate.
+ * Otherwise, p is the point of the intersection,
+ * t and u determine where the intersection lies relative to the points of line A and line B respectively.
+ * 0.0 means the intersection is on the first point, 1.0 on the second point.
+ */
+function beamIntersection(a1, a2, b1, b2) {
```
HaasJona, line 119: note: Not sure if we want to export this method. Access to t and u values could be useful for some designs, but would add something that requires stable versioning,

joostdecock:
> *I recommend to only push this for a major update.*

@HaasJona Are you saying this PR should only go into a 4.x branch? Or am I misunderstanding this?

HaasJona:
> > _I recommend to only push this for a major update._
> 
> @HaasJona Are you saying this PR should only go into a 4.x branch? Or am I misunderstanding this?

I'm not sure what your current release plan is. Since we're doing a major restruturing with the markdown documentation removal and so on, you probably won't push the current development version to stable anyway and I think it can go in now, but I'm warning that this can possibly cause compatibility issues, especially with unreleased designs. (I'm pretty confident that the official designs should all still work, with the 2 Opal issues now fixed).

joostdecock:
> I'm not sure what your current release plan is. Since we're doing a major restruturing with the markdown documentation removal and so on, you probably won't push the current development version to stable anyway and I think it can go in now, but I'm warning that this can possibly cause compatibility issues, especially with unreleased designs. (I'm pretty confident that the official designs should all still work, with the 2 Opal issues now fixed).

Right, that makes sense. Well, we currently support and default to NodeJS 18. But that went from active support to maintenance 8 days ago: https://github.com/nodejs/release#release-schedule

So we should drop support for NodeJS 18 at some point in the near future, and while it might not impact 99.9% of our users, it is a breaking change so that would mean a 4.0 release.

However, I would like for the current _renovations_ to find their footing prior to cutting a major release so that if anything else needs breaking we can do that.

Also note that the coming and going of NodeJS LTS releases creates a de-facto yearly major release cycle, and I don't hate the idea of having a major release every year.

So we'll merge this now, and try to get our homework done before releasing 4.x ASAP, hopefully before the end of the year.

joostdecock (approved):

PR #7138, opened 2024-10-02
Fix issue #7123 (Opal strap taper position calculation bug)

### LINKED ISSUE #7123: [bug]: Opal calculates intersection between two parallel vertical beams

Thrunic:
Closes #7123


designs/opal/src/back.mjs
```diff
@@ -136,11 +136,10 @@ function draftBack({
     options.backBibBaseCurve
   )
 
-  points.strapTaperCurveMax = utils.beamsIntersect(
+  points.strapTaperCurveMax = utils.beamIntersectsX(
```
HaasJona, line 139: Looks good to me, no issues anymore when ran with the changes in #7137

HaasJona (approved):

BenJamesBen:
I tested this PR in the org development environment. However, because this PR already includes PR #7128, I was not able to test whether this PR fixes the Back bib base angle issue of #7123. And the Strap taper position issue of #7123 does not appear on freesewing.org, so I am not able to test that aspect of this PR either.

joostdecock (approved):

PR #7144, opened 2024-10-04
fix(org): Fix generate a social media image

### LINKED ISSUE #7142: [bug]: "Generate a social media image" not working

On the "Generate a social media image" page, when I fill in the form and press the "Generate Image" button, nothing happens.

https://freesewing.org/new/img

_No response_

### Additional context

_No response_

BenJamesBen:
Fixes #7142

Web browsers can have call stack limitations that prevent calling functions with large arrays as parameters. (The bug was seen on the Safari desktop web browser. It is unclear if the bug affects any other browsers, but it is possible.)

This PR fixes the issue by iterating the array instead of calling a function with the entire array.

Iterating is a slower operation, and there is a noticeable delay between pressing the generate button and seeing the result. For this reason, this PR also adds loading status dialogs so the user is kept informed about the progress of the operation. (The "nailed it" success dialog is shown after the array iteration rather than immediately after the backend response. This is because the backend response is relatively quick. It is the iteration has the delay.)

### DIFF
--- a/sites/org/components/genimg/index.mjs
+++ b/sites/org/components/genimg/index.mjs
@@ -1,5 +1,6 @@
 import { nsMerge } from 'shared/utils.mjs'
-import { useState } from 'react'
+import { useState, useContext } from 'react'
+import { LoadingStatusContext } from 'shared/context/loading-status-context.mjs'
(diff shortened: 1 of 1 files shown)


joostdecock:
Good catch Ben :+1:

PR #7154, opened 2024-10-05
fix(github): Update GitHub labels and designs

BenJamesBen:
This PR adds recently-added designs to GitHub bug reports and auto-labelling.

- I removed `sites/lab` as a bug report area and auto-labelling since the lab no longer exists.
- However, I left intact `sites/email` and `sites/sanity`. It might be that we want to remove those as well? (Though it doesn't really matter since the directories no longer exist, there is no harm in having the auto-labelling set up for them.)
- I did not add auto-labelling for `naomiwu`. Since the GitHub label does not yet exist for that design, I'll just wait on it. (Though, `naomiwu` was already previously added as a bug report area in the new bug template.)
- There might be other, obsolete bug report areas and labels that we'll want to remove from the templates and auto-labelling, but that might be a project for another day.
- The camelcase for the `plugin-ringSector` label is correct. The existing GitHub label uses that capitalization/styling.


BenJamesBen:
I'm not sure what to do about the GitHub lint errors? GitHub seems to expect line 1 of `01_bug-report.yaml` to contain an assignment or function call, and it also expected it to end in a semicolon. But, instead the file contains 3 hyphens `---`:

```
Original file line number	Diff line number	Diff line change
@@ -1,4 +1,4 @@
---
```

 Check failure on line 1 in .github/ISSUE_TEMPLATE/01_bug-report.yaml
GitHub Actions
/ lint (18.x)
Expected an assignment or function call and instead saw an expression

 Check failure on line 1 in .github/ISSUE_TEMPLATE/01_bug-report.yaml
GitHub Actions
/ lint (18.x)
Missing semicolon                                                    

```
 name: 🐛 Bug report
description: Report a problem, or something that went wrong
title: '[bug]: '
```

joostdecock (approved):

PR #7169, opened 2024-10-14
Port Albert, Bee, Benjamin to Docusaurus

VSinerva:
Port of the mentioned design docs to Docusaurus based on the instructions by Joost. I've tried my best to test all the docs to make sure they work as intended.


joostdecock (changes requested):
I presume you started out by copying Aaron (which is fine) but you've kept the various illustrations from Aaron's sewing instructions (it should be rather clear when you review the changes in this PR).

Please remove those files from the PR, then we're good to go.

VSinerva:
> I presume you started out by copying Aaron (which is fine) but you've kept the various illustrations from Aaron's sewing instructions (it should be rather clear when you review the changes in this PR).
> 
> Please remove those files from the PR, then we're good to go.

Good catch! I removed those files. I missed them, since I copied the albert docs directly, and not aaron. Seems they've been carried forward with the albert docs for quite a while :)

joostdecock (approved):

Issue #7168, opened 2024-10-12
[chore]: Port designs documentation to Docusaurus


BenJamesBen:
### What seems to be the problem? 🤔

This Issue is to help track work/progress on porting designs documentation to Docusaurus.

### Claimed

---

### Done
- aaron: by @joostdecock 
- brian: by @joostdecock 
- albert: By VSinerva
- bee:  By VSinerva
- bella: by Ben
- benjamin:  By VSinerva
- bent: By VSinerva
- bibi: by HaasJona
- bob By VSinerva
- carlita: by Ben
- carlton: by Ben
- jaeger: by Ben
- simon: by Ben
- simone: by Ben
- umbra: by HaasJona
- breanna by @joostdecock 
- bruce by @joostdecock 
- cathrin by @joostdecock 
- florence by @joostdecock 
- florent by @joostdecock 
- hugo: by @joostdecock 
- lunetius: by @joostdecock 
- onyx by @joostdecock 
- opal vy @joostdecock 
- paco: by @joostdecock 
- shin: by @joostdecock 
- sven: by @joostdecock 
- sandy by @joostdecock
- shelly by @joostdecock  
- tamiko: by @joostdecock 
- teagan: by @joostdecock 
- tiberius by @joostdecock 
- trayvon: by @joostdecock 
- uma by @joostdecock 
- wahid: by @joostdecock 
- walburga by @joostdecock 
- yuri by @joostdecock 
- lily by @joostdecock 
- cornelius: by woutervdub 
- gozer: by woutervdub
- hi: by woutervdub
- hortensia: by woutervdub
- lumina: by woutervdub
- lumira: by woutervdub
- noble: by woutervdub
- octoplushy: by woutervdub
- otis: by woutervdub
- penelope: by woutervdub
- skully: by woutervdub
- tristan: by @joostdecock 
- waralee: by @joostdecock 
- lucy: By @joostdecock 
- jane: By @joostdecock 
- holmes: By @joostdecock
- huey: by @joostdecock 
- titan: by @joostdecock 
- diana: by @joostdecock 
- charlie: by @joostdecock 
### Does not need to be done
- examples
- legend
- plugintest
- rendertest
- magde

HaasJona:
I can definitely do Bibi and Umbra and I'm happy to help with leftovers, if needed.

VSinerva:
I could use some mindless grunt work right now. I'll grab Albert, Bee and Benjamin to start with.

[...]

PR #7170, opened 2024-10-15
fix: Fix DesignInfo mdx component handling of multiple design/code authors

HaasJona:
The variable `person` was undefined in the scope of the join method.

Related to #7168

### DIFF
--- a/sites/orgdocs/src/theme/MDXComponents/design-info.mjs
+++ b/sites/orgdocs/src/theme/MDXComponents/design-info.mjs
@@ -200,9 +200,14 @@ export const DesignInfo = ({ design }) => {
                 {typeof designInfo[design].design === 'string' ? (
                   <span>{designInfo[design].design}</span>
                 ) : (
-                  designInfo[design].design
-                    .map((person) => <span key={person}>{person}</span>)
-                    .join(<span key={`${person}_join`}>, </span>)
+                  designInfo[design].design.map((person, index) => (
+                    <span key={person}>
+                      {person}
+                      {index < designInfo[design].design.length - 1 && (
+                        <span key={`${person}_join`}>, </span>
+                      )}
+                    </span>
+                  ))
                 )}
                 <Subtle>
                   ( designer{typeof designInfo[design].design === 'string' ? '' : 's'} )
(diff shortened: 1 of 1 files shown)


sites/orgdocs/src/theme/MDXComponents/design-info.mjs
```diff
@@ -200,9 +200,14 @@ export const DesignInfo = ({ design }) => {
                 {typeof designInfo[design].design === 'string' ? (
                   <span>{designInfo[design].design}</span>
                 ) : (
-                  designInfo[design].design
-                    .map((person) => <span key={person}>{person}</span>)
-                    .join(<span key={`${person}_join`}>, </span>)
+                  designInfo[design].design.map((person, index) => (
+                    <span key={person}>
+                      {person}
+                      {index < designInfo[design].design.length - 1 && (
```
joostdecock, line 206: Can't we just replace:
```js
.join(<span key={`${person}_join`}>, </span>)
```
with:
```js
.join(<span>, </span>)
```

As the `join()` call turns the array in a string, we don't need a key, do we? (didn't try this)

sites/orgdocs/src/theme/MDXComponents/design-info.mjs
```diff
@@ -217,9 +222,14 @@ export const DesignInfo = ({ design }) => {
                 {typeof designInfo[design].code === 'string' ? (
                   <span>{designInfo[design].code}</span>
                 ) : (
-                  designInfo[design].code
-                    .map((person) => <span key={person}>{person}</span>)
-                    .join(<span key={`${person}_join`}>, </span>)
+                  designInfo[design].map((person, index) => (
+                    <span key={person}>
+                      {person}
+                      {index < designInfo[design].design.length - 1 && (
```
joostdecock, line 228: Same remark

joostdecock (commented):
@HaasJona Can't we just drop the key here?

sites/orgdocs/src/theme/MDXComponents/design-info.mjs
HaasJona: I don't think we want a string, we want an array of react elements. Maybe we could build a string, but then we wouldn't need/want any `span`s at all, but then we also couldn't highlight or link the people in the future.

BenJamesBen: I tried what I think the suggested alternative was, and it resulted in:
Screenshot 2024-10-23 at 6 00 44 AM

While the PR code changes produced a more correct result:
Screenshot 2024-10-23 at 6 01 26 AM

HaasJona: That's about what I expected, as converting react elements to string just results in `[object Object]`.

What could work is something like

```js
                  designInfo[design].design.join(", ")
```

which means we wouldn't create any react elements at all and just build a string, but long-term that would prevent us from formatting the names differently than the commas (e.g. by linking to a profile page of that person).

joostdecock: Alright, I got it wrong. Merging.

joostdecock (approved):

PR #7216, opened 2024-10-28
fix(core): Fix e-mail login with mfa enabled

### LINKED ISSUE #7215: [bug]: Mail login doesn't work with 2FA enabled

Logging in with an e-mail link doesn't work if the account has 2FA enabled. In that case you get a `mfaTokenRequired` notification from the backend server, which is displayed to the user as `susi:who knows` error.

The server answering with `mfaTokenRequired` when an mfa token is needed sort of makes sense, but there's nothing implemented in the client, that would ask the user for a token in the link-sign-in-case and then resubmit the request with a token in the POST body. The backend itself would register the login correctly if you send a POST request roughly like this:

```
POST https://backend3.freesewing.org/signinlink/xxxxa0a8-xxxx-4a93-b72c-94e361469375/0abcdefc6df4be08
Content-Type: application/json

{"token":"092787"}
```

so this is clearly a client problem.

sites/org

### Additional context

_No response_

HaasJona:
Fixes #7215 

This should works but needs some more testing


HaasJona:
I think this should be ready for review apart from some wonky react stuff.

Issue #1102, opened 2021-06-09
Add support for a sloping waistband to titan/charlie


joostdecock:
Titan/Charlie is drafted with a horizontal waistband. As in, when you look at it from the side, it's supposed to be horizontal.

That's fine for stick figures, but it's not so great for more curvy people. There, the waistband should slope downwards (once again when looking from the side). This will provide better fit for people who either have more pronounced behinds (and thus can use the extra height) or a bit of a belly (and are better served with a front that dips under the belly region).

Needs to be added to Titan, and inherited by Charlie. We can create an option for it. One question is how we can inform the user or perhaps automatically detect it.

joostdecock:
Example of a problematic draft: https://freesewing.org/account/patterns/duixa/

HaasJona:
I'm working on a prototype

HaasJona:
See also relevant discord threads: 

* https://discord.com/channels/698854858052075530/1226466696961851442
* https://discord.com/channels/698854858052075530/1281170445244174377

HaasJona:
Nevermind, there's already #3493 for that

PR #7228, opened 2024-11-05
Enhance titan waist balance

HaasJona:
This is a follow-up to #3493 and #1102 and adds the missing documentation and i18n strings for titan. (I haven't added documentation for all titan based designs, since that would be a lot of dumb copy-paste. As suggested before, it would be good in option documentation could be inherited by child designs somehow.)

This also adds a small feature that tries to calculate the optimal waistband angle based on the measurements and flags it to the user. (This flag message can be disabled by designs who don't want to use it, e.g. because they want to hardcode a waistband angle or do their own adjustments.) The formula is based on this sketch:

image

It's tested with the curated measurements, and my measurements, though, and seems to give reasonable results based on how curvy the bodies are. I haven't made a mock though or did large-scale testing.

For reference, the suggested angles for the measurement sets are:
* Elanor: 0° (not needed)
* Vanessa: 13.18°
* Ghislain: 0° (not needed)
* Joost: 2.88°
* My measurements: 3.8°
with my measurements definitely needing some adjustments as discussed on Discord.

This commit also contains minor code format fixes.

### DIFF
--- a/designs/titan/i18n/en.json
+++ b/designs/titan/i18n/en.json
@@ -88,6 +88,26 @@
       "t": "Waist ease",
       "d": "Controls the amount of ease at your waist"
     },
+    "waistAngle":  {
+      "t": "Waistband angle",
+      "d": "Change the angle of the waistband when viewed from the side"
+    },
+    "useWaistAngleFor": {
+      "t": "Use waistband angle for",
+      "d": "Apply the waistband angle to the front, back or both"
+    },
+    "useWaistAngleFor.both": {
+      "t": "Both",
+      "d": "Apply the waistband angle to the front and back"
+    },
+     "useWaistAngleFor.backOnly": {
+      "t": "Back",
+      "d": "Apply the waistband angle only to the back"
+    },
+     "useWaistAngleFor.frontOnly": {
+      "t": "Front",
+      "d": "Apply the waistband angle only to the front"
+    },
     "grainlinePosition": {
       "t": "Grainline position",
       "d": "Controls the horizontal position of the leg relative to the seat"
--- a/designs/titan/src/back.mjs
+++ b/designs/titan/src/back.mjs
@@ -200,7 +200,7 @@ function titanBack({
       // Revert back to the previous rotation.
       for (const i of rotate) {
         points[i] = saved[i]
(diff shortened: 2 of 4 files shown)


HaasJona:
@anna-puk I've disabled the waist angle for Lily here before realising that both the waist angle patch and lily are made by you. If you want to enable it there, it might need some additional i18n strings because Lily doesn't inherit them from titan.

HaasJona:
@anna-puk there are some minor test failures with Lily because that doesn't inherit the i18n messages from Titan. Want me to fix that? Not sure if Lily needs the new waist angle option.

HaasJona:
There's a mistake in the calculations that I want to check

joostdecock:
This is both in draft, and pending review. So I'm not certain what the intent is @HaasJona ?
Did you forget to take it out of draft? Or did you want me to review it as-is now?

In any case, some feedback on this:

> As suggested before, it would be good in option documentation could be inherited by child designs somehow.

I am not in favor of this as a general principle (making this the default), because in practice, this means that there is just no documentation written for those options, ever.
Ideally, each design would document each option in a way that makes sense for that design, rather than just re-use the block/base docs. 

That being said, maintaining the same documentation in two places is obviously not something we want.
But that is a more general rule than just design options. And thus we have a more general solution for that. As shown here: https://github.com/freesewing/freesewing/blob/develop/sites/orgdocs/docs/designs/sven/options/sleevecapfrontfactorx/readme.mdx

HaasJona:
I had been requesting a review and then notices some minor issues in the calculations myself, so I put it temporarily into draft so that it doesn't get merged too fast. Couldn't find an obvious way to remove the review request.

I've now improved the calculations, so feel free to review it.

The point about documentation makes sense. I think in this case, it would be good if individual authors of titan based designs could check if they want to include these wasitband angle options (and document them) or if they just want to hardcode them to `0`, `'both'` and `false` respectively to disable them.

HaasJona:
Oh wait, I've found another bug in the calculation, let me check that again.

HaasJona:
OK, I confused waist-to-hips and hips-to-seat, should be better now.

PR #7232, opened 2024-11-14
fix(scripts): Remove references to lab development environment

BenJamesBen:
I've used the term "org development environment". I'm not sure if there might be a better term for it.


joostdecock (approved):
We will probably need to revisit this once we reach the end of the somewhat turbulent period where we're rearranging things, but for now this will do.

PR #7271, opened 2024-11-25
overhaul(Bee) Update, Fixes & Overhaul for Bee

### LINKED ISSUE #6869: [bug]: Bee incorrect cutlist & tie configurations

https://discord.com/channels/698854858052075530/1251656063418372206/1251656063418372206
When the cutlist was added it did not fully encompass all configurations of Bee.

designs/bee

### Additional context

_No response_

bobgeorgethe3rd:
Closes #6869

This is a major clean up and overhaul of Bee's code.

Changes
- Reinstating the ability to toggle the colours of neck and band tie independently
- Reinstating the ability to toggle pointed edge of neck and band tie independently
- Simplify snap options
- Simplfication of the code
- Simplify Cutting Instructions
- Cutlist re-work
- Flag message also now appears when `!options.ties`
- Neck tie is no longer hidden when `options.pointedNeckTieEnds == true`
- Band tie is no longer hidden when `options.crossBackTies == true`
- Band tie is no longer hidden when `options.pointedBandTieEnds == true && !options.crossBackTies`

Fixes
- Neck ties no longer cut on fold (This was an error from when originally coded)
- Band tie locks in duo colours when `options.crossBackTies == true` (this is why there were incorrect notch placements)
- `options.bandLength` move to `neck,mjs`

Additions
- `options.duoNeckTieColours`
- `options.duoBandTieColours`
- `options.pointedNeckTieEnds`
- `options.pointedBandTieEnds`
- New option markdown

Removed
- `options.duoTieColors`
- `options.pointedTieEnds`
- removed option markdown
- Flag message has had expand flags removed

Things to Note
- I have not edited the i18n files due to their depreceation. If I do need to do this please let me know.
- The cut list is not using the mirrored preset to hopefully lessen confusion.

Moving forward, as the original coder of this design, in an effort to maintain the design as intended by the original designer in the future, it would streamline the process and be helpful if any requests for maintenance were sent directly to me.


BenJamesBen:
- I notice that the 3 Dart options (bustDartAngle, bustDartMinimum, waist DartCurve) appear in the UI for Bee. However, they don't seem to affect the generated pattern?
- I don't understand how the design works, but I notice that for the Cross Back Tie variant, increasing the Band Length option causes the Neck Ties to become shorter (in addition to the Band becoming longer as expected). Is this the correct behavior?

bobgeorgethe3rd:
> * I notice that the 3 Dart options (bustDartAngle, bustDartMinimum, waist DartCurve) appear in the UI for Bee. However, they don't seem to affect the generated pattern?
> 
>     * I don't understand how the design works, but I notice that for the Cross Back Tie variant, increasing the Band Length option causes the Neck Ties to become shorter (in addition to the Band becoming longer as expected). Is this the correct behavior?
> 
>     * I notice that band tie width seems to snap at certain widths, but I don't see this behavior documented.
> 
>     * The Neck Tie Part 2 title seems a little small, and I had difficulty reading it in the UI.
> 
> 
> But, otherwise, I tested the PR and the design seems to work without any issues.

- `bustDartAngle, bustDartMinimum, waistDartCurve` These are new additions to Bella since Bee was created I will look and check if lock them.
- I believe it is but I will double check the old code and make sure a positive hasn't become a negative
- It was always a snapping option. This is the same for the neck ties. It is based on a quarter inch.
- That's because the neck-tie part is small and intention is for it too fit the boundaries of the neck tie.

bobgeorgethe3rd:
@BenJamesBen 
Ok so upon inspection

- Locked `bustDartAngle`, `bustDartMinimunFabric` & `waistDartCurve`
Yep so I as I thought these were new additions Bella that were not present when Bee was drafted.
As you saw they do not affect the draft so I have locked them.
As a point of reference this is what I did the first time when drafting Bee for any option that did not affect Bee hence why some are unlocked and others are not.  

- Upated neckTieLength & Altered the option behaviour of `bandLength` & `neckTieLength` when in crossBackTie variation.
Yea so somehwere in my original this got a bit overengineerede and incorrect. Looking at how I was coding the inital length I could tell that it was not meant to be shrinking so I have elected to change this behaviour as I said judging by my original code this was an unintended error so good catch.

- Updated `neckTieWidth` & `bandTieWidth` markdown to reflect that they are snap options.
Understand if this is the new standard to document that it is snap, wasn't at time but it doesn't hurt to add :)

- Updated `neck-tie` & `band-tie` title scale
I can see that at some point the neck tie scale was changed from it's original 0.2 to 0.5 and initially overhaul I changed it back it unkowingly. I assume this is the new standard to make it UI legible rather than fitting in the boundaries so have change `neck-tie` back to 0.5 but I have also shrunk `band-tie` back to it's original 0.5 rather than 0.7 for some consistancy with the ties.

bobgeorgethe3rd:
Branch rebased to v4 develop. Conflicts resolved.

PR #7280, opened 2024-11-26
fix(core): Improve Path.offset() for curves

### LINKED ISSUE #7219: [bug]: Part 2 of Cathrin show weird SA

When a `cathrin` is generated with specific measurements, part of the SA is 'clipped'. Can be reproduced with curated set 'Vanessa'. Using set 'Joost' or 'Elanor' does not show this anomaly. 

image

_No response_

### Additional context

_No response_

BenJamesBen:
Fixes #7219

This PR does not change the behavior of `Path.offset()`. It only tries to improve how offset curves are generated by changing `Path.__pathOffset()`. It does this to work around a `bezier-js` issue that sometimes causes offset curves to be incomplete.

Previous behavior:
1. The array of forward offset bezier curves is generated via `bezier-js`.
2. All legitimate curves in the forward array are converted to the forward offset path.
3. The forward offset path is used.

New behavior:
1. The array of forward offset bezier curves is generated via `bezier-js`.
2. The array of reverse offset bezier curves is generated via `bezier-js`.
3. All legitimate curves in the forward array are added to the forward offset path.
4. All legitimate curves in the reverse array are added to the reverse offset path.
5. Whichever offset path, forward or reverse, has the larger number of operations is used.

The algorithm uses a shortcut to determine which offset path is longer and therefore more correct. Instead of calculating the length of the offset paths, it instead looks at the number of operations. It assumes that more operations means a longer and more correct path.

### DIFF
--- a/packages/core/src/path.mjs
+++ b/packages/core/src/path.mjs
@@ -1553,16 +1553,39 @@ function __pathOffset(path, distance, log) {
         cp2 = new Path().__withLog(path.log).move(op.to).curve(op.cp2, op.cp1, current)
         cp2 = cp2.shiftAlong(cp2.length() > 2 ? 2 : cp2.length() / 10)
       } else cp2 = op.cp2
-      let b = new Bezier(
+      const curve_offset_forward = []
+      const curve_offset_reverse = []
+      const forward = new Bezier(
         { x: current.x, y: current.y },
         { x: cp1.x, y: cp1.y },
         { x: cp2.x, y: cp2.y },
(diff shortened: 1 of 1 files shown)


HaasJona:
I'm not a huge fan of this, it would be better to figure out the bug of bezier-js and solve it upstream. Same as #3367. I suspect that these are actually the same bug.

HaasJona:
The upstream PR https://github.com/Pomax/bezierjs/pull/219 might fix this is a more sensible way. However I haven't tested this explicit bug.

PR #7305, opened 2025-01-02
designer notes for lily

anna-puk:


joostdecock:
Added to the `v4` branch since the target file has been moved :)

PR #7324, opened 2025-01-25
Fix the part selection ("only") menu

HaasJona:
- Handle mission i18n better
- The t() function is not available in this scope
- set config.choiceTitles for the menu so the menuitems get labels


joostdecock (approved):

joostdecock:
Nice work. I've been working on this recently so sorry if it's still a bit of a mess

PR #7370, opened 2025-03-15
Prevent missing settings object causing exceptions when opening design editor

HaasJona:
Potential fix for the crash accessing `state.settings` in `DesignOptionsMenu.mjs` when opening https://v4.freesewing.org/editor/#s={%22design%22%3A%22hi%22%2C%22view%22%3A%22draft%22}

This could possibly be also fixed in `DesignOptionsMenu.mjs` instead, but there seem to be a lot of code locations where `settings` is potentially accessed.

### DIFF
--- a/packages/react/components/Editor/index.mjs
+++ b/packages/react/components/Editor/index.mjs
@@ -92,6 +92,11 @@ export const Editor = ({ config = {}, design = false, preload = {}, setTitle = f
         }
       : { ...state, _: { ...ephemeralState, missingMeasurements } }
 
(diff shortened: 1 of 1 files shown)


joostdecock:
Thanks for creating this PR with the fix. Good idea to always ensure settings is intialized, however I prefer to make it part of the initial initialisation, so I implemented that in [e156a9c](https://github.com/freesewing/freesewing/commit/e156a9ca694487f7b570e900d3421ff50e391cea)

Issue #7365, opened 2025-03-12
[bug]: Simon Pattern Says to Only Cut 1 Yoke


jpmunic:
### What seems to be the problem? 🤔

The generated Simon shirt pattern says to only cut 1 yoke from main fabric, but the sewing instructions say to cut two. The correct amount is two, and the pattern should be updated.

### Design / Plugin / Package 🧐

designs/simon

### Additional context

<img width="974" alt="Image" src="https://github.com/user-attachments/assets/c85faafb-fb46-4aa0-ad81-4f6ffcc686fb" />

<img width="691" alt="Image" src="https://github.com/user-attachments/assets/ed111455-f385-42c5-af09-820a3e27bd1b" />

PR #7378, opened 2025-03-22
Fix the cutting instructions for the Simon Yoke

jpmunic:
- Previously it said to cut 1 on the fold, which is wrong. You don't cut the yoke on the fold, it's two pieces.
- Now it says to cut 2 mirrored from main fabric, which is correct.

Bug report that I submitted: https://github.com/freesewing/freesewing/issues/7365

before:
yoke

after:
Screenshot 2025-03-21 at 10 37 23 PM

### DIFF
--- a/designs/simon/src/yoke.mjs
+++ b/designs/simon/src/yoke.mjs
@@ -1,10 +1,24 @@
 import { back } from './back.mjs'
 import { splitYoke } from './options.mjs'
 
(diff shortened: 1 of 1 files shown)


joostdecock (approved):

joostdecock:
@all-contributors please add @jpmunic for bug

joostdecock:
Thanks for the fix @jpmunic :pray: 
I've cherry-picked this in the `v4` branch which is what we're working on now for the upcoming v4 release.

PR #7379, opened 2025-03-22
Add measurement validation

HaasJona:
This is something I've been working with and an improved version of #7030. Feedback welcome.

Basically, what i did was to create a "sewing pattern" (with the code name bonny, but I don't mind renaming it if someone has a better idea) that tries to render a drawing of how the software thinks your body looks based on the measurements. I also ported the warnings from #7030 to flag the user about suspicious measurements.

To create the pattern I used the MakeHuman Blender plugin to create a 3D model of an "average" human. I then used some Blender plug-in to render that 3D model to a scale SVG file and then I created a sewing pattern from that SVG file, scaling the distances with the appropriate body measurements.

Since bonny is not a real sewing pattern, I omitted sewing instructions and listing it in the designs, and I just made it available on the measurement set page.

Preview images:

image
New Button to validate measurements (only visible if Compare measurements is enabled in account)

image
Result after clicking on the button. At the top it lists potential issues with the measurement (these are fine) and an estimation of your height.

At the bottom is a button to display a body preview:
image
Preview displayed after clicking button.

Here is a validation result of a measurement set with some warnings:
image


i18n/designs.mjs
```diff
@@ -23,7 +23,7 @@ import { i18n as examples } from '@freesewing/examples'
 import { i18n as florence } from '@freesewing/florence'
 import { i18n as florent } from '@freesewing/florent'
 import { i18n as gozer } from '@freesewing/gozer'
-import { i18n as hi } from '@freesewing/hi'
+import { i18n as bonny, i18n as hi } from '@freesewing/hi'
```
joostdecock, line 26: You are importing from hi here? That is probably a mistake?

Please add an import for bonny (these imports are also kept in alphabetical order)

joostdecock (changes requested):
Very excited about this. Happy to merge this into v4 as-is, except that I think there's an incorrect import. Please check that out first.

i18n/designs.mjs
HaasJona: Oops.

HaasJona:
I think that was caused by some rogue IDE auto import. Should be fixed now.

joostdecock (approved):

PR #7380, opened 2025-03-22
(potential) changes to hot-patch BezierJS in core

joostdecock:
This is to address issues with BezierJS. See:

- https://github.com/Pomax/bezierjs/issues/203
- https://github.com/Pomax/bezierjs/pull/219
- https://github.com/freesewing/freesewing/issues/7219

Putting this up to be able to cooperate on this.
Currently the tests are failing because we do not have a good test case.
I am also not sure that throwing is the way to go here.

Ping @HaasJona


HaasJona:
We can probably improve on this function by doing a binary search approach. Instead of linearly going to the curve and trying to split it at each point we could instead split it in half, check if each half is simple, if it isn't, split that half again in half, and so on, until we have only simple curves left.

Of course we wouldn't want to return all these halves as result, we would only use that method to find the best points to split the curve. But we would avoid running into super small step sizes with potentially bad performance and could do a binary search instead of a linear search, which would be much more efficient.

